PR #7, opened 2015-06-05
TAUR-1217 Fix grok pipeline package install

unixorn:
* Ensure pip version is current on the AMI
* Add warning about installing yum-s3 before s3-backed repo files. Doing this in the wrong order breaks yum, so give people an explicit warning so they don't get to find that out the hard way.
* Rakefile cleanups
    * Add grok_repos dependency to validate stanzas so they don't choke when run in a fresh checkout
    * Don't bother to re-download yum s3 plugin if it is present
    * Don't download bootstrap-salt.sh if it is already present
    * Don't download epel rpm if it is already present
    * Replace hardcoded url for yum s3 plugin with a variable
* Run pip install -r requirements.txt in grok RPM postinstall script

### DIFF
--- a/grok/grok/pipeline/scripts/rpm-creator/post_install_grok
+++ b/grok/grok/pipeline/scripts/rpm-creator/post_install_grok
@@ -49,4 +49,7 @@ fi
 
 pip install psutil==1.2.1
 
+# Install requirements.txt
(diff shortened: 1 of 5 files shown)


unixorn:
cc: @jcasner, @shantanoo

PR #10, opened 2015-06-08
Add install-infrastructure.sh bash script for consistency with other …

oxtopus:
…install-* scripts in root.

cc @unixorn


unixorn:
:+1:

PR #15, opened 2015-06-08
TAUR-1217 Fix rpm versions

unixorn:
We use the commit count as the rpm release version, so when we switched to the numenta-apps repo, we started generating rpms that had lower release versions, and yum assumed they were outdated.

Bumping the rpm version to 1.7.1 to get yum to see that newer rpms are in fact newer.

### DIFF
--- a/infrastructure/infrastructure/create-grok-rpm
+++ b/infrastructure/infrastructure/create-grok-rpm
@@ -37,7 +37,7 @@ time ./create-numenta-rpm --rpm-flavor grok \
   --rpm-name nta-products-grok \
   --postinstall-script "${reporoot}/grok/grok/pipeline/scripts/rpm-creator/post_install_grok" \
   --cleanup-script grok/grok/pipeline/scripts/rpm-creator/clean-grok-tree-for-packaging \
-  --base-version 1.7.0 \
+  --base-version 1.7.1 \
   --whitelist grok \
   --whitelist htmengine \
   --whitelist install-grok.sh \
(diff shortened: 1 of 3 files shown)


unixorn:
@jcasner please CR

unixorn:
cc: @shantanoo

infrastructure/infrastructure/create-grok-rpm
```diff
@@ -37,7 +37,7 @@ time ./create-numenta-rpm --rpm-flavor grok \
   --rpm-name nta-products-grok \
   --postinstall-script "${reporoot}/grok/grok/pipeline/scripts/rpm-creator/post_install_grok" \
   --cleanup-script grok/grok/pipeline/scripts/rpm-creator/clean-grok-tree-for-packaging \
-  --base-version 1.7.0 \
+  --base-version 1.7.1 \
```
jcasner: Instead of hard coding, we should use the build parameter in Jenkins

unixorn:
@jcasner ok, I'm having `create-numenta-rpm` default the base-version to RELEASE_VERSION now.

infrastructure/infrastructure/create-numenta-rpm
```diff
@@ -66,6 +66,10 @@ def parseCLA():
 
   parser = argparse.ArgumentParser(description="Create a products RPM")
 
+  defaultVersion = os.environ.get("RELEASE_VERSION")
```
jcasner: @unixorn Please call this env var out in the README file.  Also, you can simplify this to:
```python
defaultVersion = os.environ.get("RELEASE_VERSION", "1.7.1")
```
which will use the env var if it exists and the default value of 1.7.1 if not.

jcasner:
:+1:

PR #18, opened 2015-06-09
Update Grok pipeline to stop building / installing nupic from a wheel file. We're managing nupic version via requirements.txt now, so no need to do it through command lines in the pipeline; it just confuses things.

jcasner:
CC @shantanoo @unixorn @rupali

I still need to test this, but just a heads up that this is a work in progress. I started this after noticing build 229 failed with a timeout trying to download the wheel file from S3, even though we have nupic 0.2.2 specified in requirements.txt.


jcasner:
@shantanoo @rupalibehera @unixorn this is now ready for review. Appreciate your feedback!

unixorn:
:+1:

PR #27, opened 2015-06-09
Update version of nupic

jcasner:


htmengine/requirements.txt
```diff
@@ -4,6 +4,6 @@ sqlalchemy==0.9.4
 msgpack-python==0.3.0
 validictory==0.9.1
 mysql-python==1.2.5
-nupic==0.2.2
+nupic==0.2.3
 nta.utils>=0.0.0
```
shantanoo: Should this be `>` instead of `>=`?

jcasner: No.  The current version is effectively 0.0.0 because it hasn't been defined.

PR #28, opened 2015-06-10
Change how NuPIC gets installed

jcasner:
@oxtopus I think this will do the trick, but I might be missing something for the pip install command. I'd love your feedback to make sure this is done correctly. Thanks!

### DIFF
--- a/htmengine/setup.py
+++ b/htmengine/setup.py
@@ -1,4 +1,19 @@
+import os
+import sys
+
 from setuptools import setup, find_packages
+
+from infrastructure.utilities.cli import runWithOutput
+
+
+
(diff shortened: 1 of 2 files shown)


htmengine/setup.py
```diff
@@ -1,5 +1,20 @@
...
+import sys
+
 from setuptools import setup, find_packages
 
+from infrastructure.utilities.cli import runWithOutput
+
+
+
+# Install NuPIC from the right place; do NuPIC version locking in 
+if "linux" in sys.platform:
```
oxtopus: :-1:  We shouldn't be shelling out to pip here.

jcasner:
Per discussion with @oxtopus closing this PR and looking for a new approach.

PR #30, opened 2015-06-11
HOTIFIX Update settings to fabric.api.run so we can catch errors properly

unixorn:
If warn_only isn't explicitly set to true, fabric.api.run fails so hard we can't check the command exit code, which breaks our retry logic.

This was causing our instance start script to sporadically fail during boot even though we had set it to retry the startup scripts.


unixorn:
@jcasner, @shantanoo, please CR.

jcasner:
:+1:

PR #35, opened 2015-06-11
Fix NPE on unit test

lscheinkman:
@jcasner Please review

This commit (ec1ae92) should fix the taurus-mobile pipeline.

This PR also updates the `docker` image to install android emulator and run the tests just like the pipeline. The image is not an exact copy of the pipeline but it is close enough for now.


jcasner:
:+1:

PR #37, opened 2015-06-12
TAUR-1217 Use a base AMI that has stock CentOS upstream OpenSSL

unixorn:
* Add packer configuration to build an infrastructure AMI that has a 256GB root partition
* Non-grok packer configurations do not install the grok yum repo configurations
* CentOS base packer configuration now has a minimal set of yum repo files
* Install EPEL repo before installing Saltstack
* grok-plumbing packer configuration now uses stock base CentOS

When Heartbleed came out, AWS insisted that instead of using the patched OpenSSL RPMs that CentOS provided, we had to patch and compile our own 1.0.1.h version or the Grok AMIs would not pass acceptance tests for being added to the marketplace.

This causes a bunch of follow on aggravation when other CentOS rpms (and third party ones like the ones in the mysql community repository) depend on OpenSSL and sanely expect that we would have the upstream version instead of a hacked together one.

This reverts that brittle hack so people can build their own Grok AMI without having problems when they try to install things that depend on OpenSSL.


unixorn:
@jcasner @shantanoo please CR

jcasner:
:+1:

PR #39, opened 2015-06-12
Address nupic installation issues on CentOS linux boxes

jcasner:
@unixorn @rhyolight @oxtopus this PR should hopefully address the nupic install issues in CentOS. In the nupic pipeline, it now generates an egg file as well as a wheel file.  Those both get uploaded to the numenta pypi server (https://pypi.numenta.com).  It also adds a dependency_links section when the system is linux.  I think we may want to get more clever to support other flavors of linux in the future, but this should work for the immediate problem in our pipelines.


oxtopus:
LGTM

PR #49, opened 2015-06-19
table_name_suffix should never be blank

unixorn:
Set the default dynamo table suffix so it isn't blank and is obvious that it should be changed


unixorn:
@lscheinkman please CR

lscheinkman:
:+1: LGTM

PR #50, opened 2015-06-19
InstanceNotFoundError exception wasn't being imported

unixorn:
@jaredweiss this fixes the error you ran into, please CR.


jaredweiss:
LGTM

PR #52, opened 2015-06-19
708 fix htm engine test grok dep

mihail911:
TAUR 708 -- fixing reference to grok in htmengine tests


scottpurdy:
:+1:

htmengine/tests/unit/runtime/metric_streamer_util_test.py
```diff
@@ -84,8 +83,7 @@ def testScrubDataSamplesOutOfOrder(self, _repositoryMock):
 
   def testSendInputRowsToModel(self):
     """ Test MetricStreamer._sendInputRowsToModel """
-    metricDataOutputChunkSize = grok.app.config.getint(
-      "metric_streamer", "chunk_size")
+    metricDataOutputChunkSize = 1440 #Hardcode value
```
vitaly-krugl: @mihail911 and @scottpurdy: we shouldn't hardcode this value here and in the other tests in this file. Hardcoding makes this test brittle. `metric_streamer_util` creates the correct config object, so let's go ahead and take advantage of it in this unit test. It's more robust than hardcoding and also makes it self-documenting where this value comes from:

```
metricDataOutputChunkSize =  metric_streamer_util.config. getint(
  "metric_streamer", "chunk_size")
```

PR #53, opened 2015-06-19
TAUR-1217 Fix stale update-individual-repo

unixorn:
Rely on configuration management to have the correct `update-individual-repo` script installed in /usr/local/bin instead of messing about with stale repo versions.


unixorn:
@jcasner please CR

jcasner:
:+1: but we need to get back to the premise of an atomic commit

unixorn:
Agreed.

PR #54, opened 2015-06-19
TAUR-990 We now publish rpms to public.groksolutions.com/yum

unixorn:
Update to read from public.groksolutions.com/yum/x86_64 since that's where the rpms are now published.


grok/grok/pipeline/src/bake_ami.py
```diff
@@ -47,7 +47,7 @@
 g_config = yaml.load(resource_stream(__name__, "../conf/config.yaml"))
 
 
-S3_YUM_BUCKET = "yum.groksolutions.com"
+S3_YUM_BUCKET = "public.groksolutions.com"
```
jcasner: public.numenta.com, right??

jcasner:
:-1:

PR #55, opened 2015-06-19
TAUR-990 Fix domain in bucket name

unixorn:
Can't believe I missed that the domain was old. Thanks for the catch, @jcasner


jcasner:
:+1:

PR #59, opened 2015-06-22
TAUR-1137 Revert tweet aggregation due to unacceptable processing latency

vitaly-krugl:
We'll need to rethink the solution. The initial solution was having trouble clearing the backlog and keeping up with incoming data even when we bumped up the cumulative write throughput to 900 and increased dynamodb-service concurrency to 16 on a powerful EC2 instance.

CC @oxtopus @jcasner @lscheinkman

### DIFF
--- a/htmengine/htmengine/runtime/anomaly_service.py
+++ b/htmengine/htmengine/runtime/anomaly_service.py
@@ -19,6 +19,7 @@
 #
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
+
 import itertools
 import json
 import logging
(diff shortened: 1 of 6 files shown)


jcasner:
It should be noted that @oxtopus warned us of the risk involved in this approach and cautioned that performance would take a hit.  It was unclear when the work was started how bad the performance impact would be, so we decided to proceed anyhow.

:+1: to backing these out. Thanks @vitaly-krugl

vitaly-krugl:
Verified locally on my dev laptop

vitaly-krugl:
NOTE: `taurus.metric_collectors` integration test `integration.metric_utils_test.MetricUtilsTestCase testMethod=testCreateAllModels` was failing in master before this change and is still failing.

PR #66, opened 2015-06-24
TAUR-1248: Load tweets 100 tweets at the time from both sides of the selection #send-for-review

lscheinkman:


lscheinkman:
CC: @jcasner

mobile-core/android/src/main/java/com/numenta/core/data/Metric.java
```diff
@@ -265,6 +267,26 @@ public void setLastTimestamp(long lastTimestamp) {
...
+
+    /**
+     * Process custom deserialization
+     */
+    private void readObject(java.io.ObjectInputStream in)
+            throws IOException, ClassNotFoundException {
+        // Deserialize serializable fields
+        in.defaultReadObject();
+
+        // "JSONObject" is not serializable and marked as "tansient" therefore we have to parse the
```
jcasner: s/tansient/transient

taurus-mobile/android/app/src/main/java/com/numenta/taurus/twitter/TwitterDetailActivity.java
```diff
@@ -588,36 +609,71 @@ public boolean onData(Tweet tweet) {
...
-                                // Force flush current bucket
+                            long left = selectedTimestamp;
+                            long right = selectedTimestamp;
+                            while (left > range.first || right < range.second) {
+                                if (right < range.second) {
+                                    // Load upper half in chunks of 100 tweets at the time
+                                    long to = right + DataUtils.METRIC_DATA_INTERVAL;
+                                    int count = _tweetCountByDate.get(to);
+                                    Integer value;
+                                    while (to < range.second && count < 100) {
```
jcasner: @lscheinkman it looks like you're still loading all tweets before, then all tweets after, the mark.  I think what we discussed was to alternate loading.  Something like this:

`... - 8 - 6 - 4 - 2 - 1 - 3 - 5 - 7 - ...`

lscheinkman: @jcasner This code will load all tweets in chunks of 100, first from the 'right' and then from the 'left'  until there no more tweets in the range (see line 614). Just as you describe.
This is the inner loop used to calculate the date range containing at 100 tweets for each side.

jcasner: :+1: got it - thanks for the clarification!

PR #69, opened 2015-06-25
TAUR-1248: Disable "fast scroll" bar and try to prevent "jitter" on the list

lscheinkman:
@jcasner Please review.
There is still some "jitter" on the white bar that I'm looking into it and I was not able to reproduce the bucket shift


jcasner:
Closing this one since we opted for the message approach.  We can redo this later if needed.

PR #70, opened 2015-06-25
Only install numpy on instances that actually need it.

unixorn:
Bad practice to install software on machines that don't use it.


unixorn:
@jcasner, @shantanoo please CR.
Make sure to also merge [numenta-private-salt-infrastructure/pull/97](https://github.com/GrokSolutions/numenta-private-salt-infrastructure/pull/97) when this is merged.

jcasner:
I'm only neutral at best on this... I would much rather have python dependencies be managed by the apps themselves via requirements.txt.  While this is circumventing a known issue, I think the safer approach is to include a `pip install numpy==1.9.2` command in the `install-grok.sh` and `install-taurus.sh` files and leave all python requirements out of salt entirely.

unixorn:
I'm even happier with that solution.

jcasner:
:+1:

oxtopus:
@jcasner let's talk about this when I get back from vacation...

jcasner:
@oxtopus that sounds awesome - might need to wait until I'm back from vacation - I'm probably taking next week off

PR #71, opened 2015-06-25
Fix test after removing test specific API from GrokApplication

lscheinkman:
@jcasner Please review.
This should fix 'MetricListAdapterTest'.
See #64


jcasner:
:+1:

PR #57, opened 2015-06-22
TAUR-1217 More updates for the grok pipeline

jcasner:
@unixorn @shantanoo please review

PR #72, opened 2015-06-25
TAUR-1217 Fix epel repo issue found during testing

unixorn:
We were installing EPEL by directly copying files into `/etc/yum.repos.d`. This was a horrible idea since EPEL provides an RPM to take care of this, which we can install directly from their website so we are installing the current version instead of a stale version.

When reviewing [#57](https://github.com/numenta/numenta-apps/pull/57), I noticed that the raw file we were using had mirrorlist disabled. Switching that back on caused yum to fail because of an SSL error, which revealed that we had a stale ca-certificates RPM installed. However, now that mirrorlist was on, the SSL error caused yum to fail when we tried updating ca-certificates.

We now install (both in Salt and in the epel installation packer script) ca-certificates first, then install EPEL.


unixorn:
@jcasner, @shantanoo please CR

jcasner:
:+1: after you merge from master

PR #73, opened 2015-06-26
Add TODOs re: removing out-of-place `pip install` commands.

oxtopus:
@jcasner the `install-*.sh` scripts are the wrong place to include `pip install numpy==1.9.2`, which was added, AFAIK as a workaround for our CentOS+Anaconda approach specific to our own infrastructure.

The TODOs are added as a reminder for future work.


jcasner:
:-1: Those are to work around an issue specific to setup.py and numpy==1.9.2.  AFAIK, the issue has nothing to do with anaconda, and everything to do with `python setup.py develop` being unable to install `numpy==1.9.2` from requirements.txt.

oxtopus:
Closing in favor of #74

PR #74, opened 2015-06-26
Include numpy in `setup_requires` (and `install_requires`)

oxtopus:
@jcasner Seems to work for me with an instance started with internal `nta-start-instance` tool.

### DIFF
--- a/htmengine/setup.py
+++ b/htmengine/setup.py
@@ -3,7 +3,13 @@ import sys
 
 from setuptools import setup, find_packages
 
-requirements = map(str.strip, open("requirements.txt").readlines())
+setup_requirements = []
+install_requirements = []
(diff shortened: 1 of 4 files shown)


jcasner:
:+1:

htmengine/setup.py
```diff
@@ -3,7 +3,13 @@
 
 from setuptools import setup, find_packages
 
-requirements = map(str.strip, open("requirements.txt").readlines())
+setup_requirements = []
+install_requirements = []
+for req in open("requirements.txt").readlines():
+  req = req.strip()
+  if req.startswith("numpy"):
+    setup_requirements.append(req)
```
vitaly-krugl: It's a shame that there isn't a cleaner standard way, and we're forced to have special-case logic for this.

PR #77, opened 2015-06-26
Add compiler dependency onto anaconda

unixorn:
Some of the packages we're installing in python by default require gcc to build.


unixorn:
@jcasner @shantanoo please CR

jcasner:
:+1:

PR #78, opened 2015-06-26
TAUR-1125: stock symbol renaming tool

jaredweiss:
@vitaly-krugl could you take a look at this when you get a chance


jaredweiss:
This tool doesn't deal with the issue of remonitoring the new metrics. **Before merging we should make a decision regarding this.** As of right now, metrics are manually activated in order to make sure we have a certain amount of records already recorded. 

We can either:
1. continue to just manually activate a metric when we have the appropriate number of records
2. create a separate tool that watches all unmonitored metrics and creates the models once a certain threshold of records is passed. a simple cronjob running on a server could do this. 
3. Have the ability to set the minimum number of rows a metric needs before it is automatically monitored. Right now this happens at 288 rows (not for taurus). In speaking with @vitaly-krugl this seems to be the best long-term option.

taurus.metric_collectors/setup.py
```diff
@@ -21,6 +21,7 @@
        "%s.twitterdirect.process_tweet_deletions:main" % name),
       ("taurus-purge-old-tweets = "
        "%s.twitterdirect.purge_old_tweets:main" % name),
+      "taurus-resymbol = %s.twitterdirect.resymbol_twitter_metric:main" % name,
```
oxtopus: This is a little confusing to me.  What about xignite stock metrics?

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -288,8 +297,38 @@ def unmonitorMetric(host, apiKey, modelId):
...
+  """ Delete a metric
+
+  :param host: API server's hostname or IP address
+  :param apiKey: API server's API Key
+  :param modelId: id of the metric to be deleted
+
+  :raises: RetriesExceededError
+  :raises: MetricDeleteRequestError
+  """
+  for _retries in xrange(20):
```
oxtopus: Are there already any retry decorators that we can use, say https://github.com/numenta/numenta-apps/blob/master/nta.utils/nta/utils/error_handling.py#L121-L237 ?

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -385,6 +424,44 @@ def getAllModelIds(host, apiKey):
...
+      modelResult.uid = uid of the model
+      modelResult.name = name of the model
+
+  :raises: GetModelsRequestError
+  :raises: RetriesExceededError
+  """
+  modelResult = namedtuple("modelResult", "uid name")
+  return tuple(modelResult(obj["uid"], obj["name"])
+               for obj in getAllModels(host, apiKey)
+               if ".%s." % symbol in obj["name"])
```
oxtopus: There's been discussion on the mailing list about use of `str.format()` in lieu of `%` formatting.

oxtopus:
:+1:  LGTM, only minor feedback and/or questions

taurus.metric_collectors/setup.py
jaredweiss: I'm not 100% sure whether or not we need to copy those over. xignite already has a small backlog of data so it may not be necessary. Do you think it's important to also copy the xignite data? If so it's a pretty easy implementation.

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
jaredweiss: I was just following the pattern set by the other functions in ```metric_utils```. I'm fine using one of our decorators if thats the better option but I think that would be better addressed in a separate PR dealing with cleaning up all functions in ```metric_utils```

jaredweiss:
Committed new changes to make resymboling tool ready to also migrate xignite data. Will add this function in a new PR

oxtopus:
:+1: LGTM

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -20,6 +20,7 @@
 # ----------------------------------------------------------------------
 
 import contextlib
+from collections import namedtuple
```
shantanoo: Move `collections` before `contextlib`.

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -0,0 +1,271 @@
...
+           epochFromNaiveUTCDatetime(oldestRecordTs)) / options.aggPeriod) * options.aggPeriod
+        aggStartDatetime = (lastEmittedAggTime -
+                            timedelta(seconds=aggOffset) -
+                            timedelta(seconds=options.aggPeriod))
+
+        metric_utils.updateLastEmittedSampleDatetime(
+            key=_EMITTED_TWEET_VOLUME_SAMPLE_TRACKER_KEY,
+            sampleDatetime=aggStartDatetime)
+
+        MetricDataForwarder.runInThread(
```
vitaly-krugl: @jaredweiss: `runInThread()` is not a clean entry point for what you needed to do here. This entry point was intended for a long-running service, hence `run...` in the name. It's designed to run indefinitely, waiting for aggregation rows to become available, aggregating/forwarding them, then waiting again ... forever. It's an aberration to take something like that and then add a hack that aborts it the first time through the loop. What you needed was find an existing method (or refactor existing code into one that's also used by the main loop) that sends a range of data samples and exists - something very much like `MetricDataForwarder._forwardTweetVolumeMetrics(...)`.

vitaly-krugl:
@jaredweiss: please update docstrings with detailed param entries (type/description) for the new args that you add to existing functions.

vitaly-krugl:
Also, you need to run pylint and fix pylint findings. At the minimum, I see a few statements that are too long in this commit; perhaps there are other issues as well that pylint might discover. Thx.

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -0,0 +1,271 @@
...
+                 "to new-symbol=%s",
+                 options.old_symbol, options.new_symbol)
+      raise NotImplementedError
+    else:
+      g_log.info("Migrating BOTH twitter and xignite stock data from "
+                 "old-symbol=%s to new-symbol=%s",
+                 options.old_symbol, options.new_symbol)
+      raise NotImplementedError
+
+    oldSymbolTweetPrefix = "TWITTER.TWEET.HANDLE.{symbol}.".format(symbol=options.old_symbol)
```
vitaly-krugl: @jaredweiss: it's a bad idea that the recipe for synthesizing the metric name got duplicated all over the place in this PR. Previously, it was present in production code only in `gen_metrics_config.py`. An acceptable approach would be to create a shared method (e.g., in `gen_metrics_config.py`) that takes a stock symbol and returns the tweet volume metric name that all of these places would call.

taurus.metric_collectors/taurus/metric_collectors/twitterdirect/migrate_tweets_to_dynamodb.py
```diff
@@ -87,11 +87,22 @@ def main():
...
+               sql.func.min(schema.twitterTweetSamples.c.seq),
+               sql.func.max(schema.twitterTweetSamples.c.seq)])
+             .where(schema.twitterTweetSamples.c.agg_ts >= startingTimestamp))
+    else:
+      sel = (sql.select(
+              [sql.func.count(),
+               sql.func.min(schema.twitterTweetSamples.c.seq),
+               sql.func.max(schema.twitterTweetSamples.c.seq)])
+             .where(schema.twitterTweetSamples.c.agg_ts >= startingTimestamp)
+             .where(sql.or_(
```
vitaly-krugl: @jaredweiss: `or_` is not documented to accept a generator. Are you sure that it would work this way? See http://docs.sqlalchemy.org/en/rel_1_0/core/sqlelement.html#sqlalchemy.sql.expression.or_.

PR #80, opened 2015-06-26
Add kernel version to /etc/motd

unixorn:
@shantanoo @jcasner please CR

### DIFF
--- a/infrastructure/saltcellar/motd/files/20-banner.motd
+++ b/infrastructure/saltcellar/motd/files/20-banner.motd
@@ -43,6 +43,9 @@ if [ -n "${salt_sha}" ]; then
   echo "Salt configuration SHA: ${salt_sha}"
 fi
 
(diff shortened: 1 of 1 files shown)


infrastructure/saltcellar/motd/files/20-banner.motd
```diff
@@ -43,6 +43,9 @@ if [ -n "${salt_sha}" ]; then
   echo "Salt configuration SHA: ${salt_sha}"
 fi
 
+# Display Kernel version
+echo "Kernel version: $(uname -r)"
```
jcasner: Why do we need this?

unixorn: Debugging is easier if you can see the kernel version, esp when you're working with docker

jcasner:
:+1:

PR #91, opened 2015-07-06
TAUR-1252: Fix "BadTokenException" when the background task is canceled #send-for-review

lscheinkman:
@jcasner Please review


jcasner:
:+1:

PR #93, opened 2015-07-06
DO NOT MERGE TAUR-1125: resymboling with xignite

jaredweiss:
Merge after https://github.com/numenta/numenta-apps/pull/78

Adds xignite metric migration to resymboling tool. 

@oxtopus Please review when you get a chance


taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -0,0 +1,327 @@
...
+            forwardOnlyBacklog=True)
+
+        metric_utils.updateLastEmittedSampleDatetime(
+            key=_EMITTED_TWEET_VOLUME_SAMPLE_TRACKER_KEY,
+            sampleDatetime=lastEmittedAggTime)
+
+      if options.stocks:
+        forwardStockBars(metricSpecs=[spec for spec
+                                      in loadStockBarsMetricSpecs()
+                                      if spec.symbol == 'AAPL'],
```
oxtopus: What's with the "AAPL"?

jaredweiss: @oxtopus Thanks for catching this. I was using 'AAPL' as a symbol to test with. I'll double check all my code and make sure none of it is left over from my original testing tomorrow morning

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -137,17 +149,32 @@ def _parseArgs():
...
     parser.error("Required \"--newsymbol\" option was not specified")
 
+
   if (not options.twitter) and (not options.stocks):
     parser.error("Flags specifying a single type of metric to migrate can "
                  "only be used exclusively. Forwarding all metrics is already "
                  "the default behavior of this tool.")
 
-  return options
+  optionsTuple = namedtuple("optionsTuple", "htmServer "
```
vitaly-krugl: Need to fix `_parseArgs` docstring: this no longer returns a dict

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -208,13 +235,54 @@ def main():
...
+        conn.execute(clearEmittedVolumeQuery)
+
+        updateStockHeadlineQuery = (stockHeadlineSchema
+                                    .update()
+                                    .where(stockHeadlineSchema.c.symbol ==
+                                           options.oldSymbol)
+                                    .values(symbol=options.newSymbol))
+        conn.execute(updateStockHeadlineQuery)
+
+        updateStockReleaseQuery = (stockReleaseSchema
```
vitaly-krugl: @jcasner: I recal. that we weren't going to do anything further to support news releases and headlines. I think that we should remove the news headlines/releases logic from here. Do you agree?

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -170,36 +197,36 @@ def main():
     g_log.info("Verifying that the old symbol has been removed from the "
                "metrics configuration")
     for stockData in metric_utils.getMetricsConfiguration().itervalues():
-      assert(stockData["symbol"] != options.old_symbol)
+      assert(stockData["symbol"] != options.oldSymbol)
```
vitaly-krugl: Should we also verify that the new symbol is present in metric specs?

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -170,36 +197,36 @@ def main():
...
                  "old-symbol=%s to new-symbol=%s",
-                 options.old_symbol, options.new_symbol)
-      raise NotImplementedError
+                 options.oldSymbol, options.newSymbol)
 
-    oldSymbolTweetPrefix = "TWITTER.TWEET.HANDLE.{symbol}.".format(symbol=options.old_symbol)
-    newSymbolTweetPrefix = "TWITTER.TWEET.HANDLE.{symbol}.".format(symbol=options.new_symbol)
+    oldSymbolTweetPrefix = "TWITTER.TWEET.HANDLE.{symbol}.".format(symbol=options.oldSymbol)
+    newSymbolTweetPrefix = "TWITTER.TWEET.HANDLE.{symbol}.".format(symbol=options.newSymbol)
     oldSymbolTweetMetricsList = []
```
vitaly-krugl: The code adds elements to `oldSymbolTweetMetricsList`, but then doesn't do anything with `oldSymbolTweetMetricsList`

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -170,36 +197,36 @@ def main():
...
       if options.twitter:
-        oldSymbolTweetsQuery = (sql.select([tweetSamplesSchema])
-                                .where(tweetSamplesSchema.c.metric
-                                       .contains(oldSymbolTweetPrefix)))
-        oldSymbolTweets = conn.execute(oldSymbolTweetsQuery)
-        for tweetSample in oldSymbolTweets:
+        oldSymbolTweetMetricsQuery = (sql.select([tweetSamplesSchema.c.metric
+                                                 .distinct()])
+                                      .where(tweetSamplesSchema.c.metric
+                                             .contains(oldSymbolTweetPrefix)))
```
vitaly-krugl: Shouldn't this be `.where(schema.twitterTweetSamples.c.metric == tweetSample.metric)`, instead ?

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -208,13 +235,54 @@ def main():
...
+                                   .update()
+                                   .where(stockReleaseSchema.c.symbol ==
+                                          options.oldSymbol)
+                                   .values(symbol=options.newSymbol))
+        conn.execute(updateStockReleaseQuery)
+
+
+      g_log.info("Forwarding new metric data to Taurus engine...")
       if options.twitter:
         oldestRecordTs = conn.execute(sql.select(
```
vitaly-krugl: What if there aren't any?

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -234,26 +302,33 @@ def main():
...
 
         MetricDataForwarder.runInThread(
-            metricSpecs=loadMetricSpecs(),
+            metricSpecs=loadTwitterMetricSpecs(),
             aggSec=options.aggPeriod,
-            symbolList=[options.new_symbol],
+            symbolList=[options.newSymbol],
             forwardOnlyBacklog=True)
 
         metric_utils.updateLastEmittedSampleDatetime(
```
vitaly-krugl: @jaredweiss: if something goes wrong between this call to `updateLastEmittedSampleDatetime` and the previous one, then all other models will get hosed, since your change (from prior PR) that calls updateLastEmittedSampleDatetime affects **all** twitter models.

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -234,26 +302,33 @@ def main():
             sampleDatetime=aggStartDatetime)
 
         MetricDataForwarder.runInThread(
-            metricSpecs=loadMetricSpecs(),
+            metricSpecs=loadTwitterMetricSpecs(),
             aggSec=options.aggPeriod,
-            symbolList=[options.new_symbol],
+            symbolList=[options.newSymbol],
```
vitaly-krugl: @jaredweiss: the symbolList and forwardOnlyBacklog changes ended up permeating throughout twitter_direct_agent.py code just to aggregate and forward some samples. This should not have been the case as it unnecessary complicated the code. It would have been acceptable to refactor a small section of the existing code there to simply expose a single function that aggregates and sends metric data samples for given metrics for a given range of time.

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -208,13 +235,54 @@ def main():
...
+                                           options.oldSymbol)
+                                    .values(symbol=options.newSymbol))
+        conn.execute(updateStockHeadlineQuery)
+
+        updateStockReleaseQuery = (stockReleaseSchema
+                                   .update()
+                                   .where(stockReleaseSchema.c.symbol ==
+                                          options.oldSymbol)
+                                   .values(symbol=options.newSymbol))
+        conn.execute(updateStockReleaseQuery)
```
vitaly-krugl: @jaredweiss: updateStockReleaseQuery and updateStockHeadlineQuery are executed, but what would be responsible for forwarding their metrics to Taurus Engine? Also, why are they tied to options.stocks?

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -234,26 +302,33 @@ def main():
...
-            symbolList=[options.new_symbol],
+            symbolList=[options.newSymbol],
             forwardOnlyBacklog=True)
 
         metric_utils.updateLastEmittedSampleDatetime(
             key=_EMITTED_TWEET_VOLUME_SAMPLE_TRACKER_KEY,
             sampleDatetime=lastEmittedAggTime)
 
+      if options.stocks:
+        forwardStockBars(metricSpecs=[spec for spec
```
vitaly-krugl: @jaredweiss: this call to forwardStockBars looks out of place. The preceding stock metric operations essentially deleted the old stock data, so what's there to be forwarded?

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -208,13 +235,54 @@ def main():
...
+                               .where(tweetSamplesSchema.c.metric
+                                      .contains(oldSymbolTweetPrefix))
                                .values(metric=newMetricName))
 
           conn.execute(updateSampleQuery)
 
-
-      g_log.info("Forwarding new twitter metric data to Taurus engine...")
+      if options.stocks:
+        renameStockQuery = (stockSchema
```
vitaly-krugl: @jaredweiss: this is a problem; each stockSchema row contains additional information about the security besides its symbol. So, when you update the symbol, the rest of the information becomes likely inconsistent with the symbol.

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
jaredweiss: @vitaly-krugl The reason I have those deletes ([like here](https://github.com/numenta/numenta-apps/pull/93/files#diff-d7e0c19a9e68f52493640149cfee5767R259)) is so that `_transmitMetricData` (imported in the tool as `forwardStockBars`) actually forwards the samples since they won't have been marked as sent ([see here](https://github.com/numenta/numenta-apps/blob/master/taurus.metric_collectors/taurus/metric_collectors/xignite/xignite_stock_agent.py#L556))

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -208,13 +235,54 @@ def main():
...
+                            .update()
+                            .where(stockSchema.c.symbol ==
+                                   options.oldSymbol)
+                            .values(symbol=options.newSymbol))
+        conn.execute(renameStockQuery)
+
+        updateStockBarsQuery = (stockBarsSchema
```
vitaly-krugl: @jaredweiss: stockBarsSchema table has a foreign key relationship on stockSchema symbol column with "on update cascade" constraint; so, I am guessing that the above statement accomplishes the renaming automatically.

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
vitaly-krugl: @jaredweiss: there is likely also a race condition here: when the stock agent is restarted per your "Renaming a stock (symbol)" instructions in the README, IIRC the agent begins to backfill the data for the new symbol automatically, which might be competing/conflicting with the stock metric workflow in this tool.

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
vitaly-krugl: @jaredweiss: In that case, wouldn't stock agent forward anything that's marked as "not sent" on its own when it enters ACTIVE mode, anyway?

jaredweiss: @vitaly-krugl Yes. I guess that this call isn't actually necessary since the data would be sent anyway.

vitaly-krugl: @jaredweiss: not picking on you, by the way :) Some of this knowledge is a bit esoteric.

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
jaredweiss: @vitaly-krugl to address your first point, yes, this tool wouldn't work if any of the other info associated with the symbol chaneged (i.e. market or name). I felt that that was outside of the scope of this task. 

And yes, there could be a race condition if the agent is restarted before the MySQL server finishes executing the query. I'm not quite sure if that's how SQLAlchemy + MySQL work, but if the execute command doesn't block until it's complete, there could definitely be race conditions.

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
jaredweiss: You wouldn't have tweetSample yet so you couldn't do that. The reason I used `tweetSamplesSchema.c.metric.contains(oldSymbolTweetPrefix)` was because getting the rows where the metric name has the symbol in it is the only way to get those samples. There's no key in the twitter_tweet_samples for just 'symbol'

vitaly-krugl: @jaredweiss: I was in the middle of refactoring the code, so the question made sense in that context, but you're right - tweetSample is not available yet at this point in your original code.

vitaly-krugl:
Closing this PR in favor of PR #149

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -208,13 +235,54 @@ def main():
...
+                                .update()
+                                .where(stockBarsSchema.c.symbol ==
+                                       options.oldSymbol)
+                                .values(symbol=options.newSymbol))
+        conn.execute(updateStockBarsQuery)
+
+        clearEmittedPriceQuery = (stockEmittedPriceSchema
+                                  .delete()
```
vitaly-krugl: @jaredweiss: this doesn't work properly due to a chain of cascading relationships: there is a foreign key relationship on symbol (among other columns) with a cascading update constraint between the {{stockEmitted*}} tables and the stock bars table. So, when the old symbols in stock bars got replaced with new symbols, the cascading relationships caused the symbols in the {{stockEmitted*}} tables to be updated as well. So trying to delete those rows at this stage using oldSymbol in the "where" clause will delete nothing since they have all been renamed already to newSymbol.

taurus.metric_collectors/taurus/metric_collectors/resymbol_metrics.py
```diff
@@ -234,26 +302,33 @@ def main():
...
     g_log.info("Unmonitoring and deleting existing metrics associated with "
-               "symbol=%s", options.old_symbol)
+               "symbol=%s", options.oldSymbol)
     oldModels = metric_utils.getSymbolModels(options.htmServer,
                                              options.apikey,
-                                             options.old_symbol)
+                                             options.oldSymbol)
     for model in oldModels:
       metric_utils.unmonitorMetric(options.htmServer, options.apikey, model.uid)
       metric_utils.deleteMetric(options.htmServer, options.apikey, model.name)
```
vitaly-krugl: @jaredweiss: This would only work if the old metrics were already promoted to a model. Otherwise, metric_utils .getSymbolModels would be missing those that weren't promoted, so we wouldn't be able to delete the metric.

Also, it's sufficient to delete just the metric; this deletes the model, too (if any)

PR #94, opened 2015-07-06
Fix some errors with setting up Grok

jcasner:


grok/README.rst
```diff
@@ -26,15 +26,9 @@ automatically with grok.
 
 ::
 
-    python setup.py develop --install-dir="$HOME/Library/Python/2.7/lib/python/site-packages"
+    ./install-grok.sh "$HOME/Library/Python/2.7/lib/python/site-packages" /usr/local/bin
```
vitaly-krugl: @jcasner: I think that this is not quite consistent. The `bin` location that parallels the given `lib` location is `$HOME/Library/Python/2.7/bin`. It's what you would get with if you passed `--user` to `pythone setup.py` or to `pip install`. Also, see the examples in https://github.com/numenta/numenta-apps/blob/master/taurus.metric_collectors/README.md.

jcasner:
@vitaly-krugl appreciate your feedback again. Thanks!

grok/README.rst
```diff
@@ -47,7 +43,7 @@ Bootstrap your Grok configuration files and initialize the database:
 generated Grok configuration files should be stored by `python setup.py init`. If not set,
 it presently defaults to the location of Grok configuraiton template files (e.g., `grok/conf/`)
 
-You'll need to install/start MySQL server prior to running setup.
+You'll need to install/start MySQL server and RabbitMQ server prior to running setup.
```
vitaly-krugl: prior to running `python setup.py init`. I am pretty sure that you don't need mysql and rabbitmq when running setup just to install the software.

vitaly-krugl:
LGTM after fixing "prior to running setup" as noted above.

PR #95, opened 2015-07-06
Miscellaneous fixes to Grok

jcasner:
Currently a WIP while I find and fix bugs in the current version of Grok


jcasner:
@rhyolight FYI, I found a couple requirements that were missing and have added those

@unixorn FYI, there are some pipeline changes in here

Still working through some issues, but appreciate the early feedback

grok/grok/app/aws/instance_utils.py
```diff
@@ -45,7 +45,7 @@ def getInstanceData():
     }
 
   """
-  url = 'http://instance-data/latest/dynamic/instance-identity/document'
+  url = "http://169.254.169.254/latest/dynamic/instance-identity/document"
```
unixorn: I'd prefer setting a `EC2METADATA` constant and setting it to 169.254.169.254 instead of using a raw IP.

jcasner: The raw IP is a constant from AWS, so I could go either way.  Merged this for now to get it in, but we can update to a constant later if we decide to

unixorn: Let's leave it until/unless we use another bit of instance data. Once we're reading more than one chunk of instance data I'd prefer to use a constant if only to prevent typos in the raw IP.

oxtopus: Why change it at all?

grok/DEPENDENCIES.md
```diff
@@ -16,7 +16,9 @@ fabric | 1.9.0 | [License](https://github.com/fabric/fabric/blob/master/LICENSE)
 msgpack-python | 0.3.0 | [Apache 2.0](https://github.com/msgpack/msgpack-python/blob/0.3.0/COPYING)
 MySQLdb1 (imported as `mysql-python`) | 1.2.5 | [Python 1.5.2 License](https://docs.python.org/2/license.html) via [MySQLdb-1.2.5/INSTALL](https://github.com/farcepest/MySQLdb1/blob/MySQLdb-1.2.5/INSTALL#L249)
 paste | 1.7.5.1 | [MIT license](http://pythonpaste.org/)
+prettytable | 0.7.2 | [New BSD License](https://code.google.com/p/prettytable/)
```
oxtopus: prettytable is not a dependency of grok.

On that note, we should move https://github.com/GrokSolutions/grok-cli into https://github.com/numenta/numenta-apps

grok/DEPENDENCIES.md
```diff
@@ -16,7 +16,9 @@ fabric | 1.9.0 | [License](https://github.com/fabric/fabric/blob/master/LICENSE)
...
 pytz | 2015.2 | [MIT License](http://pythonhosted.org/pytz/)
+PyYAML | 3.10 | [License](http://pyyaml.org/browser/pyyaml/tags/3.10/LICENSE)
```
oxtopus: Same for PyYAML.

grok/grok/app/aws/instance_utils.py
jcasner: I'm not sure why it was set to `instance-data` in the first place @oxtopus, but all of the AWS documentation lists it as `169.254.169.254`.  Perhaps your question was to Joe?  This IP address is well advertised by AWS, so it shouldn't need to change ever...

grok/DEPENDENCIES.md
jcasner: They're both used by tests and the pipeline... @rhyolight what's the right approach here?

unixorn:
I prefer human readable constant names to raw ips, even if they're well publicized. It is just a lot easier for humans to typo an IP than a name.

grok/DEPENDENCIES.md
oxtopus: yaml is used:

```
$ git grep "import yaml"
grok-mobile/run_pipeline:import yaml
grok/behavioral_tests_driver:import yaml
grok/grok/pipeline/src/bake_ami.py:import yaml
grok/grok/pipeline/src/package.py:import yaml
grok/grok/pipeline/src/run_grok_integration_tests.py:import yaml
grok/grok/pipeline/src/sync_yum_repo_s3.py:import yaml
grok/tests/py/integration/app/adapters/datasource/cloudwatch/cw_datasource_adapter_base_test.py:import yaml
grok/tests/py/integration/app/adapters/datasource/cloudwatch/cw_datasource_metric_specific_test.py:import yaml
infrastructure/infrastructure/utilities/infrastructure-rpm-pipeline/run_pipeline:import yaml
infrastructure/infrastructure/utilities/nupic/build_commands.py:import yaml
infrastructure/infrastructure/utilities/nupic/run_pipeline:import yaml
```

prettytable is not.  I can only assume that the prettytable requirement comes from grokcli.

grok/grok/app/aws/instance_utils.py
oxtopus: Personally, I'm in favor of `169.254.169.254`.  It sticks out like a sore :+1:, and is well advertised as a special endpoint for instance metadata.

I do recall, however, some back-and-forth about the use of "instance-data".  There's a difference in behavior on local dev vs EC2 wrt DNS resolution and how it's handled in the `requests` library and other exception handling scenarios.

grok/DEPENDENCIES.md
rhyolight: > They're both used by tests and the pipeline... @rhyolight what's the right approach here?

Only include production dependencies, not test or pipeline.

jcasner: `prettytable` +was+ a requirement, but I removed it in this PR, so I'll update the reference in DEPENDENCIES.md and requirements.txt (https://github.com/numenta/numenta-apps/pull/95/files#diff-20b0df09be166a4c59c2551083de0298L277)

PR #96, opened 2015-07-06
Fix issues with mysql formula

unixorn:
* Make the state names generic so we can change where we get the RPMs from more easily. This should make the CentOS 7 upgrade less painful.
* Fix parse error in mysql.server.sls


unixorn:
@jcasner @shantanoo please CR

jcasner:
:+1:

PR #99, opened 2015-07-07
TAUR-708 Fixup last remaining grok namespace dependencies in lower-level packages

oxtopus:
This should enable htmengine, nta.utils, and taurus tests to pass where grok is not installed.


jcasner:
:+1:

PR #107, opened 2015-07-09
Converted while loop to for

shantanoo:
In `while` loop, `fileList` was updated in each iteration. Changed it to `for` loop.


shantanoo:
Cc: @jcasner, @unixorn

unixorn:
@shantanoo This was an import of code that got missed in the chef->salt transition before we open sourced numenta-apps. I mainly polished up the docstrings and removed some unused imports, but here's still a lot of room for improvement :smile:

jcasner:
:+1:

PR #110, opened 2015-07-09
TAUR-1292 fall back to git rev-parse if no grok.sha file exists

jcasner:
@brev @unixorn please review

### DIFF
--- a/grok/grok/app/webservices/webapp.py
+++ b/grok/grok/app/webservices/webapp.py
@@ -103,9 +103,7 @@ except IOError:
   try:
     buildSha = check_output(["git", "rev-parse", "--verify", "HEAD"])
   except CalledProcessError:
-    # Do nothing; just set to 0 as below
-    pass
-  buildSha = "0"
+    buildSha = "0"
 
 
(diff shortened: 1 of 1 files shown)


brev:
:+1:

grok/grok/app/webservices/webapp.py
```diff
@@ -99,7 +100,12 @@
 try:
   buildSha = open(os.path.join(grok.app.GROK_HOME, "static/grok.sha")).read()
 except IOError:
-  buildSha = '0'
+  try:
+    buildSha = check_output(["git", "rev-parse", "--verify", "HEAD"])
+  except CalledProcessError:
+    # Do nothing; just set to 0 as below
+    pass
+  buildSha = "0"
```
unixorn: I think this needs to be inside the except clause or we'll always reset to 0 even if git rev-parse returns a sha

jcasner: Good catch!

jcasner:
@unixorn updated with your feedback

unixorn:
:+1:

PR #112, opened 2015-07-10
Restart mysqld service when we turn on the slow query log

unixorn:
We weren't guaranteeing a restart of the mysqld server when we update `/etc/my.cnf`

It is bad practice to name the state after the config file path instead of using a symbolic name, so fixed that too.

@jcasner @shantanoo please CR


shantanoo:
:+1:

PR #113, opened 2015-07-10
Stop zsh and bash from complaining about no ssh-agent

unixorn:
Make login less spammy. We now only show ssh keys when they're actually present.

This also squelches console spam when the grok init script drops privileges to ec2-user during service startup.

@shantanoo @jcasner please CR.

### DIFF
--- a/infrastructure/saltcellar/dotfiles/files/bash_profile_skeleton.sh
+++ b/infrastructure/saltcellar/dotfiles/files/bash_profile_skeleton.sh
@@ -114,8 +114,9 @@ if [ -f ~/.sh_aliases ]; then
   source ~/.sh_aliases
 fi
 
-df -h
-
-echo
-echo "SSH keys:"
-ssh-add -l
+SSHkeylist=$(ssh-add -l 2>&1)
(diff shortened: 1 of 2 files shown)


infrastructure/saltcellar/dotfiles/files/bash_profile_skeleton.sh
```diff
@@ -114,8 +114,9 @@ if [ -f ~/.sh_aliases ]; then
   source ~/.sh_aliases
 fi
 
-df -h
-
-echo
-echo "SSH keys:"
-ssh-add -l
+ssh-add -l &> /dev/null
```
shantanoo: Instead of running the `ssh-add -l` command twice, we can modify this to

```
output=$(ssh-add -l)
if [[ $? == "0" ]]; then
  echo
  echo "SSH keys:"
  echo ${output}
fi
```

unixorn:
@shantanoo Good catch, I fixed it.

@jcasner @shantanoo, I'm ready for another CR

shantanoo:
LGTM.

PR #116, opened 2015-07-10
TAUR-1296 fix some issues with pylint in grok/app/adapters

jcasner:
@vitaly-krugl please review

### DIFF
--- a/grok/grok/app/adapters/datasource/cloudwatch/aws_autoscaling_group.py
+++ b/grok/grok/app/adapters/datasource/cloudwatch/aws_autoscaling_group.py
@@ -20,8 +20,7 @@
 # ----------------------------------------------------------------------
 
 # Note: we don't allow autostacks of ASGs, so disable warnings about
-# `getMatchingResources` not being implemented
-#pylint: disable=W0223
+# `getMatchingResources` not being implemented (disable=W0223 comments below)
 
 from boto.ec2 import autoscale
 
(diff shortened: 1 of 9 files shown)


grok/grok/app/adapters/datasource/cloudwatch/aws_autoscaling_group.py
```diff
@@ -19,6 +19,10 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
+# Note: we don't allow autostacks of ASGs, so disable warnings about
+# `getMatchingResources` not being implemented
+#pylint: disable=W0223
```
vitaly-krugl: @jcasner, I am not crazy about disabling these warnings at global level, especially in production code. This will prevent pylint from helping us detect other symbols that we accidentally failed to override.

oxtopus: Why not implement `getMatchingResources()` as a no-op, including a comment explaining why?

jcasner:
@oxtopus I just pushed my changes to address @vitaly-krugl 's feedback; please let me know if this is a better approach.  Thanks!

oxtopus:
:+1: better.  Thanks!

grok/grok/app/adapters/datasource/cloudwatch/aws_autoscaling_group.py
jcasner: @oxtopus I could certainly do that; I'm a little concerned that doing so will mask an issue later.  Given it's a no-op, having it raise `NotImplemented` makes a lot of sense...

vitaly-krugl:
@oxtopus and @jcasner: it looks like an issue with pylint, but the way; it's complaining on reference to a fully-implemented static method in an abstract base class. I don't think that there is a rule about not accessing members of ABC's. It should only apply when trying to instantiate an ABC.

PR #117, opened 2015-07-11
TAUR-1297 fix pylint issues in grok/app/repository

jcasner:
@vitaly-krugl @oxtopus @scottpurdy @lscheinkman  please review


scottpurdy:
This looks fine to me although I'm not sure about `EngineSingleton`. Did that get moved somewhere else or does our current usage of SQLAlchemy no longer use it? @vitaly-krugl ?

vitaly-krugl:
@scottpurdy: engine singleton is shared via `from htmengine.repository import _EngineSingleton`

vitaly-krugl:
Disabling pylint warnings/errors at global level is a bad idea, because it will cause illegitimate issues to be missed. This also has a way of infecting other files (copy/paste and sanctioning of approach) to a point where pylint becomes ineffective. I've seen this happen elsewhere to a detrimental outcome, so believe my conclusion to be on the mark.

jcasner:
@vitaly-krugl there are 2 issues present here:

1. We import a lot of queries from `htmengine.repository.queries` so that we can use them downstream with the Grok code agnostic as to whether they are defined in `htmengine.repository.queries` or `grok.app.repository.queries`.  See https://github.com/jcasner/numenta-apps/blob/TAUR-1297/grok/grok/app/repository/queries.py#L46-L84 If we don't disable `W0611` globally in the file, then we either get a `Locally disabling W0611 (locally-disabled)` message for each of those lines _OR_ we see pylint fail for each line.  In either case, it makes the pylint output unreadable.
1. There is a bug in pylint that causes erroneous error messages when using sqlalchemy. Details are available here: https://github.com/jcasner/numenta-apps/blob/cae9664e9126333708645d1bf2db384b512a20d6/grok/grok/app/repository/queries.py#L39-L42 

I'm not sure what the right approach is... Open to thoughts and would love to get this right.

CC @oxtopus @scottpurdy

jcasner:
@vitaly-krugl I have addressed your comments; please take another look.

grok/grok/app/repository/queries.py
```diff
@@ -447,7 +454,9 @@ def getAutostackMetricsWithMetricName(conn, autostackId, name, fields=None):
 
 
 
-def _getCloudwatchMetricReadinessPredicate(conn):
+# TODO Determine why `conn` is used as a parameter and either delete it or use
+# it. Until then, disable pylint warning
+def _getCloudwatchMetricReadinessPredicate(conn): #pylint: disable=W0613
```
oxtopus: We should either fix this now, or include a concrete JIRA in the TODO for removal.

vitaly-krugl: Let's get rid of `conn` arg in this method.

jcasner: Removed

vitaly-krugl:
LGTM

PR #118, opened 2015-07-13
TAUR-1300 update setuptools version

jcasner:
@oxtopus this should get the right version of setuptools for you.  Not sure why we were version locked to such an old version.


oxtopus:
:+1:

PR #120, opened 2015-07-13
TAUR-834 Don't retrieve EC2 metadata if running on Mac OS.

vitaly-krugl:
Don't retrieve EC2 metadata if running on Mac OS. This removes the significant delay that was occurring when loading the Grok GUI on Mac OS X as well as anything else that used grok/app/aws/instance_utils.py

### DIFF
--- a/grok/grok/app/aws/instance_utils.py
+++ b/grok/grok/app/aws/instance_utils.py
@@ -18,6 +18,8 @@
 #
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
+import sys
+
 import requests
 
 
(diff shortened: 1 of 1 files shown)


vitaly-krugl:
CC @oxtopus

oxtopus:
I disagree with this approach.  You should be able to legitimately run this within linux not inside of EC2.  So now there's a difference in behavior between OS X and Linux.  Better to detect whether or not we're specifically within EC2 environment, and best way I know is via instance metadata endpoint.

vitaly-krugl:
I agree that this is not ideal, and I would prefer to have a deterministic way to check for EC2 without paying the prohibitive 80-second delay penalty, but we don't have such a mechanism right now, and this change unblocked my efforts to debug/fix TAUR-834 on the Mac.

vitaly-krugl:
I tried the solution suggested by @oxtopus of using `socket.gethostbyaddr` to check 169.254.169.254 before connecting to it, but that still delays for 5 or more seconds (depending on platform). The `host` command appears to be near-instant (on my laptop, but is not portable. So, I am going to leave the darwin check in place until we have a better solution.

PR #123, opened 2015-07-13
hack to get the pipeline moving

jcasner:
@shantanoo FYI


oxtopus:
:-1:  This is failing due to an old version of setuptools.  Need to install the newer setuptools, and if that's still failing we have a bigger problem.

jcasner:
@oxtopus totally agree; this should not be merged. @shantanoo needed to have a temporary way to get the pipeline to the point where it fails in Salt.  

FYI, this is actually a different issue than the setuptools issue you are referencing, also...

PR #124, opened 2015-07-13
TAUR-834 Extend _AutostackDatasourceAdapter.getInstanceNameForModelSpec to support import-model-spec

vitaly-krugl:
If import-model-spec is passed to _AutostackDatasourceAdapter.getInstanceNameForModelSpec, attempt to map the autostack name/region to autostack id. If autstack doesn't exist, return None as permitted by DatasourceAdapterIface.getInstanceNameForModelSpec


vitaly-krugl:
Please review

CC @scottpurdy

scottpurdy:
:+1:

PR #125, opened 2015-07-14
TAUR-1284 Increase timeout in checkModelDeleted() from 25 to 60

oxtopus:
@jcasner This causes the test to pass in the slower instance type. re: https://jira.numenta.com/browse/TAUR-1284


jcasner:
This is going to potentially more than double the execution duration of this test.  Is there a smaller number that we can use that will be successful?

oxtopus:
No, that is not the case.  It makes an attempt once per second.  This `duration` kwarg is the upper limit.

jcasner:
Aha, gotcha - thanks for the clarification!  In which case, :+1: :ship: it

PR #127, opened 2015-07-16
TAUR-873 Fixed incorrect SP configuration property names

vitaly-krugl:
This issue was introduced in https://github.com/GrokSolutions/products/pull/768. The fix renames "columnDimensions" -> "potentialPct" in htmengine/algorithms/modelSelection parameter files.

NOTE that this change does not fully address TAUR-873. There are also issues in nupic that separately contribute to the failure.


vitaly-krugl:
CC @oxtopus, @scottpurdy

scottpurdy:
:+1:

PR #128, opened 2015-07-16
TAUR-873 Log scores in result_quality_test.py when the test passes to help with debugging and comparisons

vitaly-krugl:


vitaly-krugl:
CC @scottpurdy

scottpurdy:
I find the new function a little confusing. Why not leave as is and just concatenate either the success or the failure message with the common part depending on the two cases? But otherwise seems fine.

vitaly-krugl:
@scottpurdy, that would work, too. Is it worthwhile to make the suggested change?

scottpurdy:
@vitaly-krugl - your call

PR #129, opened 2015-07-16
TAUR-873 Retune expected results to make result-quality-test pass.

vitaly-krugl:
@scottpurdy: I am not too happy with a change like this as some of the individual results for one of the tests seem a lot worse than in the past. I am submitting the PR to help with discussion.


scottpurdy:
Results are dependent on the random seed so changes that don't have a material impact overall may have some small positive or negative impact on this tests score. In this case it is mostly positive and overall a very small change so I think we are safe. We know part of the change was a negative hit from changing the random number generator in the encoder. The other part, which is a positive change to the results, is unknown but would likely take a lot of work to track down so I think we can just update the numbers and focus on making sure we identify changes soon, with something like #2322

:+1:

PR #131, opened 2015-07-20
TAUR-1319 ensure the correct version of mock gets installed for Taurus

jcasner:
@oxtopus please review

### DIFF
--- a/nta.utils/requirements.txt
+++ b/nta.utils/requirements.txt
@@ -1,7 +1,9 @@
 boto==2.38.0
 mysql-python==1.2.5
-mock
 -e https://github.com/vitaly-krugl/haigha/tarball/temp-numenta#egg=haigha-0.7.4rc100
 pytz==2015.2
 sqlalchemy==0.9.4
+
+# required by tests
+mock==1.0.1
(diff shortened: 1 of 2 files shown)


oxtopus:
:+1: however, we need to do the same to `nta.utils` for consistency.

jcasner:
@oxtopus made the relevant change to nta.utils.  Good catch!

nta.utils/requirements.txt
```diff
@@ -1,7 +1,9 @@
 boto==2.38.0
 mysql-python==1.2.5
-mock
 -e https://github.com/vitaly-krugl/haigha/tarball/temp-numenta#egg=haigha-0.7.4rc100
 pytz==2015.2
 sqlalchemy==0.9.4
+
+# required by tests
+mock==1.0.1
 unittest2==0.5.1
```
vitaly-krugl: Not related to this PR, but we should get rid of `unittest2` references from `numenta-apps`, because `numenta-apps` is a python 2.7.x project, and the builtin `unittest` in python 2.7.x has the necessary features. `unittest2` was needed for python 2.6.x.

PR #133, opened 2015-07-20
TAUR-1313 move result quality test

jaredweiss:
Moved `result_quality_test.py` from `grok/tests/py/integration/app/` to `htmengine/tests/integration`. 

Also changed the test to use sqlalchemy queries when verifying anomaly results as opposed to API calls. htmengine itself doesn't have a defined web api so it made sense to simply use only functionality built into the htmengine.


oxtopus:
:+1:

PR #134, opened 2015-07-20
Declare model results exchange as the more flexible topic exchange

oxtopus:
This also adds a new CLI option "--migrate" to htmengine.runtime.anomaly_service so that this, and other future migrations may be applied by calling `python -m htmengine.runtime.anomaly_service --migrate`.

Not tied to any specific issue, but this is in response to working on https://github.com/nupic-community/skeleton-htmengine-app in which I wanted to demonstrate how to synchronously receive results for a specific model, which is not possible with a fanout exchange.  This change is also anticipatory for potential future requirements in a distributed systems setting.

cc @vitaly-krugl

### DIFF
--- a/grok/grok/app/runtime/notification_service.py
+++ b/grok/grok/app/runtime/notification_service.py
@@ -153,7 +153,7 @@ class NotificationService(object):
   """ Notification Service to monitor model inference results and trigger
       notifications where appropriate.
 
-      Binds a "notifications" queue to the model results fanout exchange
+      Binds a "notifications" queue to the model results topic exchange
       defined in the ``results_exchange_name`` configuration directive of the
       ``metric_streamer`` configuration section.
   """
@@ -382,13 +382,13 @@ class NotificationService(object):
         # make sure the queue and exchanges exists and the queue is bound
         amqpClient.declareExchange(self._modelResultsExchange,
                                    durable=True,
-                                   exchangeType="fanout")
+                                   exchangeType="topic")
 
         result = amqpClient.declareQueue("notifications", durable=True)
 
         amqpClient.bindQueue(queue=result.queue,
                              exchange=self._modelResultsExchange,
-                             routingKey="")
+                             routingKey="#")
 
         # Start consuming messages
         consumer = amqpClient.createConsumer(result.queue)
--- a/htmengine/htmengine/runtime/anomaly_service.py
+++ b/htmengine/htmengine/runtime/anomaly_service.py
@@ -23,7 +23,6 @@
 import itertools
 import json
 import logging
-from collections import namedtuple
 import math
 from optparse import OptionParser
 import os
@@ -83,6 +82,17 @@ def _getLogger():
 
 
 g_log = _getLogger()
+
+
+
+def _migrateResultsExchangeFromFanoutToTopic(modelResultsExchange):
+  # Results exchange was previously a fanout exchange, delete it so that it may
+  # be re-declared later as a topic exchange
+  with amqp.synchronous_amqp_client.SynchronousAmqpClient(
+      amqp.connection.getRabbitmqConnectionParameters()) as amqpClient:
+    g_log.info("Deleting {} exchange.".format(modelResultsExchange))
+    amqpClient.deleteExchange(modelResultsExchange, ifUnused=False)
+    g_log.info("{} exchange deleted.".format(modelResultsExchange))
 
 
 
@@ -169,7 +179,7 @@ class AnomalyService(object):
   configuration directive from the ``metric_streamer`` section of
   ``config``.
 
-  Other services may be subscribed to the model results fanout exchange for
+  Other services may be subscribed to the model results topic exchange for
   subsequent (and parallel) processing.  For example,
   ``htmengine.runtime.notification_service.NotificationService`` is one example
   of a use-case for that exchange.  Consumers must deserialize inbound messages
@@ -191,6 +201,12 @@ class AnomalyService(object):
       config.getint("anomaly_likelihood", "statistics_sample_size"))
 
     self.likelihoodHelper = AnomalyLikelihoodHelper(self._log, config)
+
+
+  def _migrate(self):
+    """ Migrations go here.
+    """
+    _migrateResultsExchangeFromFanoutToTopic(self._modelResultsExchange)
 
 
   def _processModelCommandResult(self, metricID, result):
@@ -648,7 +664,7 @@ class AnomalyService(object):
     with amqp.synchronous_amqp_client.SynchronousAmqpClient(
         amqp.connection.getRabbitmqConnectionParameters()) as amqpClient:
       amqpClient.declareExchange(self._modelResultsExchange,
-                                 exchangeType="fanout",
+                                 exchangeType="topic",
                                  durable=True)
 
     with ModelSwapperInterface() as modelSwapper, MessageBusConnector() as bus:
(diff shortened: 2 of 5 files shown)


grok/grok/app/runtime/notification_service.py
```diff
@@ -382,13 +382,13 @@ def configChannel(amqpClient):
         # make sure the queue and exchanges exists and the queue is bound
         amqpClient.declareExchange(self._modelResultsExchange,
                                    durable=True,
-                                   exchangeType="fanout")
+                                   exchangeType="topic")
```
vitaly-krugl: I agree: a "topic" or "direct" exchange makes more sense than "fanout" for our use case.

htmengine/htmengine/runtime/anomaly_service.py
```diff
@@ -193,6 +203,12 @@ def __init__(self):
     self.likelihoodHelper = AnomalyLikelihoodHelper(self._log, config)
 
 
+  def _migrate(self):
+    """ Migrations go here.
```
vitaly-krugl: The docstring comment and function name are not clear about what would be migrated.

htmengine/htmengine/runtime/anomaly_service.py
```diff
@@ -742,15 +758,30 @@ def main(args):
 
   parser = OptionParser(helpString)
 
-  (_options, args) = parser.parse_args(args)
+  parser.add_option("--migrate",
```
vitaly-krugl: Would `--migrate-amqp` be a better, less ambiguous,  name for this option?

htmengine/htmengine/runtime/anomaly_service.py
```diff
@@ -193,6 +203,12 @@ def __init__(self):
     self.likelihoodHelper = AnomalyLikelihoodHelper(self._log, config)
 
 
+  def _migrate(self):
+    """ Migrations go here.
+    """
+    _migrateResultsExchangeFromFanoutToTopic(self._modelResultsExchange)
```
vitaly-krugl: Since `_migrateResultsExchangeFromFanoutToTopic` is an instance member function and its name is specific to the "model results" exchange, there is no need to pass `self._modelResultsExchange` to it.

htmengine/htmengine/runtime/anomaly_service.py
```diff
@@ -673,7 +689,7 @@ def run(self):
                 else:
                   bus.publishExg(
                     exchange=self._modelResultsExchange,
-                    routingKey="",
+                    routingKey="modelID.{}".format(batch.modelID),
```
vitaly-krugl: Anomaly service presently publishes two types of data to this exchange: model command results (here) and model inference results. 

Here are a few notables:
1. The call that publishes model inference results has not been updated with an explicit routing key. Is this intentional?
2. If this routing key is intended to distinguish model command results from other data published on this exchanged, then it should contain some substring linking it to "model command result". This way, a service that is only interested in model command results could use a topic binding pattern to get only those messages.
3. What is the anticipated use of model ID in `"modelID.{}".format(batch.modelID)`?

grok/grok/app/runtime/notification_service.py
```diff
@@ -382,13 +382,13 @@ def configChannel(amqpClient):
...
 
         result = amqpClient.declareQueue("notifications", durable=True)
 
         amqpClient.bindQueue(queue=result.queue,
                              exchange=self._modelResultsExchange,
-                             routingKey="")
+                             routingKey="#")
```
vitaly-krugl: Since we now use a topic exchange, we could do better than `"#"` here in order to get exactly what notification service needs. In particular, it cares only about inference results. Once the routing key is fixed to select just the inference results, we could get rid of this snippet from notification service:

```
    if message.properties.headers and "dataType" in message.properties.headers:
      # Not a model inference result
      return
```

vitaly-krugl:
@oxtopus: I reviewed this PR and created feedback for you.

htmengine/htmengine/runtime/anomaly_service.py
oxtopus: `_migrateResultsExchangeFromFanoutToTopic()` is not an instance member function.  It is at module level.

htmengine/htmengine/runtime/anomaly_service.py
oxtopus: I figured the class would have a generic `_migrate()` entry point which would call out to the specific migrations.  Hence, the docstrings of the individual migrations are more descriptive (see` _migrateResultsExchangeFromFanoutToTopic()`).

vitaly-krugl: I meant that it's not clear that it's related to AMQP

htmengine/htmengine/runtime/anomaly_service.py
vitaly-krugl: Oh, you're right about _migrateResultsExchangeFromFanoutToTopic.

htmengine/htmengine/runtime/anomaly_service.py
oxtopus: > The call that publishes model inference results has not been updated with an explicit routing key. Is this intentional?
> If this routing key is intended to distinguish model command results from other data published on this exchanged, then it should contain some substring linking it to "model command result". This way, a service that is only interested in model command results could use a topic binding pattern to get only those messages.

Nope.  My mistake.  I missed the separation between the two.  How about prefixing the model command results with "modelCommandResult." and the model inference results with "modelInferenceResult"?

> What is the anticipated use of model ID in "modelID.{}".format(batch.modelID)?

The impetus for this was that in creating https://github.com/nupic-community/skeleton-htmengine-app I saw the need for a use-case for a single process to handle the full cycle for one, or more _specific_ models.  With the current fanout scheme it's not possible as you have to subscribe to everything and drop what you don't need.  By using a topic exchange and routing keys, it's possible to bind a queue to the results exchange by a routing key that matches only that model, ignoring everything else.  In Taurus and Grok, for now, we're ignoring this feature and binding to the '#' wildcard to preserve existing functionality.

htmengine/htmengine/runtime/anomaly_service.py
oxtopus: I think `--migrate` is sufficient.

htmengine/htmengine/runtime/anomaly_service.py
oxtopus: It doesn't need to be exclusive to AMQP in the long term.  In practice, this function calls another function that is specific to amqp, but I don't think it's relevant to the higher-level interface.

grok/grok/app/runtime/notification_service.py
oxtopus: Good point.  Thanks!

htmengine/htmengine/runtime/anomaly_service.py
vitaly-krugl: "modelCommandResult" and "modelInferenceResult" or something shorter would work.

vitaly-krugl: > The impetus for this was that in creating .. handle the full cycle for one, or more specific models
Makes sense

jcasner:
@oxtopus @vitaly-krugl can we get this one cleaned up and / or closed out please?

jcasner:
@oxtopus @vitaly-krugl I'm going to close this stale PR. We can revive it later as and when it makes sense.

oxtopus:
No problem.  I'm also considering creating an exchange-exchange binding for my use-case, which would be non-intrusive.

Ultimately, a change like this is the right way to go, but not a priority at the moment.

PR #136, opened 2015-07-21
TAUR-1245: Make sure to apply log scale matching the UI behavior before firing notification

lscheinkman:
@jcasner Please review


jcasner:
:+1:

PR #137, opened 2015-07-21
TAUR-1314 remove application dependencies result quality test

jaredweiss:
Pending merging of #133 

Removes dependencies `result_quality_test.py` has on grok or taurus running and provides a method for running integration tests in htmengine.


jaredweiss:
@scottpurdy @oxtopus or @vitaly-krugl please review this if you have a chance.

scottpurdy:
Looks fine to me if tests all still pass

oxtopus:
I'm hesitant to intermangle these new configuration files like this, so as to not confuse the roles between htmengine (framework) and application.

htmengine/conf/supervisord.conf
```diff
@@ -0,0 +1,36 @@
...
+; variables can be expanded using this syntax: "%(ENV_HOME)s".
+
+[unix_http_server]
+file=%(here)s/../htmengine-supervisor.sock   ; (the path to the socket file)
+
+[inet_http_server]
+port=127.0.0.1:9001
+
+[supervisord]
+environment=APPLICATION_CONFIG_PATH=/Users/jweiss/nta/applications/htmengine/conf
```
oxtopus: `/Users/jweiss`...

vitaly-krugl:
I didn't dive into this deeply, but HTM Engine does need to maintain functional "templates" for configuration. Following initial extraction of HTM Engine from Grok, we had Grok and Taurus configuration files that included HTM Engine settings as well as their own, and there was nothing in place to delineate what's required by HTM Engine.

jaredweiss:
@oxtopus @vitaly-krugl 
The scope of this task was to make it so the HTMEngine integration tests can be run without having Grok or Taurus running (see TAUR-1314). What would our other options be for having the result_quality_test be runnable without having Grok/Taurus running other than having a standalone HTMEngine application? Or is this task's description not accurate anymore?

oxtopus:
Add the requisite application implementation somewhere obvious for it to be part of the test, say, `htmengine/tests/integration/support`, or `htmengine/result-quality-test-application`

vitaly-krugl:
HTM Engine needs to be able to stand up on its own to support any HTM Engine integration tests, not just result-quality-test. I recommend not tying its name to result-quality-test. This would also server as an example of how to stand up HTM Engine components in general.

oxtopus:
I disagree with the overall sentiment that all integration tests should be able to run independent of an installed application.  It defeats the purpose of an integration test, IMHO.  The result quality test is something of a special case, however, and may better be classified differently.

vitaly-krugl:
@oxtopus, HTM Engine is a component that includes a set of services that need to be fully tested through both integration and unit tests on their own. It should not rely on being embedded in Taurus or Grok or any other specific product in order to test it. Any app considering or choosing to embed HTM Engine should have the full confidence that HTM Engine is reliable and passes its own integration and unit tests, without side-effects of and dependencies on other specific applications.

Tests that verify live functionality of HTM Engine's components on their own as well as the integration between them (Metric listener, storer, scheduler and anomaly service) are integration tests.

oxtopus:
htmengine is not a _standalone_ component.  I agree that there should be a reasonable expectation that htmengine has a batteries-included approach to testing, but there's no point in bundling an arbitrary app for integration testing purposes if that app is not representative of the application in which htmengine is intended to be used, requiring the end user to duplicate the integration tests to verify that everything does indeed work together.

scottpurdy:
@vitaly-krugl @oxtopus - Sounds like everyone is on the same page - we should have tests for `htmengine` to validate its functionality but app-specific tests for Taurus, etc. belong in the corresponding project.

I think result_quality_test makes sense in `htmengine` because it is validating that we don't have result quality regressions on `htmengine`-like models. It is too specific for `nupic` and too general for `taurus` or `grok`. In other words, it applies to all users of HTM Engine but not all users of NuPIC.

I don't think that conflicts with what either of you are saying but please follow up if it does.

vitaly-krugl:
@scottpurdy: makes sense.

jaredweiss:
@vitaly-krugl @oxtopus @scottpurdy 

Please review again

htmengine/README.md
```diff
@@ -97,3 +97,36 @@ example:
     [include]
     files = ../../htmengine/conf/supervisord-base.conf
 
+Running HTMEngine Integration Tests
+-----------------------------------
+
+The HTM engine has a basic skeleton application exclusively for running tests.
+To run the tests, you must have MySQL, RabbitMQ, and Supervisor running.
+
+Make sure that `APPLICATION_CONFIG_PATH` is set to point to tests/
```
vitaly-krugl: "is set to point to tests/" doesn't match the location of the conf folder

htmengine/README.md
```diff
@@ -97,3 +97,36 @@ example:
...
+For example:
+```export APPLICATION_CONFIG_PATH=/Users/{name}/nta/numenta-apps/htmengine/tests/support/skeleton-app/conf```
+(alternatively you can use the `APPLICATION_CONFIG_PATH` from an app based off
+of the HTMEngine, just be careful to not reset an existing MySQL database)
+
+With MySQL already started start/restart RabbitMQ:
+`rabbitmq-server -detached`
+
+And if you have an old copy of any of the numenta apps, run:
```
vitaly-krugl: Would be good to say what this will do and why it might help

htmengine/README.md
```diff
@@ -97,3 +97,36 @@ example:
...
+Setup the htmengine MySQL database:
+(don't do this if you're testing on Grok, Taurus, or another HTMEngine application)
+`./tests/integration/skeleton-app/reset_db.py`
+
+Start `supervisord`:
+```supervisord -c tests/support/skeleton-app/conf/supervisord.conf```
+*NOTE: Be sure to run `mkdir logs` within the `skeleton-app` directory so
+supervisor has a place to store its logs*
+
+Run the integration tests (currently only `result_quality_test.py`):
```
vitaly-krugl: The sentence would suffice sans 

> (currently only `result_quality_test.py`)

htmengine/tests/integration/result_quality_test.py
```diff
@@ -115,18 +114,19 @@ def deleteAmqpQueue(queue):
...
-          amqpClient):
+        amqpClient):
       amqpClient.declareQueue(self.resultsQueueName)
       self.addCleanup(deleteAmqpQueue, self.resultsQueueName)
 
       amqpClient.bindQueue(
-        queue=self.resultsQueueName,
-        exchange=self.config.get("metric_streamer", "results_exchange_name"),
-        routingKey="")
+          queue=self.resultsQueueName,
```
vitaly-krugl: Why did this need to change to 4-space indent? It should be just 1 level of 2-space indent. Same problem in many other statements here.

htmengine/README.md
```diff
@@ -97,3 +97,36 @@ example:
...
+`py.test tests/integration/result_quality_test.py`
```
vitaly-krugl: Make it more generic as `py.test tests/integration`

htmengine/tests/integration/result_quality_test.py
```diff
@@ -102,7 +102,6 @@ def setUp(self):
```
vitaly-krugl: @oxtopus's changes in PR #134 will need to be reconciled with these changes.

htmengine/tests/support/skeleton-app/conf/application.conf
```diff
@@ -0,0 +1,39 @@
...
+passwd =
+port = 3306
+
+[metric_streamer]
+# Exchange to push model results
+results_exchange_name = htmengine.model.results
+# Max records per batch to stream to model
+chunk_size = 1440
+
+[metric_collector]
```
vitaly-krugl: `metric_collector` is specific to Grok (AWS Cloudwatch metrics), so need to remove this section.

htmengine/tests/support/skeleton-app/conf/application.conf
```diff
@@ -0,0 +1,39 @@
...
+
+[anomaly_likelihood]
+# Minimal sample size for statistic calculation
+statistics_min_sample_size=100
+# How often to refresh the anomaly statistics in rows
+# We refresh once every two hours (ideally we would do this every record)
+statistics_refresh_rate=24
+# Sample size to be used for the statistic calculation
+# We keep a max of one month of history (assumes 5 min metric period)
+statistics_sample_size=8640
```
vitaly-krugl: The new .conf files should probably end with a newline char

htmengine/tests/support/skeleton-app/conf/supervisord.conf
```diff
@@ -0,0 +1,36 @@
...
+
+[rpcinterface:supervisor]
+supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
+
+[supervisorctl]
+serverurl=http://:
+
+
+[include]
+files = supervisord-base.conf
```
vitaly-krugl: Isn't `supervisord-base.conf` in `... htmengine/conf`? Would supervisor be able to find this file on its own?

htmengine/README.md
vitaly-krugl: Actually, it's not clear to me why this would help at all. The htmengine skeleton application conf file uses different exchange and queue name prefixes than our other apps, so there shouldn't be any conflict that necessitates resetting rabbitmq, right?

vitaly-krugl:
I am done with this round of code review. @jaredweiss will need to test this change with HTM Engine running standalone (without grok or taurus), debug, address existing review feedback, and resubmit for review.

htmengine/README.md
jaredweiss: The tests set up their own queue's but use the exchanges defined by the skeleton app's config. I guess since both should only be used in testing, those instructions aren't actually necessary.

htmengine/tests/integration/result_quality_test.py
jaredweiss: Pylint says that this is the proper indentation level and complains on 2-space indent. It's to differentiate between a continuation of the previous line (via the open `(`) and something like the inner part of a `while` or `if` statement. 

I verified this with @scottpurdy earlier this summer.

jaredweiss: @scottpurdy Actually my version of pylint is newer and is doing things not in our coding convention. Need to talk w/ you and @vitaly-krugl tomorrow about this

shantanoo: @jaredweiss I think issue is not with `pylint`. You need to use the `pytlintrc` from the repository. Can you try following command.

 `pylint --rcfile _path_to_numeta-apps_repo_/pylintrc _py_file_name_`

jaredweiss: @shantanoo 
@vitaly-krugl and I took a look at this and found out that my version of pylint was actually checking for indentation that his wasn't. We actually need to update the pylintrc to reflect the coding style we use.

oxtopus:
:goat::dash: 

:+1::exclamation:

PR #138, opened 2015-07-21
Updated synchronous_amqp_client_test to assert RabbitMQ has keys in responses

jaredweiss:
Makes sure that tests won't fail if responses from RabbitMQ don't have the proper keys in their responses. Since these tests are wrapped with the `_RETRY_ON_ASSERTION_ERROR` decorator, they will now retry rather than failing. 

@oxtopus Please review


oxtopus:
:+1: thanks!

PR #143, opened 2015-07-22
Hotfix/taur 1272 update rabbitmq server version to ensure version 3.5.3 gets installed

shantanoo:


infrastructure/saltcellar/rabbitmq/init.sls
```diff
@@ -38,11 +38,29 @@
     - group: root
     - mode: 0755
 
+# TODO: Find out a way to remove 'humanname' key
```
jcasner: Why do you want to remove the `humanname` key? According to the [Salt docs](http://docs.saltstack.com/en/latest/ref/states/all/salt.states.pkgrepo.html#salt.states.pkgrepo.managed), the field is required on yum based systems, like Centos.

PR #147, opened 2015-07-22
TAUR-1272 go back to using pkgrepo.managed

jcasner:
@unixorn @shantanoo please review

### DIFF
--- a/infrastructure/saltcellar/rabbitmq/init.sls
+++ b/infrastructure/saltcellar/rabbitmq/init.sls
@@ -38,19 +38,28 @@
     - group: root
     - mode: 0755
 
-# TODO: Use pkgrepo instead of cmd.
-add-packagecloud-rabbitmq-repo:
-  cmd.run:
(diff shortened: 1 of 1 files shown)


infrastructure/saltcellar/rabbitmq/init.sls
```diff
@@ -38,19 +38,28 @@
     - group: root
     - mode: 0755
 
-# TODO: Use pkgrepo instead of cmd.
-add-packagecloud-rabbitmq-repo:
-  cmd.run:
-    - name: curl -s https://packagecloud.io/install/repositories/rabbitmq/rabbitmq-server/script.rpm.sh > /tmp/script.rpm.sh && bash /tmp/script.rpm.sh && rm -f /tmp/script.rpm.sh
-    - user: root
+# Pulled from https://github.com/saltstack-formulas/rabbitmq-formula, which we
+# should consider using the entirety of.
```
unixorn: If we do decide to use it, we should add it to thirdparty as a direct git checkout instead of importing it into our repo. That'll just cause sync issues.

unixorn:
:cake:

PR #150, opened 2015-07-22
Updates to taurus initialization script

oxtopus:
This initializes a pair of instances to the point of being able to run tests!  cc @jcasner


taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
```diff
@@ -20,54 +20,217 @@
...
+This script will push the local numenta-apps repository to the remote Taurus
+instances, and reset to the commit sha specified in \$COMMIT_SHA.  The
+requisite python packages will be installed, configuration commands executed,
+and services started.  The end result upon successful completion of this script
+(as evidenced by a return code of 0), will be a fully configured, and running
+pair of Taurus instances suitable for use and/or testing.
+"
+
+if [ "${1}" == "-h" ]; then
+  echo "$USAGE"
```
jcasner: s/`"$USAGE"`/`"${USAGE}"`

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
```diff
@@ -20,54 +20,217 @@
...
+  exit 0
+fi
 
 set -o errexit
 set -o pipefail
-set -o verbose
 set -o nounset
+set -o verbose
```
jcasner: Considering using a `DEBUG` flag - either as an env var or a command line param - to turn on `verbose` mode

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
```diff
@@ -20,54 +20,217 @@
...
+  ssh -v -t "${TAURUS_SERVER_USER}"@"${TAURUS_SERVER_HOST}" \
+    "mkdir -p /opt/numenta/products/taurus/conf/ssl"
 
-  # Sync git histories with taurus server
-  git fetch "${TAURUS_SERVER_USER}"@"${TAURUS_SERVER_HOST}":/opt/numenta/products
-  git push "${TAURUS_SERVER_USER}"@"${TAURUS_SERVER_HOST}":/opt/numenta/products HEAD
+  # Generate env.sh credential files, to be copied to respective remote
+  # instances later
+  echo "
+    export PATH=/opt/numenta/anaconda/bin:\$PATH
```
jcasner: I don't think you need the `\` before the `$` here.  I think it should be s/`\$PATH``/${PATH}`

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
```diff
@@ -20,54 +20,217 @@
...
+    export PYTHONPATH=/opt/numenta/anaconda/lib/python2.7/site-packages:\$PYTHONPATH
```
jcasner: s/`\$PYTHONPATH`/`${PYTHONPATH}`

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
```diff
@@ -20,54 +20,217 @@
...
+    export ERROR_REPORT_EMAIL_SENDER_ADDRESS=${ERROR_REPORT_EMAIL_SENDER_ADDRESS}
+    export ERROR_REPORT_EMAIL_SES_ENDPOINT=${ERROR_REPORT_EMAIL_SES_ENDPOINT}
+    export AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
+    export AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}" | \
+    sed -e 's/^[ \t]*//' > \
+    taurus/pipeline/scripts/taurus.metric_collectors-env.sh
 
-  # Reset server state, apply database schema, start taurus services
+  echo "
+    export PATH=/opt/numenta/anaconda/bin:\$PATH
```
jcasner: s/`\$PATH`/`${PATH}`

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
```diff
@@ -20,54 +20,217 @@
...
+    export PYTHONPATH=/opt/numenta/anaconda/lib/python2.7/site-packages:\$PYTHONPATH
```
jcasner: s/`\$PYTHONPATH`/`${PYTHONPATH}`

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
```diff
@@ -20,54 +20,217 @@
...
   ssh -v -t "${TAURUS_SERVER_USER}"@"${TAURUS_SERVER_HOST}" \
     "cd /opt/numenta/products &&
-     git reset --hard ${COMMIT_SHA} &&
-     ./install-taurus.sh /opt/numenta/anaconda/lib/python2.7/site-packages /opt/numenta/anaconda/bin &&
-     cd /opt/numenta/products/taurus/taurus/repository &&
+     ./install-taurus.sh \
+        /opt/numenta/anaconda/lib/python2.7/site-packages \
+        /opt/numenta/anaconda/bin &&
+     taurus-set-rabbitmq \
+        --host=127.0.0.1 \
```
jcasner: Should we also use a `RABBITMQ_HOST` as we've done for MySQL?  In Staging and Production, for example, we use a remote server...

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
```diff
@@ -20,54 +20,217 @@
...
+  ssh -v -t "${TAURUS_SERVER_USER}"@"${TAURUS_SERVER_HOST}" \
+    "cd /opt/numenta/products &&
+     git reset --hard ${COMMIT_SHA}"
+
+  # Sync git histories with taurus collector for current HEAD
+  git push --force \
+    "${TAURUS_COLLECTOR_USER}"@"${TAURUS_COLLECTOR_HOST}":/opt/numenta/products \
+    `git rev-parse --abbrev-ref HEAD`
+
+  # Reset server state
```
jcasner: s/`server`/`collector`

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
```diff
@@ -20,54 +20,217 @@
...
-  # Return metric collector to active state
+exit 0;
+
+  # Reset metric collector state, apply database schema updates
   ssh -v -t "${TAURUS_COLLECTOR_USER}"@"${TAURUS_COLLECTOR_HOST}" \
-    "taurus-collectors-set-opmode active
-     cd /opt/numenta/products/taurus.metric_collectors/
-     supervisorctl -s http://127.0.0.1:8001 restart all"
+    "cd /opt/numenta/products &&
+     git reset --hard ${COMMIT_SHA} &&
```
jcasner: This seems redundant with line 130

taurus/pipeline/scripts/overrides/taurus/conf/nginx-taurus.conf
```diff
@@ -0,0 +1,80 @@
...
+    add_header "X-UA-Compatible" "IE=edge";
+
+    # Some endpoints require non-secure http access.  Create special-case
+    # location directives, and redirect everything else to the https-equivalent
+    # endpoint.  Note: = (exact match) and ^~ (prefix match) instruct nginx to
+    # stop searching for a match if one is found.
+    #
+    # See http://wiki.nginx.org/HttpCoreModule#location
+
+    location ^~ /_annotations {
```
jcasner: Do we need the `_annotations` end point in Taurus?

taurus/pipeline/scripts/overrides/taurus/conf/ssl/localhost.key
```diff
@@ -0,0 +1,15 @@
+-----BEGIN RSA PRIVATE KEY-----
```
jcasner: What is this private key?  For SSH communication between collector and server?

jcasner:
@oxtopus I'm finished with this round of review.

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
oxtopus: The $PATH and $PYTHONPATH need to be escaped in the remote environment.  Without the escapes, the local values are substituted during execution, which is not what we want.  This is deliberately different from the other vars in this set.

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
oxtopus: I tried that, but it didn't work.  Maybe something about the routing and dns.  On that note, we do need to do something, since we won't always be running rabbitmq on the same host.

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
oxtopus: Thanks.  I meant to remove it here.

taurus/pipeline/scripts/overrides/taurus/conf/ssl/localhost.key
oxtopus: It's the result of following instructions in README for generating self-signed ssl cert.  In practice, this should be discarded for a new set of ssl credentials, but for now is there for batteries-included purposes, since the generating step requires manual intervention (enter a password) and can't be automated.

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
jcasner: Aha, understood.  That makes sense!

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
jcasner: It's interesting that RABBITMQ_HOST works for the collector...

taurus/pipeline/scripts/overrides/taurus/conf/ssl/localhost.key
jcasner: We can update the generation script to pass the password as a command line arg

oxtopus: I may need some help with that.  I wasn't able to get it to work.

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
oxtopus: I originally felt compelled to force this as verbose, but if that's not desired, I think it's best to remove the line altogether, not add any special flags, and let the user invoke `bash -v` externally if they want it verbose.  For example, instead of `./initialize-remote-taurus-servers.sh`, the verbose version would be `/bin/bash -v ./initialize-remote-taurus-servers.sh`

jcasner: I'll defer to @unixorn on this one...

unixorn: I'd prefer checking if `${DEBUG}` is set to calling a script with `bash -v`.

It feels non-unixy to do interpreter script instead of having a proper shebang line.

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
```diff
@@ -20,54 +20,217 @@
...
+
+This script will push the local numenta-apps repository to the remote Taurus
+instances, and reset to the commit sha specified in \$COMMIT_SHA.  The
+requisite python packages will be installed, configuration commands executed,
+and services started.  The end result upon successful completion of this script
+(as evidenced by a return code of 0), will be a fully configured, and running
+pair of Taurus instances suitable for use and/or testing.
+"
+
+if [ "${1}" == "-h" ]; then
```
unixorn: Might as well make `--help` work here too

taurus/pipeline/scripts/overrides/taurus/conf/nginx-taurus.conf
oxtopus: Probably not, but this is how the nginx config is in the source.  The reason it's duplicated here is I need to change the user from root, to ec2-user.  I support removing the unneeded location directive, but it's outside the scope of this issue in particular.

oxtopus:
With the exception of the ssl cert, I've addressed your feedback.  I've restored the rabbitmq host var to read from the env as it should, but it looks like the connectivity issue I was facing had to do with the default guest:guest credentials not accepted on a public interface: https://www.rabbitmq.com/access-control.html

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
```diff
@@ -35,39 +123,175 @@ REPOPATH=`dirname "${SCRIPT}"`/../../..
...
+    commonName=${SSL_DOMAIN_NAME}
+    organizationalUnitName=${SSL_ORGANIZATIONAL_UNIT_NAME}
+    emailAddress=${SSL_EMAIL_ADDRESS}
+    " | sed -e 's/^[ \t]*//'`
+
+    # Generate the server private key
+    openssl genrsa \
+      -des3 \
+      -out taurus/pipeline/scripts/overrides/taurus/conf/ssl/${TAURUS_SERVER_HOST}.key \
+      -passout pass:myvoiceismypassportverifyme \
```
jcasner: Might be good to tokenize this, despite the excellent reference!

oxtopus: See latest, I made it random.

jcasner:
:+1:

PR #153, opened 2015-07-23
TAUR-1281: Make sure to add one-line title to notification …

lscheinkman:
…so it is also visible when the notification UI gets compressed.

@jcasner Please review


jcasner:
:+1:

PR #156, opened 2015-07-24
TAUR-1279 Add agamotto tests for local mysqld

unixorn:
Confirm that mysqld is enabled at boot and running.

@jcasner @shantanoo please CR


unixorn:
@oxtopus please also CR

oxtopus:
:+1:

PR #157, opened 2015-07-25
Strip out complain/logline from AMI build packer-scripts

unixorn:
If we need to log the output of the individual packer scripts, we should be doing it by piping the stdout/in from our rake task to a file. Writing it to logs on the instance being used for ami bakes isn't effective because it leaves junk on the AMI, and if the AMI bake fails, packer terminates the instance and we won't be able to look at the log files anyway.

@jcasner @shantanoo @oxtopus please CR


jcasner:
:+1:

PR #158, opened 2015-07-26
TAUR-851 Updater script consistent with initialization 

oxtopus:
This also adds some graceful handling of nginx and supervisor daemon states as well as invokes tests.

@jcasner update-remote-taurus-servers.sh will effectively form the basis for the taurus pipeline entry point.


oxtopus:
Merging so that I may test pipelines, but this still needs review.

PR #162, opened 2015-07-28
Including 'devtools' which has 'compiler-toolchain'

shantanoo:


shantanoo:
Cc: @jcasner, @unixorn

PR #163, opened 2015-07-28
Including 'nta-nucleus' which has 'ec2-user' state.

shantanoo:


shantanoo:
Cc: @jcasner, @unixorn

PR #165, opened 2015-07-28
TAUR-1324 Remove psutil, supervisord from standard infrastructure load

unixorn:
Instead, let the applications that need them specify the version they need in their requirements.txt.

If we use a third party app that needs them and _doesn't_ have them in their requirements.txt we can make a state that installs the version they need on just the instances using that role.

@jcasner @shantanoo please CR.

### DIFF
--- a/infrastructure/saltcellar/numenta-python/init.sls
+++ b/infrastructure/saltcellar/numenta-python/init.sls
@@ -60,28 +60,9 @@ anaconda-pip:
     - require:
       - pkg: anaconda-python
 
-# Install this from an S3 wheel so we don't need devtools on everything
-anaconda-psutil:
-  pip.installed:
-    - name: psutil == 1.0.1
-    - bin_env: /opt/numenta/anaconda/bin/pip
-    - watch_in:
-      - cmd: enforce-anaconda-permissions
-    - require:
-      - pkg: anaconda-python
-
 anaconda-setuptools:
   pip.installed:
     - name: setuptools == 18.0.1
-    - bin_env: /opt/numenta/anaconda/bin/pip
-    - watch_in:
-      - cmd: enforce-anaconda-permissions
-    - require:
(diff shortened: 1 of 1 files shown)


shantanoo:
@unixorn if two different packages need different versions of same package, are there any thoughts/guidelines to avoid the conflict? Having it at single place could be useful. Can we get rid of `paver` and `yaml` modules and keep only `pip`, `setuputils` and `wheel`? Otherwise changes LGTM.

infrastructure/saltcellar/numenta-python/init.sls
```diff
@@ -60,16 +60,6 @@ anaconda-pip:
...
-    - name: psutil == 1.0.1
-    - bin_env: /opt/numenta/anaconda/bin/pip
-    - watch_in:
-      - cmd: enforce-anaconda-permissions
-    - require:
-      - pkg: anaconda-python
-
 anaconda-setuptools:
   pip.installed:
     - name: setuptools == 18.0.1
```
jcasner: Should we leave this up to the requirements.txt too?  CC @oxtopus

unixorn: @jcasner I'll defer to @oxtopus on whether we should have setuptools by default.

unixorn:
@shantanoo If two applications need conflicting versions, at least one of them should go into a virutalenv, or they shouldn't be running on the same instance.

infrastructure/saltcellar/numenta-python/init.sls
unixorn: Talked to @oxtopus, he confirmed that we should keep setuptools since we need it to process requirements.txt. He also said we can just go with latest for the version.

unixorn:
Ready for another CR pass, @jcasner @shantanoo

shantanoo:
LGTM.

PR #167, opened 2015-07-28
TAUR-1324 Don't version-lock setuptools

unixorn:
We want whatever is the latest, not 18.0.1 specifically.

@oxtopus, please CR. Also, do you want to make pip install the latest pip too?


unixorn:
@oxtopus, @jcasner, @shantanoo please CR

shantanoo:
:+1:

PR #170, opened 2015-07-30
TAUR-1349 need systemd init scripts for grok

shantanoo:
@jcasner, @unixorn  
This is the initial checkin. Created `systemd` files which internally call scripts from `/etc/init.d`. Since `systemd` does not support non-standard state, e.g. `grok-error`, have created new `systemd` service.

NOTE: Do not merge


infrastructure/saltcellar/grok-plumbing/files/supervisor-tooling/grokservices.initd
```diff
@@ -57,7 +57,7 @@ run-rootboot-helper() {
   if [ -x "${ROOTBOOT_HELPER}" ]; then
     "${ROOTBOOT_HELPER}" 2>&1 | tee >(logger -t rootboot-helper)
     if [ $? != "0" ]; then
-      service grok-preload grok-error
+      /etc/init.d/grok-preload grok-error
```
jcasner: let's leave the salt-cellar files alone and instead update the ami-scripts that are centos 7 specific.

infrastructure/saltcellar/rabbitmq/init.sls
```diff
@@ -63,6 +63,17 @@ rabbitmq-server:
       - file: /etc/rabbitmq
       - pkg: rabbitmq-server
 
+# NOTE: SELinux should be disabled to get rabbitmq working with systemd
+rabbitmq-create-systemd:
+  file.managed:
+    - name: /etc/systemd/system/rabbitmq-server.service
```
jcasner: remove from here and add to the centos 7 scripts in the ami tools

jcasner:
@shantanoo in general, the systemd scripts look OK to me - i'll defer to @unixorn on those.

We need to leave the salt stuff alone - that's for centos 6.5. Instead, please modify against the centos 7 scripts in https://github.com/numenta/numenta-apps/tree/master/infrastructure/ami-tools

jcasner:
@shantanoo looks like we still need to symlink these into place in [configure-grok-pipeline-centos7-ami](https://github.com/numenta/numenta-apps/blob/master/infrastructure/ami-tools/packer-scripts/configure-grok-pipeline-centos7-ami) (similar to [the init.d scripts](https://github.com/numenta/numenta-apps/blob/master/infrastructure/ami-tools/packer-scripts/configure-grok-pipeline-centos7-ami#L140-L144)).  Let me know when you're ready for another round of review!

unixorn:
@shantanoo the bash scripts LGTM, but I can't speak to the systemd service files. Since they're working, I call them good :-)

PR #171, opened 2015-07-30
TAUR-1351 Disable requiretty and !visiblepw in /etc/sudoers

unixorn:
Given that we already require SSH keys to login and don't permit password authentication, we're already granting passwordless sudo.

Having requiretty enabled doesn't give us any extra security anyway per: https://unix.stackexchange.com/questions/65774/is-it-okay-to-disable-requiretty and !visiblepw is irrelevant since we're granting passwordless sudo and relying on SSH key authentication.

@jcasner, @oxtopus, @shantanoo please CR.


oxtopus:
:+1:

PR #178, opened 2015-08-04
Specify PYTHON_LIBRARY cmake cli arg to prevent discovery of incorrec…

oxtopus:
…t version

CC @jcasner @akhilaananthram


akhilaananthram:
It works! :smile:

PR #180, opened 2015-08-04
Specify nupic.core remote repository, branch, and sha as parameters to run_pipeline

akhilaananthram:
This allows someone to run their own version on nupic.core on jenkins.  If the parameters aren't specified then the values in .nupic_modules are used.

@jcasner


infrastructure/infrastructure/utilities/cli.py
```diff
@@ -107,3 +107,7 @@ def runWithOutput(command, env=os.environ, logger=None):
     check_call(command, env=env)
   except CalledProcessError:
     raise CommandFailedError("Failed to execute command: %s" % command)
+  # Catch other exceptions like empty environment variable
+  except Exception as ex:
+    template = "An exception of type {0} occured.\nArguments:{1!r}"
+    logger.debug(template.format(type(ex).__name__, ex.args))
```
jcasner: Instead, use `logger.exception("Helpful message")`.  That will include a stack trace for you automatically.

jcasner:
One change to exception handling, but otherwise looks good. Let me know when you make the change and we can proceed from there

infrastructure/infrastructure/utilities/cli.py
vitaly-krugl: Example:
```
except Exception:
  logger.exception("check_call failed for command=%s", command)
```

infrastructure/infrastructure/utilities/cli.py
```diff
@@ -107,3 +107,7 @@ def runWithOutput(command, env=os.environ, logger=None):
     check_call(command, env=env)
   except CalledProcessError:
     raise CommandFailedError("Failed to execute command: %s" % command)
```
vitaly-krugl: `CalledProcessError` actually contains useful information, such as returncode, that may be helpful for debugging. By remapping the exception to a different exception and not including the details of the original exception in the message, we deny the engineer an opportunity to debug the problem more quickly.

infrastructure/infrastructure/utilities/cli.py
```diff
@@ -107,3 +107,7 @@ def runWithOutput(command, env=os.environ, logger=None):
     check_call(command, env=env)
```
vitaly-krugl: This entire thing with having both `logger` and `log` is confusing. How would someone maintaining the code know when to use one versus the other?

Also, why is `logger` used as the flag to decide whether to print `log.printEnv(env, logger)`? This makes no sense whatsoever based on the information provided (or not provided) in this function.

infrastructure/infrastructure/utilities/cli.py
vitaly-krugl: And what if `logger` is None, or should we be using `log` instead?

infrastructure/infrastructure/utilities/cli.py
jcasner: @vitaly-krugl appreciate the feedback. I think the changes you're suggesting are outside the scope of the task that @akhilaananthram is working on right now.  May I suggest you add a JIRA to clean up this code, perhaps including a proposal for how it would be more effective?

akhilaananthram:
@jcasner: I updated the exception handling

vitaly-krugl:
@jcasner: I filed https://jira.numenta.com/browse/TAUR-1363 for the out-of-scope issues.

infrastructure/infrastructure/utilities/cli.py
```diff
@@ -107,3 +107,8 @@ def runWithOutput(command, env=os.environ, logger=None):
...
   except CalledProcessError:
     raise CommandFailedError("Failed to execute command: %s" % command)
+  # Catch other exceptions like empty environment variable
+  except Exception as ex:
+    template = "An exception of type {0} occured.\nArguments:{1!r}"
+    logger.exception(template.format(type(ex).__name__, ex.args))
```
vitaly-krugl: @akhilaananthram: you still have this bug here when logger is None. Also, as both Jared and I advised, you shouldn't be doing all this `template.format(type(ex).__name__, ex.args)` business. `logger.exception` automatically logs the actual exceptions for us in addition to the traceback. So, this should suffice: `logger.exception("check_call failed for command=%s", command)`

akhilaananthram: Sorry, I misunderstood. I'll fix that

akhilaananthram:
@vitaly-krugl: @jcasner: I corrected the logger message and check to see if it is None

infrastructure/infrastructure/utilities/cli.py
```diff
@@ -107,3 +107,8 @@ def runWithOutput(command, env=os.environ, logger=None):
     check_call(command, env=env)
   except CalledProcessError:
     raise CommandFailedError("Failed to execute command: %s" % command)
+  # Catch other exceptions like empty environment variable
+  except Exception as ex:
+    if logger:
+      logger.exception("check_call failed for command=%s" % command)
```
vitaly-krugl: LGTM

jcasner: s/` %`/`,` - if you run pylint on this, it will give you a warning about using string formatting.  Essentially, if you use the `%` here, it will always do the string formatting.  However, if you pass `command` as an argument, it will only replace the value in the string if there's actually an exception.

jcasner:
:+1:

infrastructure/infrastructure/utilities/cli.py
```diff
@@ -107,3 +107,8 @@ def runWithOutput(command, env=os.environ, logger=None):
     check_call(command, env=env)
   except CalledProcessError:
     raise CommandFailedError("Failed to execute command: %s" % command)
+  # Catch other exceptions like empty environment variable
+  except Exception as ex:
+    if logger:
+      logger.exception("check_call failed for command=%s" , command)
```
vitaly-krugl: @akhilaananthram: need to get rid of the space before the comma. Are you running pylint on your changes?

akhilaananthram: @vitaly-krugl: I did not. Should I make a new PR to make that change?

vitaly-krugl:
It's a good idea to fix things. Also, please always run pylint and fix its findings before submitting code for (re)review. The legacy code may have a lot of pylint warnings in some cases; in those cases, only fix warnings related to your changes to make it easier to review your changes and file a JIRA for fixing the remaining warnings. If pylint finds a real error in the legacy code, then it's a good idea to get that fixed ASAP, either in your own PR or by filing a JIRA bug.

akhilaananthram:
Will do.

infrastructure/infrastructure/utilities/nupic/build_commands.py
```diff
@@ -281,9 +285,12 @@ def buildNuPIC(env, logger):
       shutil.rmtree("external/linux32arm")
 
       # build the wheel
-      command = ("python setup.py bdist_wheel bdist_egg --nupic-core-dir=%s " 
-                 "upload -r numenta-pypi" % os.path.join(env["NUPIC_CORE_DIR"],
-                                                         "build", "release"))
+      command = ("python setup.py bdist_wheel bdist_egg --nupic-core-dir=%s" %
```
shantanoo: Instead of `string` use `list`. e.g.

`command = ["python", "setup.py", "bdist_wheel"]`

`runWithOutput` function converts string to list using `split(" ")` function.

`runWithOutput` uses `check_call` function from `subprocess` module.

Consider the following example.

```
➤ date > 'a b c'
➤ python
>>> from subprocess import *
>>> check_call(['cat', 'a b c'])
Fri Aug  7 14:31:47 IST 2015
0
>>> check_call('cat "a b c"')
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 535, in check_call
    retcode = call(*popenargs, **kwargs)
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 522, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
>>>
```

In this case, spaces in the filename are properly handled when the command uses `list`. In case of `string`, we need to manually escape the special characters.

PR #183, opened 2015-08-04
TAUR-1349 complete AMI creation process

jcasner:
@shantanoo @unixorn please review

May need some more work based on my comments in TAUR-1349


shantanoo:
LGTM.

PR #184, opened 2015-08-04
Updated cli.py based on pylint

akhilaananthram:
Follow up to: https://github.com/numenta/numenta-apps/pull/180#issuecomment-127796707

There are still warnings for the arguments to runWithOutput.

@vitaly-krugl

### DIFF
--- a/infrastructure/infrastructure/utilities/cli.py
+++ b/infrastructure/infrastructure/utilities/cli.py
@@ -108,7 +108,7 @@ def runWithOutput(command, env=os.environ, logger=None):
   except CalledProcessError:
     raise CommandFailedError("Failed to execute command: %s" % command)
   # Catch other exceptions like empty environment variable
-  except Exception as ex:
+  except Exception:
     if logger:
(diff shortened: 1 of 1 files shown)


jcasner:
:+1:

vitaly-krugl:
@akhilaananthram: what are the warnings for the arguments? In the future, if you paste a sampling of the actual warnings into a comment, it will help speed up the conversation.

akhilaananthram:
@vitaly-krugl:
W0102,  32: 0 - Dangerous default value os.environ (__builtin__.dict) as argument (dangerous-default-value)
W0102,  91: 0 - Dangerous default value os.environ (__builtin__.dict) as argument (dangerous-default-value)

vitaly-krugl:
Thanks @akhilaananthram. @unixorn showed me a change that he's making that will also address this problem. He is redefining the keyword arg as `env=None` instead of `env=os.environ`, and then (inside the function body) checks `if env is None` and initializes env from os.environ, if so.

PR #188, opened 2015-08-06
removed nupic core branch env var

akhilaananthram:
I removed NUPIC_CORE_BRANCH since it was redundant because we have the sha.

@jcasner


jcasner:
:+1:

PR #189, opened 2015-08-06
TAUR-1370 Graceful handling of missing `lang` in twitter api response

oxtopus:


taurus.metric_collectors/tests/unit/twitter_direct_agent_test.py
```diff
@@ -164,6 +164,103 @@ def testReapMessagesWithLimitNotifications(self):
...
+    storer = twitter_direct_agent.TweetStorer(
+      taggingMap=Mock(),
+      aggSec=300,
+      msgQ=Mock(),
+      echoData=False)
+
+    aggRefDatetime = datetime(2015, 8, 5, 14, 44, 33)
+
+    msg = {
+      u'contributors': None,
```
scottpurdy: This looks like it was probably copied from somewhere but should we use double quotes and break longer lines?

scottpurdy:
:+1:

taurus.metric_collectors/tests/unit/twitter_direct_agent_test.py
oxtopus: Yes, it was copied from the error message associated with the breakage.

PR #190, opened 2015-08-06
TAUR-1372 #send-for-review Update how the branch is specified in the pipeline_controller

jcasner:


pipeline_controller
```diff
@@ -238,6 +238,11 @@ def parseArgs():
                             "have changed. If unspecified, the script will use "
                             "the latest green build commit SHA from the "
                             "product-master-build Jenkins job."))
+  parser.add_argument("--branch", dest="branch", type=str,
+                      help=("**OPTIONAL** If specified, this determines which "
+                            "branch to work off of. Otherwise, this will be "
```
oxtopus: This help text is not accurate, since if the repo is in a detached head state (due to https://github.com/numenta/numenta-apps/blob/master/infrastructure/infrastructure/utilities/git.py#L132-L133), then there is no active branch.

oxtopus:
:+1: with reservations

jcasner:
@oxtopus addressed your feedback from our hipchat discussion and verified that this should not break the other pipelines because of the initial commit (02af824)

oxtopus:
Thanks!

PR #194, opened 2015-08-08
ENG-12 Created Unicorn back-end design via realistic pseudocode.

vitaly-krugl:
This is an incomplete prototypical implementation of the Unicorn "Python Engine" that demonstrates the proposed design.


vitaly-krugl:
CC @scottpurdy, @brev, @oxtopus, @lscheinkman, @jcasner

unicorn/engine/engine.py
```diff
@@ -0,0 +1,289 @@
...
+
+  def run(self):
+    g_log.info("Processing model=%s", self._modelId)
+
+    metricBuffer = []
+
+    rxRowCount = 0
+    while True:
+      # Read and decode the next sample
+      inputMessage = self._getInputMessage()
```
oxtopus: `_getInputMessage()` -> `_getNextInputMessage()` or implement iterator protocol and use something like `for inputMessage in self._readInputMessages():` instead of `while True:`

unicorn/engine/engine.py
```diff
@@ -0,0 +1,289 @@
...
+      timestamp, scalarValue = json.loads(inputMessage)
+      timestamp = datetime.utcfromtimestamp(timestamp)
+      inputFields = (timestamp, scalarValue)
+
+      rxRowCount += 1
+
+      if self._model is None:
```
oxtopus: Consider encapsulating some of this business logic elsewhere since `_processInputRow()` has side effects depending on the state of `_model`, and maybe implement a context manager.  `_processInputRow()` might even be a better place to implement this buffering and model creation mechanism.

unicorn/engine/engine.py
```diff
@@ -0,0 +1,289 @@
...
+  # getScalarMetricWithTimeOfDayParams will apply its default.
+  #
+  # TODO: is default acceptable?
+  _MIN_METRIC_RESOLUTION = None
+
+
+  def __init__(self, modelId):
+    self._modelId = modelId
+    self._model = None
+    self._anomalizer = _Anomalizer()
```
oxtopus: Does this need to be an instance property?

oxtopus:
LGTM with minor feedback.  Thanks!

scottpurdy:
High level this looks great. I have some organization suggestions that I'd like to discuss Monday.

unicorn/engine/engine.py
vitaly-krugl: The `_Anomalizer` will have state, so it needs to persist for the duration of a given engine instance's lifetime. If it's going to be used only in a single method without ever leaving the context of that method, then it can be changed to a local variable in that method, but that's not the case in the proposed design.

unicorn/engine/engine.py
vitaly-krugl: ok

unicorn/engine/engine.py
vitaly-krugl: After speaking with @marionleborgne, we've decided that the Front End will provide the min and max based on user data, enabling the Engine to create the model right away, without any need for buffering. 

@oxtopus, I am not sure what you had in mind with the Context Manager in your comment. Could you provide an example?

CC @brev

htmengine/htmengine/algorithms/modelSelection/clusterParams.py
```diff
@@ -26,7 +26,7 @@
 
 def setRandomEncoderParams(params, minVal, maxVal, minResolution):
   """
-  Given model params, figure out the correct parameters for the 
+  Given model params, figure out the correct parameters for the
```
vitaly-krugl: Changes in this module are not core to the design; just minor documentation fixes here.

htmengine/htmengine/runtime/scalar_metric_utils.py
```diff
@@ -63,7 +63,8 @@
 def generateSwarmParams(stats):
   """ Generate parameters for creating a model
 
-  :param stats: dict with "min" and "max"; values must be integer,float or None.
```
vitaly-krugl: Changes in this module are not core to the design; just minor documentation fixes here.

vitaly-krugl:
Please re-review. Summary of changes:

1. Front End passes metric data stats (min/max) to Engine via `--stats` command-line option
2. Address code review feedback.

CC @scottpurdy, @oxtopus, @brev, @marionleborgne

oxtopus:
:+1:

unicorn/engine/engine.py
```diff
@@ -0,0 +1,290 @@
...
+      json.load(pkg_resources.resource_stream(__name__, "stats_schema.json")))
+  except validictory.ValidationError as ex:
+    parser.error("--stats option value failed schema validation: %r" % (ex,))
+
+
+  return _Options(modelId=options.modelId, stats=stats)
+
+
+
+class _Anomalizer(object):
```
scottpurdy: Perhaps we can just use this directly?
https://github.com/numenta/nupic/blob/81a2c158235e305ad30850c27cd1acc42e3b9d4f/nupic/algorithms/anomaly_likelihood.py#L68

vitaly-krugl: @scottpurdy: Let's look into that when we start fleshing out this class. If I recall correctly, anomaly_likelihood_helper has logic that supports streaming data, as is the case here, but I could be wrong. I will add a comment about that here.

PR #196, opened 2015-08-10
Updated to AGPL.

rhyolight:
**DO NOT MERGE**

This PR should be merged at the same time as all other GPL Numenta codebases on github.com/numenta. This will be done by @rhyolight or @jcasner on or around August 14, 2015.


jcasner:
@rhyolight I would like to suggest that we update this PR to include a change to `htmengine/requirements.txt` and `unicorn/requirements.txt` to make them rely on `nupic==0.2.7`.  I would also block the merge of this PR until we have merged the changes in `nupic.core` and `nupic` and released `0.2.7` of nupic.  Does that make sense to you? Or is that overkill for this?

rhyolight:
Makes sense on both accounts.

oxtopus:
What about docstrings, for example, from https://github.com/numenta/numenta-apps/blob/master/taurus/taurus/__init__.py#L7-L9

```
# This program is free software: you can redistribute it and/or modify
# it under the terms of the GNU General Public License version 3 as
# published by the Free Software Foundation.
```

Does that not need to be revised to match?

rhyolight:
> What about docstrings

Aw shit.

rhyolight:
Copyright headers done on this PR.

PR #197, opened 2015-08-10
TAUR-1338 Add more git helper functions

unixorn:
Added clean and setRemoteURL functions per TAUR-1338 feedback.

@jcasner please CR

### DIFF
--- a/infrastructure/infrastructure/utilities/git.py
+++ b/infrastructure/infrastructure/utilities/git.py
@@ -143,7 +143,8 @@ def clean(path, arguments, logger):
   :param path: git directory to clean
 
   :param arguments: str containing optional extra command line arguments
-  for git clean
+  for git clean, as you would type them on the command line. If you wanted
+  to do `git clean -fd`, you'd set arguments to "-fd".
 
   :param logger: An initialized logger object
 
--- a/infrastructure/infrastructure/utilities/git.py
+++ b/infrastructure/infrastructure/utilities/git.py
@@ -152,8 +152,8 @@ def clean(path, arguments, logger):
   """
   assert logger
   assert isinstance(arguments, basestring), (
-    "arguments must be a str, but is %r" % arguments)
-  assert isinstance(path, basestring), "path must be a str, but is %r" % path
+    "arguments must be a string, but is %r" % arguments)
+  assert isinstance(path, basestring), "path must be a string, but is %r" % path
 
   command = "git clean"
   if arguments:
@@ -179,10 +179,10 @@ def setRemoteURL(remote, url, path, logger):
   """
   assert logger
   assert isinstance(path, basestring), (
-    "path must be a str, but is %r" % path)
+    "path must be a string, but is %r" % path)
   assert isinstance(remote, basestring), (
-    "remote must be a str, but is %r" % (remote))
-  assert isinstance(url, basestring), "url must be a str, but is %r" % (url)
+    "remote must be a string, but is %r" % (remote))
+  assert isinstance(url, basestring), "url must be a string, but is %r" % (url)
 
   logger.debug("* Setting url for %s to %s in %s", remote, url, path)
   with changeToWorkingDir(path):
--- a/infrastructure/infrastructure/utilities/git.py
+++ b/infrastructure/infrastructure/utilities/git.py
@@ -138,6 +138,48 @@ def getActiveBranch():
   return branch
 
 
+def clean(path, logger):
+  """
+  Changes to path, then runs git clean -fd.
+
+  @param path: git directory to clean
+
+  @param logger: An initialized logger object
+
+  @raises CommandFailedError: if git clean fails
+  """
+  assert path
+  assert logger
+
+  logger.debug("* Cleaning %s", path)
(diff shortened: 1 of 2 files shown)


infrastructure/infrastructure/utilities/git.py
```diff
@@ -138,6 +138,48 @@ def getActiveBranch():
   return branch
 
 
+def clean(path, logger):
+  """
+  Changes to path, then runs git clean -fd.
+
+  @param path: git directory to clean
```
jcasner: please use the correct docstring format (`:param`)

infrastructure/infrastructure/utilities/git.py
```diff
@@ -138,6 +138,48 @@ def getActiveBranch():
   return branch
 
 
+def clean(path, logger):
```
jcasner: Please rename to be more explicit since we're using the `-fd` options.  Alternatively, make these options available via the command line.

unixorn:
@jcasner made the changes you suggested. Ready for another CR

infrastructure/infrastructure/utilities/git.py
```diff
@@ -138,6 +138,50 @@ def getActiveBranch():
   return branch
 
 
+def clean(path, arguments="", logger):
```
jcasner: If we're doing `arguments` like this, then we need to have clear instructions in the docstring about the format.  We should probably also check the format.

The alternative would be to have some boolean values (`force`, `directories`, etc) and just build the arguments before passing to the command.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -138,6 +138,50 @@ def getActiveBranch():
...
+  """
+  Changes to path, then runs git clean -fd.
```
jcasner: This doesn't match anymore

infrastructure/infrastructure/utilities/git.py
```diff
@@ -138,6 +138,50 @@ def getActiveBranch():
...
+
+  :param path: git directory to clean
+
+  :param arguments: optional extra command line arguments for git clean
```
jcasner: As above, make this super clear how you expect people to send in args

infrastructure/infrastructure/utilities/git.py
```diff
@@ -138,6 +138,50 @@ def getActiveBranch():
...
+
+  :param logger: An initialized logger object
+
+  :raises CommandFailedError: if git clean fails
+  """
+  assert path
+  assert logger
```
jcasner: also need to validate the format of the `argument` field

infrastructure/infrastructure/utilities/git.py
```diff
@@ -138,6 +138,50 @@ def getActiveBranch():
...
+
+  logger.debug("* Cleaning %s", path)
+  with changeToWorkingDir(path):
+    return executeCommand(command="git clean %s" % (arguments), logger=logger)
```
jcasner: `arguments` does not need to be in parens here

infrastructure/infrastructure/utilities/git.py
```diff
@@ -138,6 +138,50 @@ def getActiveBranch():
...
+  :raises CommandFailedError: if git set-url fails
+  """
+  assert logger
+  assert path
+  assert remote
+  assert url
+
+  logger.debug("* Setting url for %s to %s in %s", remote, url, path)
+  with changeToWorkingDir(path):
+    return executeCommand(command="git set-url %s %s", remote, url,
```
jcasner: `command="git set-url %s %s", remote, url` --> `command="git set-url %s %s" % (remote, url)`

jcasner:
@unixorn i'm done with this round of review

infrastructure/infrastructure/utilities/git.py
vitaly-krugl: This wouldn't pass pylint and wouldn't run: in `def clean(path, arguments="", logger):`, non-initialized arg following initialized arg.

unixorn:
@jcasner this is ready for another CR.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -138,6 +136,56 @@ def getActiveBranch():
...
+  :param arguments: str containing optional extra command line arguments
+  for git clean, as you would type them on the command line. If you wanted
+  to do `git clean -fd`, you'd set arguments to "-fd".
+
+  :param logger: An initialized logger object
+
+  :raises CommandFailedError: if git clean fails
+  """
+  assert logger
+  assert isinstance(arguments, basestring), "arguments not str: %r" % arguments
```
jcasner: I'm not sure I'd understand this error message if I received it. Let's be more explicit / verbose

unixorn:
@jcasner, I've changed the message text.

jcasner:
:+1: Although I'd prefer `string` to `str` - should be obvious, but doesn't hurt us to spell it out fully

unixorn:
@jcasner, did s/str,/string,/, merging.

PR #199, opened 2015-08-11
TAUR-1369 Cleanup git.py

shantanoo:
Cc: @jcasner , @unixorn, @vitaly-krugl


infrastructure/infrastructure/utilities/git.py
```diff
@@ -24,114 +24,136 @@
...
 from infrastructure.utilities.exceptions import (CommandFailedError,
                                                  DetachedHeadError)
 from infrastructure.utilities.path import changeToWorkingDir
-from infrastructure.utilities.cli import executeCommand
 
 
+# Default logger. Need to find a better way for initializing global
+# logger for the module
+
+g_logger = diagnostics.initPipelineLogger(name="git",
```
jcasner: :-1: I don't like this approach.  This is a utility script, so all log messages should belong to the calling process, not to the utility function(s) being called.  Likewise, we want to ensure that we're always logging at the user configured level, not at a default level that is hard coded.  So, throughout this file, we'll need to make sure that `logger` is a required parameter.  Anywhere that this utility script is used, we'll need to make sure we're passing the parent logger into the function calls.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -36,102 +35,118 @@ def checkIfOptionSet(option, **kwargs):
...
 
-  @returns: total commit count for the git directory
+  :returns: total commit count for the git directory
 
-  @rtype: string
+  :rtype: string
   """
   with changeToWorkingDir(path):
-    return executeCommand("git rev-list HEAD --count")
+    return executeCommand("git rev-list HEAD --count", logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -36,102 +35,118 @@ def checkIfOptionSet(option, **kwargs):
...
     the command fails
 
-  @returns: The full path of the root folder of the current git repo
-  @rtype: string
+  :returns: The full path of the root folder of the current git repo
+
+  :rtype: string
   """
-  return executeCommand("git rev-parse --show-toplevel")
+  return executeCommand("git rev-parse --show-toplevel", logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -36,102 +35,118 @@ def checkIfOptionSet(option, **kwargs):
...
 
-  @returns: A `set` of modified files or None if the command fails
-  @rtype: set
+  :returns: A `set` of modified files or None if the command fails
+
+  :rtype: set
   """
   return set(executeCommand(
-             "git diff --name-only %s...%s" % (startSha, endSha)).split("\n"))
+             "git diff --name-only %s...%s" % (startSha, endSha),
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -36,102 +35,118 @@ def checkIfOptionSet(option, **kwargs):
...
     git repository
 
-  @returns: The current SHA
-  @rtype: string
+  :returns: The current SHA
+
+  :rtype: string
   """
-  return executeCommand("git log -n1 --pretty=%H")
+  return executeCommand("git log -n1 --pretty=%H", logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -36,102 +35,118 @@ def checkIfOptionSet(option, **kwargs):
...
 
-  @returns: The active branch name or the current SHA if in a detached head
+  :returns: The active branch name or the current SHA if in a detached head
     state
-  @rtype: string
+
+  :rtype: string
   """
-  branch = executeCommand("git rev-parse --abbrev-ref HEAD")
+  branch = executeCommand("git rev-parse --abbrev-ref HEAD", logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -139,44 +154,51 @@ def getActiveBranch():
...
-  @rtype: string
+  :returns: The blob output of git clone
+
+  :rtype: string
   """
   command = "git clone %s" % gitURL
   if checkIfOptionSet("directory", **kwargs):
     command += " " + kwargs["directory"]
-  return executeCommand(command)
+  return executeCommand(command, logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -187,87 +209,107 @@ def checkout(pathspec, **kwargs):
     command += " --theirs"
 
   command += " %s" % pathspec
-  return executeCommand(command)
+  return executeCommand(command, logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -187,87 +209,107 @@ def checkout(pathspec, **kwargs):
...
+  :raises:
     infrastructure.utilities.exceptions.CommandFailedError: if
     the command fails
+
+  :returns: The text blob output of git checkout
+
+  :rtype: string
   """
-  return checkout(pathspec, new=True)
+  return checkout(pathspec, new=True, logger=logger)
```
jcasner: `pathspec=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -187,87 +209,107 @@ def checkout(pathspec, **kwargs):
...
+  :raises:
     infrastructure.utilities.exceptions.CommandFailedError: if
     the command fails
+
+  :returns: The text blob output of git checkout
+
+  :rtype: string
   """
-  return checkout(pathspec, orphan=True)
+  return checkout(pathspec, orphan=True, logger=logger)
```
jcasner: `pathspec=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -187,87 +209,107 @@ def checkout(pathspec, **kwargs):
...
+  :returns: The exit code
+
+  :rtype: int
   """
   command = "git reset "
   if checkIfOptionSet("hard", **kwargs):
     command += "--hard"
   command += " %s" % sha
-  return executeCommand(command)
+  return executeCommand(command, logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -187,87 +209,107 @@ def checkout(pathspec, **kwargs):
...
     the command fails
 
-  @returns: The exit code
-  @rtype: int
+  :returns: The exit code
+
+  :rtype: int
   """
-  return reset(sha, hard=True)
+  return reset(sha, hard=True, logger=logger)
```
jcasner: `sha=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -297,43 +339,47 @@ def revParse(commitish, **kwargs):
     command += " --abbrev-ref"
 
   command += " %s" % commitish
-  return executeCommand(command)
+  return executeCommand(command, logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -297,43 +339,47 @@ def revParse(commitish, **kwargs):
...
-  @raises
+  :param logger: logger for additional debug info if desired
+
+  :raises:
     infrastructure.utilities.exceptions.CommandFailedError: if
     the command fails
   """
   command = "git fetch %s %s" % (repository, refspec)
-  return executeCommand(command)
+  return executeCommand(command, logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -341,34 +387,38 @@ def showRef(refList, **kwargs):
   if checkIfOptionSet("verify", **kwargs):
     command += " --verify"
   command += " %s" % refList
-  return executeCommand(command)
+  return executeCommand(command, logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -341,34 +387,38 @@ def showRef(refList, **kwargs):
...
-  @raises
+  :param logger: logger for additional debug info if desired
+
+  :raises:
     infrastructure.utilities.exceptions.CommandFailedError: if
     the command fails
   """
   command = "git add %s" % pathspec
-  return executeCommand(command)
+  return executeCommand(command, logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -376,21 +426,23 @@ def commit(message, **kwargs):
   if checkIfOptionSet("amend", **kwargs):
     command += " --amend"
   command += " %s" % message
-  return executeCommand(command)
+  return executeCommand(command, logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -398,43 +450,50 @@ def merge(path, message, **kwargs):
   if checkIfOptionSet("noFF", **kwargs):
     command += " --no-ff"
   command += " -m %s %s" % (message, path)
-  return executeCommand(command)
+  return executeCommand(command, logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/git.py
```diff
@@ -398,43 +450,50 @@ def merge(path, message, **kwargs):
...
 
-  @returns output of git rm
+  :returns output of git rm
 
-  @rtype: str
+  :rtype: str
   """
   command = "git rm -rf %s" % path
-  return executeCommand(command)
+  return executeCommand(command, logger=logger)
```
jcasner: `command=`

infrastructure/infrastructure/utilities/jenkins.py
```diff
@@ -73,11 +77,11 @@ def getWorkspace():
   if "WORKSPACE" in os.environ:
     workspace = os.environ["WORKSPACE"]
   else:
-    workspace = getGitRootFolder()
+    workspace = getGitRootFolder(logger)
```
jcasner: `logger=`

infrastructure/infrastructure/utilities/nupic/run_pipeline
```diff
@@ -152,7 +152,9 @@ def main():
   log.printEnv(env, g_logger)
 
   if not args["nupicSha"]:
-    args["nupicSha"] = git.getShaFromRemoteBranch(args["nupicRemote"], args["nupicBranch"])
+    args["nupicSha"] = git.getShaFromRemoteBranch(args["nupicRemote"],
```
jcasner: name the params please

infrastructure/infrastructure/utilities/rpm.py
```diff
@@ -177,6 +177,7 @@ def gitCloneIntoFakeroot(fakeroot,
   with a specific SHA (we normally build tip of master, for example), but we
   always want to include the exact SHA packaged in our RPM descriptions.
   """
+  # TODO: TAUR-1369 Make logger parameter mandatory.
```
jcasner: You make it mandatory in line 157; I'm confused...

infrastructure/infrastructure/utilities/nupic/build_commands.py
```diff
@@ -372,10 +373,13 @@ def cacheNuPIC(env, nupicSha, uploadToS3, logger):
                               s3Folder, logger)
       createTextFileAndUpload(nupicSha, wheelFileName, fileDir, s3Folder,
                               logger)
-      shutil.move("nupic-package-version.txt", ARTIFACTS_DIR)
+
+      artifacts_dir = createOrReplaceArtifactsDir(logger=logger)
```
unixorn: s/artifacts_dir/artifactsDir/ per Numenta guidelines.

infrastructure/infrastructure/utilities/rpm.py
```diff
@@ -177,6 +177,7 @@ def gitCloneIntoFakeroot(fakeroot,
...
   if logger:
```
unixorn: if logger is mandatory, we can remove this `if logger` clause.

shantanoo: Had modified this file first and then other files from which this function was being called. Will remove `if` and `TODO:`.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -190,135 +209,164 @@ def setRemoteURL(remote, url, path, logger):
...
-  @param pathspec: The name of the branch (commit-ish)
+  :param pathspec: The name of the branch (commit-ish)
+
+  :param logger: logger for additional debug info if desired
 
-  @param kwargs:
+  :param kwargs:
 
-  @raises
+  :raises:
```
jcasner: I think the final colon goes at the end of the exception name (so s/`:raises:`/`:raises`)

infrastructure/infrastructure/utilities/git.py
```diff
@@ -190,135 +209,164 @@ def setRemoteURL(remote, url, path, logger):
...
   """
   Switches to a given commit-ish
 
-  @param pathspec: The name of the branch (commit-ish)
+  :param pathspec: The name of the branch (commit-ish)
+
+  :param logger: logger for additional debug info if desired
 
-  @param kwargs:
+  :param kwargs:
```
jcasner: Need to define this thoroughly.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -190,135 +209,164 @@ def setRemoteURL(remote, url, path, logger):
...
+  :raises:
     infrastructure.utilities.exceptions.CommandFailedError: if
     the command fails
+
+  :returns: The text blob output of git checkout
+
+  :rtype: string
   """
-  return checkout(pathspec, new=True)
+  return checkout(pathspec=pathspec, new=True, logger=logger)
```
jcasner: You moved `**kwargs` to the end of the parameter set, so I don't think this will work properly.  Perhaps we should abandon `**kwargs` in favor of specific, named arguments?

infrastructure/infrastructure/utilities/git.py
```diff
@@ -190,135 +209,164 @@ def setRemoteURL(remote, url, path, logger):
...
 
-  @returns: The exit code
-  @rtype: int
+  :returns: The exit code
+
+  :rtype: int
   """
-  return reset(sha, hard=True)
+  assert logger
+  return reset(sha=sha, hard=True, logger=logger)
```
jcasner: As above with `checkout`

infrastructure/infrastructure/utilities/nupic/build_commands.py
```diff
@@ -52,8 +52,9 @@
 
 
 
-SCRIPTS_DIR = os.path.join(git.getGitRootFolder(), "nupic-pipeline", "scripts")
-ARTIFACTS_DIR = createOrReplaceArtifactsDir()
+# TODO: TAUR-1369 Find out where SCRIPTS_DIR is being used. Accordingly modify
```
jcasner: I don't get this... If you're not sure where it's being used and want to deprecate / remove it, then you need to figure it out _before_ you remove it __OR__ you need to remove it completely and fix whatever breaks.

infrastructure/infrastructure/utilities/rpm.py
```diff
@@ -177,24 +179,23 @@ def gitCloneIntoFakeroot(fakeroot,
...
+                                repoDirectory))
+    git.clone(gitURL=gitURL, directory=repoDirectory, logger=logger)
+    workDirectory = os.path.join(fakeroot, installDirectory, repoDirectory)
     with changeToWorkingDir(workDirectory):
       if sha:
-        git.resetHard()
+        git.resetHard(logger=logger)
         logger.debug("Checking out SHA %s in %s", sha, workDirectory)
-        git.checkout(sha)
+        git.checkout(sha, logger=logger)
```
jcasner: `sha=sha`

infrastructure/infrastructure/utilities/git.py
unixorn: +1 for abandoning **kwargs throughout. They're a pita to understand what keys are valid.

infrastructure/infrastructure/utilities/git.py
shantanoo: On line 89 (previous code in red), we are using `@raises:`. Making is consistent with the previous code.

infrastructure/infrastructure/utilities/git.py
shantanoo: Code will work. Sample code.

```
➤ cat a.py
def f(a, b, **kargs):
    print a
    print b
    print kargs


f(a=1, t=20, b=2)
➤ python a.py
1
2
{'t': 20}
```

Since these are named params, sequence doesn't matter.

I prefer to have `**kargs`. Currently we are supporting three options for `git checkout`. `new`, `orphan`, `theirs`. But there are more options available with `git checkout`. e.g. `-f`, `--detach`, etc. Supporting other features in future will be easy in case of `**kargs`, IMO.

infrastructure/infrastructure/utilities/git.py
shantanoo: Will modify code depending on the `checkout` feedback.

infrastructure/infrastructure/utilities/nupic/build_commands.py
shantanoo: Was not able to find any reference to `SCRIPTS_DIR` in this repository.

```
➤ git grep SCRIPTS_DIR
grok/grok/pipeline/utils/build_commands.py:28:SCRIPTS_DIR = os.path.join(os.environ.get("PRODUCTS"), "grok", "grok",
infrastructure/infrastructure/utilities/nupic/build_commands.py:55:# TODO: TAUR-1369 Find out where SCRIPTS_DIR is being used. Accordingly modify
infrastructure/infrastructure/utilities/nupic/build_commands.py:57:#SCRIPTS_DIR = os.path.join(git.getGitRootFolder(), "nupic-pipeline", "scripts")
```

Will fix the issues on case by case basis.

infrastructure/infrastructure/utilities/git.py
shantanoo: @unixorn docstring is incomplete hence there is confusion. Will update the docstring to reduce confusion.

jcasner: @shantanoo OK, as long as docstring is updated, I'm OK with retaining `**kwargs` - you make a compelling point

infrastructure/infrastructure/utilities/git.py
```diff
@@ -34,102 +34,121 @@ def checkIfOptionSet(option, **kwargs):
...
   Get the commit count from a git directory tree
 
-  @param: path to git directory
+  :param: path to git directory
+
+  :param logger: logger for additional debug info if desired
 
-  @raises: infrastructure.utilities.exceptions.CommandFailedError:
-  if path isn't in a git checkout
+  :raises: infrastructure.utilities.exceptions.CommandFailedError:
```
jcasner: Confirmed in the `htmengine` code that the convention is `:raises <exception>: <description>`.  Let's go ahead and remove the trailing `:` from the end of `raises` throughout these files.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -34,102 +34,121 @@ def checkIfOptionSet(option, **kwargs):
...
 
-  @param option: The option that is being looked up.
+  :param option: The option that is being looked up.
 
-  @param kwargs: Dict containing all the arguments passed to the function.
+  :param kwargs: Dict containing all the arguments passed to the function.
 
-  @return: True if option is present in kwargs and is set to True.
-  @rtype: boolean
+  :return: True if option is present in kwargs and is set to True.
```
jcasner: Confirmed in the `htmengine` code that the convention is actually `:returns: `

infrastructure/infrastructure/utilities/git.py
```diff
@@ -190,135 +209,175 @@ def setRemoteURL(remote, url, path, logger):
...
 
-  @returns: The text blob output of git checkout
-  @rtype: string
+  :returns: The text blob output of git checkout
+
+  :rtype: string
   """
-  command = "git checkout"
+  command = ["git", "checkout"]
   if checkIfOptionSet("new", **kwargs):
```
vitaly-krugl: @shantanoo: What happens if the caller passed an invalid or misspelled keyword arg in kwargs? The code will misbehave quietly. If you're using `**kwargs`, then you need to validate that `kwargs` doesn't contain any unexpected values. One way to handle this is to pop properties from it as you introspect those of interest. Then, before acting/executing the logic, check if anything is left in kwargs and raise ValueError with the contents of kwargs.

The code is too error-prone without such validation.

This needs to be fixed throughout.

CC @jcasner

infrastructure/infrastructure/utilities/git.py
shantanoo: Done.

infrastructure/infrastructure/utilities/git.py
shantanoo: Done.

infrastructure/infrastructure/utilities/git.py
shantanoo: I am ok with either approach.
If we decide with `**kwargs`, we can always add verification at the start of the function. Instead of popping properties, convert the `kwargs` keys to `set` and then subtract it from valid params.  If there are any elements, raise `InvalidParamerts` exception.
OTOH, if we decide not to continue with `**kwargs`, we should always use named parameters while calling functions. Thus, we will get rid of positional parameters and adding new parameters will be easy. (this may take more time and effort compared to other approach).

infrastructure/infrastructure/utilities/git.py
unixorn: I'm ok with leaving them as `**kwargs` as long as all the keywords are documented.

jcasner:
:+1:

unixorn:
:+1:

vitaly-krugl:
:-1:  kwargs are still not being checked for valid arg name set, which eas an error-prone situation that will lead to subtle, difficult-to-debug errors.

shantanoo:
@vitaly-krugl, so we finally have concluded that we should go ahead with `**kwargs` and do the validation; and not go with the other approach. Will update the PR accordingly.

vitaly-krugl:
@shantanoo: okay, thank you for the update.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -22,10 +22,18 @@
...
+                                                 InvalidParametersError)
 from infrastructure.utilities.path import changeToWorkingDir
-from infrastructure.utilities.cli import executeCommand
+
+
+
+def checkKwargs(kwargs, validKwargs):
+  invalidKwargs = set(kwargs) - validKwargs
+  if invalidKwargs:
+    raise InvalidParametersError("Invalid parameters passed %r" % invalidKwargs)
```
vitaly-krugl: Why are we using a custom exception for `InvalidParametersError`? Python already has the standard ValueError exception that is used for this purpose.

shantanoo: According to [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) documentation
```
exception ValueError

    Raised when a built-in operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.
```

So if we are expecting `string` and receive `list` as parameter, in that case we should raise `ValueError`.
Python raises `TypeError`. e.g.

```
>>> def f(a=None):
...  return 1
...
>>> f(1)
1
>>> f(2)
1
>>> f(a=2)
1
>>> f(b=20)
TypeError: f() got an unexpected keyword argument 'b'
>>>
```

I am not aware about the historical reasons behind use of `InvalidParametersError`. Since it was available in `exceptions` module and have been used everywhere, to keep the code consistent I have used `InvalidParametersError`. If we want to use native `TypeError` exception, will do that in another PR as it is out of scope for this PR.

vitaly-krugl: @shantanoo: sorry, you're correct in that python raises TypeError, not ValueError in this case. My main point is that we should follow established convention and not re-invent new exceptions for situations that are served by existing ones. This is the first time that I noticed this redundancy (`InvalidParametersError`), so I called it out in this PR. I can see that addressing it might be outside the scope of this PR.

unixorn: @shantanoo let's add a JIRA to replace `InvalidParametersError` with `TypeError` throughout infrastructure.utilities.

jcasner:
Other files that need to be touched still:

- /pipeline_controller.py
- grok/grok/pipeline/scripts/rpm-creator/rpm-creator.py
- grok/grok/pipeline/src/build.py

grok/grok/pipeline/scripts/rpm-creator/rpm-creator.py
```diff
@@ -170,7 +170,8 @@ def cleanseFakeroot(fakeroot, installDirectory, repoDirectory):
           g_logger.info("Found /tmp/noclean.rpms, skipping cleanup script")
         else:
           g_logger.info("Found %s, executing", cleanerScript)
-          runWithOutput("%s --destroy-all-my-work" % cleanerScript)
+          command = (cleanerScript, "--destroy-all-my-work")
```
vitaly-krugl: Our general practice has been not to create a local variable if it only gets used as a value in the next function call. An exception might be if the value is very long or complex. Otherwise, it's just unnecessary baggage to distract the reviewers. In this case, it might be better:

```
runWithOutput(command=(cleanerScript, "--destroy-all-my-work"),
              logger=g_logger)
```

grok/grok/pipeline/scripts/rpm-creator/rpm-creator.py
```diff
@@ -170,7 +170,8 @@ def cleanseFakeroot(fakeroot, installDirectory, repoDirectory):
...
+          runWithOutput(command=command, logger=g_logger)
```
vitaly-krugl: The name of `runWithOutput` implies that it returns output, and it looks strange that the output isn't being checked/used in some way. If the output isn't needed, perhaps there is a different utility function whose name makes it explicit that output is discarded?

infrastructure/infrastructure/utilities/git.py
```diff
@@ -190,135 +214,178 @@ def setRemoteURL(remote, url, path, logger):
...
   """
   Clones the given git repository
 
-  @param gitURL: The repository URL.
-  @param kwargs: Various options to git clone gitURL can be passed as keyword
+  :param gitURL: The repository URL.
+
+  :param logger: logger for additional debug info if desired
+
+  :param kwargs: Various options to git clone gitURL can be passed as keyword
```
vitaly-krugl: The docstring needs to define each keyword arg formally, not just in an example. List each keyword arg, state its type, and special values (e.g., default value, if any)

infrastructure/infrastructure/utilities/git.py
```diff
@@ -190,135 +214,178 @@ def setRemoteURL(remote, url, path, logger):
...
-  @rtype: string
+  :returns: The blob output of git clone
+
+  :rtype: string
   """
-  command = "git clone %s" % gitURL
+  validKwargs = {"directory"}
+  checkKwargs(kwargs=kwargs, validKwargs=validKwargs)
+  command = ["git", "clone", gitURL]
   if checkIfOptionSet("directory", **kwargs):
```
vitaly-krugl: Why expand kwargs in these calls? It seems like `checkIfOptionSet` should just accept the kwargs dict directly.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -190,135 +214,178 @@ def setRemoteURL(remote, url, path, logger):
...
-  @param kwargs:
+  :param logger: logger for additional debug info if desired
 
-  @raises
-    infrastructure.utilities.exceptions.CommandFailedError: if
-    the command fails
+  :param kwargs: Optional switches to git checkout. Following switches are
+                 supported at the moment.
+
+    new: Boolean. If True, create a new branch.
```
vitaly-krugl: Since these are optional, each one should explicitly state the default behavior.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -190,135 +214,178 @@ def setRemoteURL(remote, url, path, logger):
...
+
+    orphan: Boolean. If True, create a new orphan branch.
```
vitaly-krugl: Are these two parameters codependent? What if new is False and orphan is True? What if both are True?

infrastructure/infrastructure/utilities/git.py
```diff
@@ -190,135 +214,178 @@ def setRemoteURL(remote, url, path, logger):
...
 
-  @returns: The text blob output of git checkout
-  @rtype: string
+    theirs: Boolean. If True, when checking out paths from the index, check out
+            stage #3 (theirs) for unmerged paths. *
+
+  :raises infrastructure.utilities.exceptions.CommandFailedError:
```
vitaly-krugl: Our practice has been to document the exceptions after the return value: param, return/rtype, then raises.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -331,161 +398,187 @@ def revParse(commitish, **kwargs):
...
 
-def commit(message, **kwargs):
+def commit(message, logger, **kwargs):
   """
   Record changes to the repository
   Current implementation is supporting options like --amend
   This could be extended for other options as when required
 
-  @param message: Commit message.
+  :param message: Commit message.
```
vitaly-krugl: Please make sure that kwargs is documented here and everywhere else in this PR.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -331,161 +398,187 @@ def revParse(commitish, **kwargs):
...
+  :raises:
+    infrastructure.utilities.exceptions.CommandFailedError: if
+    the command fails
 
-  @raises:
+  :returns: A `String` representing the SHA
 
-  @return: A `String` representing the SHA
-  @rtype: String
+  :rtype: String
```
vitaly-krugl: the rtype value must be an actual class name; e.g., str, int

grok/grok/pipeline/scripts/rpm-creator/rpm-creator.py
shantanoo: Done.

infrastructure/infrastructure/utilities/git.py
shantanoo: Done.

infrastructure/infrastructure/utilities/git.py
shantanoo: Done.

infrastructure/infrastructure/utilities/git.py
shantanoo: Updated the docstring. Added `If 'new' is defined, this parameter will be ignored.`.

infrastructure/infrastructure/utilities/git.py
shantanoo: Done.

shantanoo: We need to have another PR for making this change in other files.

`git grep rtype | grep -i string | grep 'py:' | cut -d':' -f1 | sort -u`

infrastructure/infrastructure/utilities/git.py
shantanoo: Updated docsting for `type` and default value.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -190,302 +214,372 @@ def setRemoteURL(remote, url, path, logger):
...
-def add(pathspec):
+def add(pathspec, logger):
   """
   Add file contents to the index
 
-  @param pathspec: The file that is to be added to git.
+  :param pathspec: The file that is to be added to git.
 
-  @raises
+  :param logger: logger for additional debug info if desired
```
vitaly-krugl: @shantanoo: By "if desired", do you mean that logger is optional? I thought that logger is now required.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -190,302 +214,372 @@ def setRemoteURL(remote, url, path, logger):
...
 
-  @raises
-    infrastructure.utilities.exceptions.CommandFailedError: if
-    the command fails
+  :param new: Boolean. Defaults to False. If True, create a new branch.
+
+  :param orphan:
+    Boolean. Defaults to False. If True, create a new orphan branch.
+    If 'new' is defined, this parameter will be ignored.
```
vitaly-krugl: @shantanoo: if an invalid combination of parameters is specified, then this function should raise TypeError. We don't want to silently suppress invalid args and invalid combinations of args, which are most likely a bug in the caller.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -190,302 +214,372 @@ def setRemoteURL(remote, url, path, logger):
...
+  :param gitURL: The repository URL.
 
-  @raises
-    infrastructure.utilities.exceptions.CommandFailedError: if
-    the command fails
+  :param logger: logger for additional debug info if desired
 
-  @returns: The blob output of git clone
-  @rtype: string
+  :param directory:
```
vitaly-krugl: To be consistent with numenta code, the comment should begin on the same line as :param directory: and then wrap as needed.

infrastructure/infrastructure/utilities/git.py
shantanoo: Fixed.

vitaly-krugl:
@shantanoo: I reviewed the following incremental changes:

1. e7dcf0a4901e20406cb0cfc13948369d95c165aa
2. e2ab92e74f59ea10f55cd5681a4435ca33e508f0
3. 507601ddecd0bc1ee1b491d565458fbb26045e04


LGTM, once the noted "ignored" comment is fixed and the recommended DEPRECATED comment is added in InvalidParameterError docstring.

PR #202, opened 2015-08-12
TAUR-1379 Simple utility to recreate queues from one instance in another.

oxtopus:
Example usage:

```
nta-import-queues --source-host=REDACTED --source-user=REDACTED --source-passwd=REDACTED --destination-host=REDACTED --destination-user=REDACTED --destination-passwd=REDACTED --queue-regex="taurus\.mswapper\.model\.input\..+"
```


nta.utils/nta/utils/tools/import_queues.py
```diff
@@ -0,0 +1,134 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+from optparse import OptionGroup, OptionParser
```
jcasner: `optparse` is deprecated as of python 2.7.  Can you please use `argparse` instead?

nta.utils/nta/utils/tools/import_queues.py
```diff
@@ -0,0 +1,134 @@
...
+                  help=("Queues with names that match this pattern will be"
+                        " copied from source into destination.  If not"
+                        " specified, the default pattern '.*' will be used to"
+                        " copy all queues."))
+
+
+
+def _getClient(host, user, passwd, _cache={}):
+  """
+  :param str host: RabbitMQ host
```
jcasner: The convention I've seen used elsewhere is 
```
:param host: RabbitMQ host
:type host: str
```

oxtopus: Both are technically valid: http://sphinx-doc.org/domains.html#info-field-lists

jcasner: Cool.  We should probably put our preferred approach into our coding standards doc for clarity in the future...

vitaly-krugl: @jcasner: I started using the `:param type arg:` syntax a little while ago, and prefer it over the two-line approach. The two-line approach still has its place, such as when <type> description needs to consist of multiple space-separated words.

PR #203, opened 2015-08-13
TAUR-1371 Add supervisord checker

oxtopus:
Adds an implementation for checking supervisor status, including whether or not there are any services in a FATAL state and sends an email with the details.  Intended to be run as a periodic cron job.

Usage:

```
taurus-server-supervisor-monitor --serverUrl=http://HOST:PORT --monitorConfPath=REDACTED
taurus-collector-supervisor-monitor --serverUrl=http://HOST:PORT --monitorConfPath=REDACTED
```

cc @unixorn

### DIFF
--- a/taurus.monitoring/taurus/monitoring/supervisord_monitor/supervisord_monitor.py
+++ b/taurus.monitoring/taurus/monitoring/supervisord_monitor/supervisord_monitor.py
@@ -49,7 +49,13 @@ class SupervisorChecker(object):
   parser = OptionParser()
 
   parser.add_option("--monitorConfPath",
-                    help="Specify full path to monitor conf file")
+                    help=("Specify full path to ConfigParser-compatible"
+                          " monitor conf file, containing a [S1] section and"
+                          " the following configuration directives:\n\n"
+                          "MODELS_MONITOR_EMAIL_SENDER_ADDRESS\n"
+                          "MODELS_MONITOR_EMAIL_RECIPIENTS\n"
+                          "MODELS_MONITOR_EMAIL_AWS_REGION\n"
+                          "MODELS_MONITOR_EMAIL_SES_ENDPOINT"))
   parser.add_option("--serverUrl",
                     help="Supervisor API (e.g. http://127.0.0.1:9001)")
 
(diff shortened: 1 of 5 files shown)


taurus.monitoring/taurus/monitoring/supervisord_monitor/taurus_server_supervisord_monitor.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python
```
vitaly-krugl: The collector and this one are both servers, so the use of "server" in this script's name is not an ideal way to clearly distinguish it from the the other monitor. Since we renamed it to "taurus engine" elsewhere recently, let's make use of that name here as well. Recommending: `taurus_engine_supervisord_monitor.py`.

taurus.monitoring/taurus/monitoring/supervisord_monitor/supervisord_monitor.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+  def __init__(self):
+    (options, args) = self.parser.parse_args()
+
+    if args:
+      self.parser.error("Unexpected positional arguments: {}"
+                        .format(repr(args)))
+
+    server = xmlrpclib.Server(urljoin(options.serverUrl, "RPC2"))
```
vitaly-krugl: There is no need for the local variable `server`; assign directly to `self.server` instead.

taurus.monitoring/taurus/monitoring/supervisord_monitor/taurus_server_supervisord_monitor.py
oxtopus: Elsewhere in infrastructure, the distinction is between "server" and "collector", and this mirrors that convention.

taurus.monitoring/taurus/monitoring/supervisord_monitor/supervisord_monitor.py
```diff
@@ -0,0 +1,148 @@
...
+class SupervisorProcessInFatalState(SupervisorMonitorError):
+  pass
+
+
+
+class SupervisorChecker(object):
+  parser = OptionParser()
+
+  parser.add_option("--monitorConfPath",
+                    help="Specify full path to monitor conf file")
```
vitaly-krugl: Need to document what this code expects to find in the conf file.

vitaly-krugl:
Missing tests

vitaly-krugl:
@oxtopus: I'm done with the code review. I've noted some issues; the code looks nice, otherwise.

PR #207, opened 2015-08-17
Misc fixes

shantanoo:
This branch is used for making miscellaneous changes like modifications suggested by pylint, removing unused imports, improving docstrings, etc. Am still working on this PR. Please do not merge.

Cc: @jcasner, @unixorn


jcasner:
@shantanoo nice work so far on this PR - thanks for taking the initiative!

unixorn:
Looks good so far.

infrastructure/infrastructure/utilities/path.py
```diff
@@ -48,12 +47,12 @@ def rmrf(path, logger=None):
   you have to use different calls if a path points to a directory or file,
   which bloats the code.
 
-  @note: This behaves like `rm -rf` and will not complain if the directory
-  isn't empty, it will delete it and the contents anyway.
+  :note: This behaves like `rm -rf` and will not complain if the directory
+         isn't empty, it will delete it and the contents anyway.
 
-  @param path: path to file/directory you want deleted.
+  :param path: path to file/directory you want deleted.
```
jcasner: I have recently learned that you can (and should) include a type in the param string.  So, the format is:
`:param <type> <name>: <description> <example>`.  In this case, you could just do:
`:param str path: /path/to/delete`  
As long as we're updating, we should go ahead and make this change in the affected files.

jcasner:
Given this is weeks old and not getting any work, i'm closing it.

PR #208, opened 2015-08-17
TAUR-1381: Fix unknown metric unit #send-for-review

lscheinkman:
@jcasner Please review

### DIFF
--- a/grok-mobile/android/app/src/main/java/com/groksolutions/grok/mobile/service/MetricParser.java
+++ b/grok-mobile/android/app/src/main/java/com/groksolutions/grok/mobile/service/MetricParser.java
@@ -22,7 +22,7 @@
 
 package com.groksolutions.grok.mobile.service;
 
-import com.numenta.core.app.GrokApplication;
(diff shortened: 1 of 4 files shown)


grok-mobile/android/app/src/main/java/com/groksolutions/grok/mobile/service/MetricParser.java
```diff
@@ -251,6 +251,14 @@ void parseBody() throws IOException {
         serverName = serverName == null ? instanceId : serverName;
         Metric metric = GrokApplication.getDatabase().getDataFactory()
                 .createMetric(metricId, name, instanceId, serverName, lastRowId, parameters);
+
+        String unit = metric.getUnit();
+        if (unit == null) {
+            // Guess metric unit based on name if not given by server
```
jcasner: We don't want to guess the metric - that could lead to unpredictable outcomes.  Instead, if the server doesn't provide it, we should just have it be an empty string

jcasner:
:+1:

PR #209, opened 2015-08-17
TAUR-1381: Initialize metric unit value from custom grok factory

lscheinkman:
@jcasner This PR should fix both mobile builds (460 and 531)


jcasner:
:+1:

PR #211, opened 2015-08-18
Deleting breaking configuration and formalized indent-after-paren in pylintrc

vitaly-krugl:
Removed unnecessary nupic.support.lint.executable from load-plugins setting in pylintrc that was causing pylint to fail.

Formalize indent-after-paren=2 in pylintrc per nupic python coding guidelines of 2 spaces per indent.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -47,7 +47,7 @@ persistent=no
 
 # List of plugins (as comma separated values of python modules names) to load,
 # usually to register additional checkers.
-load-plugins=nupic.support.lint.executable
+load-plugins=
 
 
 [MESSAGES CONTROL]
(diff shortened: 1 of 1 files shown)


vitaly-krugl:
Please review

CC @scottpurdy

pylintrc
```diff
@@ -47,7 +47,7 @@ persistent=no
 
 # List of plugins (as comma separated values of python modules names) to load,
 # usually to register additional checkers.
-load-plugins=nupic.support.lint.executable
```
jcasner: What was included in this extension?  Are we losing anything important?

jcasner: Actually, looking at tip of master on nupic, this appears to still be around and seems like a useful thing to keep... https://github.com/numenta/nupic/blob/master/nupic/support/lint/executable.py

Are we sure we want to remove this?

vitaly-krugl: @jcasner: @scottpurdy determined that this should be removed. He made the same change in nupic the other day.

Perhaps the change should be accompanied either by the removal of https://github.com/numenta/nupic/blob/master/nupic/support/lint/executable.py or opening an issue regarding a fix that would permit us to continue using it. @scottpurdy ???

scottpurdy: We actually changed our policy on executables (you can search the nupic-hackers mailing list) so that plugin isn't valid anymore. And things started breaking since the way Pylint handles plugins has apparently changed as the PYTHONPATH doesn't seem to be used the same way.

PR #232, opened 2015-08-21
TAUR-1397 Helper utils for remediating a production issue in which st…

oxtopus:
…ock data is missing.

This adds the `taurus-delete-metric` command, for deleting a metric from a taurus server.  It also adds a new script (without a console entry point) to reset the emitted status for a stock symbol.

cc @vitaly-krugl please review asap for P1 issue

### DIFF
--- a/taurus.metric_collectors/taurus/metric_collectors/reset_emitted_status.py
+++ b/taurus.metric_collectors/taurus/metric_collectors/reset_emitted_status.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have purchased from
+# Numenta, Inc. a separate commercial license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


taurus.metric_collectors/taurus/metric_collectors/reset_emitted_status.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python
```
vitaly-krugl: If this is intended for stocks only, script name should reflect that, since twitter metrics also have "emitted state"

taurus.metric_collectors/taurus/metric_collectors/reset_emitted_status.py
```diff
@@ -0,0 +1,70 @@
...
+import argparse
+import random
+
+from taurus.metric_collectors import collectorsdb
+from taurus.metric_collectors.collectorsdb import schema
+
+
+
+def deleteFromEmitted(conn, table, symbol):
+  print ("Deleting `{}` from `{}` table of `{}` database"
```
vitaly-krugl: nit I personally prefer logging for its timestamps

vitaly-krugl:
If the collectors are running in active mode, this might be a problem here.

vitaly-krugl:
looks good, other than issues noted

taurus.metric_collectors/taurus/metric_collectors/reset_emitted_status.py
vitaly-krugl: @oxtopus: What about this one? I think that this was reasonable feedback regarding the script's name. The docstring in `reset_emitted_status.py` is specific to stocks "Reset emitted status for a **stock symbol**...", but the name of the script is too generic, since the concept of "emitted status" exists in the tweet metrics, too. Something like `reset_emitted_stock_status.py` would be better.

PR #234, opened 2015-08-21
ENG-25 Enable the production code path that instantiates the real NuPIC model instead of dummy model

vitaly-krugl:
the temporary work-around for ILLEGAL INSTRUCTION on my dev laptop now is to build nupic from source.


vitaly-krugl:
CC @scottpurdy

PR #235, opened 2015-08-21
Update pip to 7.1.1

unixorn:
Pip was bumped to 7.1.1

@jcasner @shantanoo please CR


shantanoo:
Should we get rid of `pip` version and always install the latest one?

unixorn:
The pip handler doesn't have a clean pip.latest function. If we set upgrade to true, it'll report that it did an upgrade every salt run, even if pip was up to date.

PR #241, opened 2015-08-24
nose and agamotto installs should set watch_in for enforce-anaconda-permissions

unixorn:
Ensure we don't end up with stray root-owned files in anaconda's site-packages

@jcasner @shantanoo please CR


jcasner:
:+1:

PR #242, opened 2015-08-25
UNI-161: Render `React` directly to `document.body`

lscheinkman:
CC: @brev


brev:
@lscheinkman ah yes that might help. I'll add this in UNI-159, but keeping the Flux action (I'll rename from Foo later heh). Will cc you back on UNI-159. thank you!

PR #247, opened 2015-08-25
TAUR-981 Make it easier to have /opt/numenta/anaconda owned by different users on different role machines

unixorn:
@jcasner, @shantanoo please CR.


jcasner:
I was talking to Subutai earlier today and he pointed out that we should really be treating Anaconda python just like system python.  In other words, we shouldn't be installing into the site-packages and bin folders directly.  Instead, we should install things with the `--user` option or similar (e.g., use a folder that's specific to a build, a user, etc).  Something to consider here...

unixorn:
@jcasner, our salt python installs are for stuff we need on the machine regardless of role. I agree that we should be installing application requirements with `python setup.py install` so that it can read and install any additional modules via their `requirements.txt` file.

That may make this obsolete in the long term, but for the short term, it solves a problem we're seeing with our jenkins installs.

@shantanoo, what do you think?

jcasner:
@unixorn we need to do more than just run `python setup.py install` to install packages; we need to specify `--install-dir` and `--script-dir` or `--prefix` with `setup.py`.  Likewise, when using `pip`, we need to specify `--user` or `--install-option="--prefix=/path/to/prefix"`.  The point is not just to install all of the dependencies.  The point is to install them into isolated locations that are not part of system python.  In the case of a Centos machine, anaconda should be treated as system python, meaning we don't `pip install` / `setup.py install` anything into `anaconda/lib/python2.7/site-packages`.  When that's true and we're using a custom install location for the python packages, then it doesn't matter who owns anaconda as long as the active user has permissions to execute `python`, `pip`, etc.  What's more important is that you have a location that **_IS_** writeable by the active user into which you can install any python packages.  I hope this clarifies my earlier point...

unixorn:
I'm agreed, we should write app-specific packages to an app-specific tree owned by that app's user by running `setup.py` with `--prefix` or `pip` with `--prefix=/path/to/prefix`

PR #253, opened 2015-08-26
ENG-37 Remove test invocations from taurus initialize/update/refresh scripts

vitaly-krugl:
ENG-37 Remove test invocations from initialize-remote-taurus-servers.sh, refresh-remote-taurus-servers.sh, and update-remote-taurus-servers.sh to avoid destructive outcome as in ENG-35.

NOTE: there is no way to tell which tests cause undesirable side-effects, so all test invocations have been removed from these scripts. Separate effort should be made to create new tests that validate the installation without any undesirable side-effects.

CC @jcasner, @oxtopus


oxtopus:
:-1:  As discussed, removing the test invocation from these scripts leaves us without a testing pipeline.  Instead, the tests should be treated as a configuration-driven _optional_ step during deployment.  Test pipeline(s) enable the option, which is disabled by default to mitigate risk to staging and production environments.

vitaly-krugl:
@oxtopus: this PR was one part of a dual-prong strategy, and was not intended to forego a testing pipeline. I explained this in our discussion(s) and also added a comment in https://jira.numenta.com/browse/ENG-37 describing the strategy. In the meantime, I am going to close this PR and open a new one per comments in ENG-37

PR #259, opened 2015-08-26
TAUR-1338 Add a mode argument to mkdirp (default 0755)

unixorn:
@jcasner please CR

### DIFF
--- a/infrastructure/infrastructure/utilities/path.py
+++ b/infrastructure/infrastructure/utilities/path.py
@@ -42,7 +42,7 @@ def changeToWorkingDir(path):
 
 
 
-def mkdirp(path, mode=0755):
+def mkdirp(path, mode=0777):
   """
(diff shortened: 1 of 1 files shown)


infrastructure/infrastructure/utilities/path.py
```diff
@@ -42,15 +42,16 @@ def changeToWorkingDir(path):
 
 
 
-def mkdirp(path):
+def mkdirp(path, mode=0755):
```
jcasner: The default `mode` for `os.mkdirs` is `0777`: https://docs.python.org/2/library/os.html#os.makedirs

If we're changing that functionality (which is OK), then we need to update where it is currently being used

unixorn:
@jcasner Updated the default mode to 777. Ready for another CR.

jcasner:
:+1:

PR #263, opened 2015-08-27
IMBU-5: Add web service configuration

lscheinkman:
@jcasner This is the web version. 
I have a small document with the extra steps needed to bring a demo server up using our internal salt roles.
FYI: @BoltzmannBrain, @unixorn

### DIFF
--- a/imbu/conf/supervisord.conf
+++ b/imbu/conf/supervisord.conf
@@ -1,0 +1,48 @@
+; Sample supervisor config file.
+;
+; For more information on the config file, please see:
+; http://supervisord.org/configuration.html
+;
+; Note: shell expansion ("~" or "$HOME") is not supported.  Environment
+; variables can be expanded using this syntax: "%(ENV_HOME)s".
+
+[unix_http_server]
+file=%(here)s/../imbu-supervisor.sock   ; (the path to the socket file)
+
+[inet_http_server]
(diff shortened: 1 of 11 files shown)


imbu/conf/supervisord.conf
```diff
@@ -0,0 +1,48 @@
...
+
+[rpcinterface:supervisor]
+supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
+
+[supervisorctl]
+serverurl=http://:
+
+;*************** FLUENT-API **************
+[program:fluent-api]
+command=uwsgi --enable-threads --socket 0.0.0.0:19002 --master --vacuum --idle 300 --processes 8 --threads 4 --listen 1024 --wsgi-file engine/fluent_api.py
```
oxtopus: Given the scope of imbu, I think we can safely reduce `--listen` to a much lower level that doesn't require system-level configuration to support it.

lscheinkman: Good point. 
Should I just remove this extra parameter and rely on `uwsgi` defaults? I think the default value is 100.

oxtopus: +1 for default value.  Can also reduce `--processes` to 2 or 1.

jcasner:
:+1:

PR #264, opened 2015-08-27
ENG-37 Implement conditional unit/integration test execution in taurus pipeline scripts

vitaly-krugl:
ENG-37 Implement conditional unit/integration test execution in refresh-remote-taurus-servers.sh based on optional arg: -t destructive-tests

### DIFF
--- a/taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
+++ b/taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
@@ -21,7 +21,7 @@
 # ----------------------------------------------------------------------
 #
 
-USAGE="Usage: `basename ${0}`
+USAGE="Usage: `basename ${0}` [-t destructive-tests]
 
 This script configures a pair of Taurus Server and Metric Collector
 instances in which the taurus services are not already running.  For example,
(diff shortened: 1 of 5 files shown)


vitaly-krugl:
@oxtopus: please review this change in refresh-remote-taurus-servers.sh. If the concept looks good to you, then I will copy this change to the other two scripts (initialize/update), test, and update the configuration in Jenkins to make use of this feature.

oxtopus:
:+1:

vitaly-krugl:
thx

vitaly-krugl:
@oxtopus @unixorn Please review

CC @scottpurdy @jaredweiss

taurus/pipeline/scripts/initialize-remote-taurus-servers.sh
```diff
@@ -324,8 +374,18 @@ pushd "${REPOPATH}"
      else
        supervisord -c conf/supervisord.conf
      fi &&
-     py.test ../nta.utils/tests &&
-     py.test tests &&
+     for run in {1..6}; do
+       if [[ \$(nta-get-supervisord-state http://localhost:8001) == \"RUNNING\" ]]; then
```
oxtopus: These files were meant to mirror the instructions to the user in the readme as much as possible.  I don't think a user would ever implement such a loop in their console.

Consider either building the retry logic into `nta.utils/nta/utils/tools/get_supervisord_state.py` or add a new command to specifically start supervisord (say, `nta-start-supervisord`) which is synchronous and doesn't return until it's up, with some reasonable timeout.  You could even do both!

Honestly, I wish there were an option to supervisord which implements such logic to avoid having to have a workaround like this altogether.

vitaly-krugl: @oxtopus: https://github.com/numenta/numenta-apps/pull/272

PR #266, opened 2015-08-28
Uni 157 init flat file system db

marionleborgne:
@brev I'm happy with jsondown :)

cc @lscheinkman @jcasner


brev:
:+1: merge away

PR #274, opened 2015-08-31
Update Grok to run htmengine and nta.utils tests when running integration tests

jcasner:
@oxtopus please review
This fixes the issue when trying to run htmengine tests from run_tests.py, addresses a failure when running nta.utils/tests/unit/logging_support_test.py from run_tests.py, and cleans up pylint issues in nta.utils/tests/unit/logging_support_test.py


infrastructure/infrastructure/utilities/nupic/build_commands.py
```diff
@@ -219,9 +219,9 @@ def buildNuPICCore(env, nupicCoreSha, logger, buildWorkspace):
       if "JENKINS_HOME" in os.environ:
         command += " bdist_wheel bdist_egg upload -r numenta-pypi"
       runWithOutput(command=command, env=env, logger=logger)
-    except Exception as originalException:
-      raise PipelineError("nupic.core building failed due to unknown reason.",
-                          originalException)
+    except:
```
vitaly-krugl: A catch-all `except` and renaming of the exception is usually not a good idea. For example, this interferes with the normal flow of `SIGINT` as well as `sys.exit(...)`.

PR #283, opened 2015-09-01
Add --nupic-core-dir to nupic.bindings setup.py call.

scottpurdy:


infrastructure/infrastructure/utilities/nupic/build_commands.py
```diff
@@ -259,7 +259,9 @@ def buildNuPICCore(env, nupicCoreSha, logger, buildWorkspace):
       shutil.rmtree("external/linux32arm")
 
       # build the distributions
-      command = "python setup.py install --prefix=%s" % buildWorkspace
+      command = (
+          "python setup.py install --prefix={} --nupic-core-dir={}".format(
+              buildWorkspace, os.path.join(os.getcwd(), "build", "release")))
       # Building on jenkins, not local
       if "JENKINS_HOME" in os.environ:
         command += " bdist_wheel bdist_egg upload -r numenta-pypi"
```
vitaly-krugl: @scottpurdy: I noticed that in other recent PRs, you broke up the command into a list of strings, while here it's all composed as a single string. Could this cause issues if the values being substituted into the single string contain spaces or anything else that would not play nice with the shell's command parsing logic?

scottpurdy: Yes exactly right. It will split based on spaces so if you have something like a string argument containing spaces it will be split into separate arguments. Most commands don't have this and it's fine to use a string. I don't have a strong opinion on some convention beyond doing whatever works in your situation but open to opinions that people have on it.

jcasner: @vitaly-krugl @scottpurdy the convention we've been following recently on the infrastructure side is as follows:

* When there are no variables in the command, use strings to promote readability
* When there are variables in the command, use tuples or lists to ensure that the spaces in the variables don't cause errors

PR #287, opened 2015-09-01
TAUR-1383 Add taurus monitoring install script and --subjectPrefix option

unixorn:
* Add install-taurus-monitoring script
* Add `--subjectPrefix` option so we can make the error message different for different machines (staging and production)

@jcasner please CR

### DIFF
--- a/install-taurus-monitoring
+++ b/install-taurus-monitoring
@@ -24,7 +24,7 @@ set -o errexit
 
 function install {
   pushd $1
-  python setup.py develop --install-dir=$2 --script-dir=$3
+    python setup.py develop --install-dir=$2 --script-dir=$3
   popd
 }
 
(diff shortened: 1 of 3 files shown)


install-taurus-monitoring
```diff
@@ -0,0 +1,33 @@
...
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+set -o errexit
+
+function install {
+  pushd $1
+  python setup.py develop --install-dir=$2 --script-dir=$3
```
jcasner: I prefer an indentation inside of `pushd`/`popd` to make it clearer what's running where

jcasner:
1 minor nit but otherwise LGTM :+1:

unixorn:
Updated the indentation

PR #299, opened 2015-09-02
TAUR-1383 Make supervisord listen on all interfaces

unixorn:
@oxtopus this is allow remote monitoring of the taurus server and collector. The security groups in EC2 will limit access to 9001 on the taurus instances to the monitoring server.

cc: @vitaly-krugl 

@jcasner please CR

### DIFF
--- a/taurus/conf/supervisord.conf
+++ b/taurus/conf/supervisord.conf
@@ -10,7 +10,7 @@
 file=%(here)s/../taurus-supervisor.sock   ; (the path to the socket file)
 
 [inet_http_server]
-port=127.0.0.1:9001
+port=*:9001
 
 [supervisord]
 environment=APPLICATION_CONFIG_PATH=/opt/numenta/taurus/conf
--- a/taurus.metric_collectors/conf/supervisord.conf
+++ b/taurus.metric_collectors/conf/supervisord.conf
@@ -2,7 +2,7 @@
 file=%(here)s/../supervisor.sock
 
 [inet_http_server]
-port=127.0.0.1:8001
+port=*:8001
 
 [rpcinterface:supervisor]
 supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface


taurus/conf/supervisord.conf
```diff
@@ -10,7 +10,7 @@
 file=%(here)s/../taurus-supervisor.sock   ; (the path to the socket file)
 
 [inet_http_server]
-port=127.0.0.1:9001
+port=*:9001
```
oxtopus: Why not 0.0.0.0?

oxtopus:
Should also need to modify https://github.com/numenta/numenta-apps/blob/master/taurus.metric_collectors/conf/supervisord.conf#L5 for the collector

taurus/conf/supervisord.conf
unixorn: The configuration documentation at [http://supervisord.org/configuration.html](http://supervisord.org/configuration.html) said to use :9001 or *:9001.

jcasner: I believe I read that if you use `0.0.0.0` it only handles IPv4, but if you use `:` or `*:` it handles both IPv4 and IPv6.  I can't find the link at the moment, so my recollection may be off...

jcasner:
:+1:

taurus/conf/supervisord.conf
oxtopus: We don't need support for IPv6 in this application, and * isn't working for me.

oxtopus: If you don't specify anything, supervisor attempts to detect the ip address using `socket.gethostbyname() and `socket.gethostname()`, which fails in ec2:

```
austins-mbp:scripts amarshall$ ssh REDACTED_EC2_HOST "python -c 'import socket; print socket.gethostbyname(socket.gethostname())'"
X11 forwarding request failed on channel 0
  File "<string>", line 1, in <module>
socket.gaierror: [Errno -2] Name or service not known
austins-mbp:scripts amarshall$ python -c 'import socket; print socket.gethostbyname(socket.gethostname())'
172.16.190.112
```

unixorn: Ok, I'll switch to 0.0.0.0, given that we're only using IPv4.

oxtopus: AFAIK, the root cause for this is how we're assigning the hostname.  `socket.gethostname()` is returning something different from the public or private dns name, and which is unrecognized in the `socket.gethostbyname()` DNS lookup.  So, not so much a problem with EC2, per se, but in how we're using it.

PR #300, opened 2015-09-03
Uni 160 button click trigger file upload

marionleborgne:
@brev: will sync with @lscheinkman for parsing data into the same data structure

cc @jcasner


brev:
looks good.

merge conflict? maybe merge master in?

otherwise, :+1:

PR #302, opened 2015-09-03
TAUR-1424 update version of rabbit in Grok

jcasner:
@unixorn please review

FYI @oxtopus @scottpurdy

### DIFF
--- a/infrastructure/ami-tools/packer-scripts/configure-grok-pipeline-centos7-ami
+++ b/infrastructure/ami-tools/packer-scripts/configure-grok-pipeline-centos7-ami
@@ -57,8 +57,15 @@ install-prerequisites-and-update-repos() {
 
   # Install MySQL and Rabbit
   sudo rpm -Uvh http://dev.mysql.com/get/mysql-community-release-el7-5.noarch.rpm
-  sudo yum install mysql-server mysql mariadb-devel rabbitmq-server nginx git -y
+  sudo yum install mysql-server mysql mariadb-devel nginx git -y
+  curl -O https://www.rabbitmq.com/releases/rabbitmq-server/v3.5.4/rabbitmq-server-3.5.4-1.noarch.rpm
+  sudo rpm --import https://www.rabbitmq.com/rabbitmq-signing-key-public.asc
+  sudo yum install rabbitmq-server-3.5.4-1.noarch.rpm
   sudo /usr/lib/rabbitmq/bin/rabbitmq-plugins enable rabbitmq_management
+  pushd /usr/local/bin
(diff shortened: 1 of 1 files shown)


unixorn:
+1

infrastructure/ami-tools/packer-scripts/configure-grok-pipeline-centos7-ami
```diff
@@ -57,8 +57,15 @@ install-prerequisites-and-update-repos() {
 
   # Install MySQL and Rabbit
   sudo rpm -Uvh http://dev.mysql.com/get/mysql-community-release-el7-5.noarch.rpm
-  sudo yum install mysql-server mysql mariadb-devel rabbitmq-server nginx git -y
+  sudo yum install mysql-server mysql mariadb-devel nginx git -y
```
oxtopus: mariadb-devel?  Why not mysql-devel?

jcasner: Because CentOS7 doesn't come with mysql...

https://www.digitalocean.com/community/questions/can-t-install-mysql-on-centos-7

oxtopus: My concern is with the combination of `mysql-server`, `mysql`, and **`mariadb-devel`**.  I think it ought to be consistent, at the very least.  Either `mysql-server`, `mysql`, and `mysql-devel`, or `mariadb-server`, `mariadb`, and `mariadb-devel`.  Regardless, that link suggests that `mysql-devel` _is_ available, and other comments don't suggest the mixed set of packages.

Remind me again why we're using CentOS7?

jcasner: I followed the instructions from that link for how to get mysql running...  In any event, trying to continue maintaining CentOS6.5 was more work than it was worth, and converting to CentOS7 was pretty trivial.  I'm happy to explore other OS's in the future.

unixorn: We're eventually going to have to update to C7 for our other infrastructure instances to keep getting security updates. It will also get us newer versions of packages than 6 without having to add extra yum repositories.

jcasner: Might be worth considering a transition to Ubuntu or CoreOS so that we can get rid of our internal NuPIC pipeline, amongst other things...  But that's a conversation for another day :)

shantanoo: +1 for ubuntu/CoreOS, since CentOS seem to contain older versions of packages compared to other distros.

PR #303, opened 2015-09-03
TAUR-1425 Remove nupic before installing taurus

oxtopus:


scottpurdy:
Is it a realistic case where code directories are left over on the path and pip cannot uninstall them? I thought the tricky case was when the easy-install.pth had vestiges of an already uninstalled package?

scottpurdy:
I think it is probably fine though.

oxtopus:
> Is it a realistic case where code directories are left over on the path and pip cannot uninstall them?

In my experience, yes, in particular wrt namespace packages

> Perhaps we should run the uninstall in a loop until it returns a non-zero exit code?

That was my initial approach, but I couldn't get it to work cleanly with both nupic and nupic.bindings so I decided to go nuclear

PR #308, opened 2015-09-03
TAUR-1356 Add function to set/unset EC2 instance termination protection

unixorn:
@jcasner @shantanoo, please CR

### DIFF
--- a/infrastructure/infrastructure/utilities/ec2.py
+++ b/infrastructure/infrastructure/utilities/ec2.py
@@ -234,6 +234,38 @@ def loadInstanceTags(instanceId,
                                 instanceId, region)
 
 
+def setEC2TerminationProtection(instanceId, status, config, logger):
+  """
+  Set termination protection for an instance to a specific value.
+
+  :param str instanceId: The instance ID to turn on termination protection for.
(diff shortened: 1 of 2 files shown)


infrastructure/infrastructure/utilities/ec2.py
```diff
@@ -234,6 +234,38 @@ def loadInstanceTags(instanceId,
                                 instanceId, region)
 
 
+def setEC2TerminationProtection(instanceId, status, config, logger):
+  """
+  Set termination protection for an instance to a specific value.
+
+  :param str instanceId: The instance ID to turn on termination protection for.
```
jcasner: s/`turn on`/`change`

infrastructure/infrastructure/utilities/ec2.py
```diff
@@ -234,6 +234,38 @@ def loadInstanceTags(instanceId,
...
+  :param bool status: New termination protection status for instance.
```
jcasner: It should be obvious, but let's be explicit that `True` is for on, and `False` is for off

jcasner:
A couple very minor docstring nits, but otherwise :+1:

PR #318, opened 2015-09-06
UNI-148 ModelServer module

marionleborgne:
WIP
@brev a first attempt to spawn a child process (start new model), write, and read data from the stream.


lscheinkman:
@marionleborgne Your editor is still adding extra white-spaces. You should try yo fix them before merging.
You can now use `eslint` for find and fix `lint` issues including extra white-spaces:
```shell
npm run lint
```

unicorn/frontend/browser/app.js
```diff
@@ -120,4 +120,31 @@ document.addEventListener('DOMContentLoaded', () => {
       }
     });
 
+  /* Example to start a model, write, and read data */
```
lscheinkman: @marionleborgne you should write **unit** tests instead. See examples in https://github.com/numenta/numenta-apps/blob/master/unicorn/tests/js/unit/FileServer_test.js

unicorn/frontend/lib/ModelServer.js
```diff
@@ -28,38 +28,64 @@
...
-  // spawn() NuPIC process here
-  this.models[model.modelId] = model;
-  // callback(error, null);
-  callback(null, { model: this.models[model.modelId] });
+ ModelServer.prototype.addModel = function (stats, callback) {
+
+  var modelId = uuid.v1();
+
+  // spawn() NuPIC process 
+  var modelRunnerPath = path.join(__dirname, '..', '..', 'backend', 'unicorn_backend', 'model_runner.py');
```
lscheinkman: This path should be declared as a `const`

unicorn/frontend/lib/ModelServer.js
```diff
@@ -28,38 +28,64 @@
...
+  var child = child_process.spawn('python', [modelRunnerPath, '--model', modelId, '--stats', stats]);
+  var modelInfo = {
+    stdio: child.stdio,
```
lscheinkman: I dont think we need to keep the `stdio` stream

unicorn/frontend/lib/ModelServer.js
```diff
@@ -28,38 +28,64 @@
...
-  this.models[model.modelId] = model;
-  // callback(error, null);
-  callback(null, { model: this.models[model.modelId] });
+ ModelServer.prototype.addModel = function (stats, callback) {
+
+  var modelId = uuid.v1();
+
+  // spawn() NuPIC process 
+  var modelRunnerPath = path.join(__dirname, '..', '..', 'backend', 'unicorn_backend', 'model_runner.py');
+  var child = child_process.spawn('python', [modelRunnerPath, '--model', modelId, '--stats', stats]);
```
lscheinkman: A reference to the `child` process should be kept and associated with the `modelId`, not just it's streams.
The `models` map should just map `modelId` with `child` process

lscheinkman: `child_process.spawn` is an asynchronous operation that returns a [ChildProcess](https://nodejs.org/api/child_process.html#child_process_class_childprocess) object which emit events on errors instead of returning error codes.
Please add error handlers by subscribing to the appropriate events.
See https://nodejs.org/api/child_process.html#child_process_class_childprocess

unicorn/frontend/lib/ModelServer.js
```diff
@@ -28,38 +28,64 @@
...
+};
+
+ModelServer.prototype.addData = function(modelId, data, callback) {
+  var modelStdinStream = this.models[modelId].stdin;
+  modelStdinStream.write(data);  // write to model stdin
+  callback(null, {inputData: data});
+};
+
+ModelServer.prototype.getData = function(modelId, callback) {
+  var modelStdoutStream = this.models[modelId].stdout;
```
lscheinkman: Need to validate if the process is still `connected` and the stream is still valid.
See https://nodejs.org/api/child_process.html#child_process_child_connected

unicorn/frontend/lib/ModelServer.js
```diff
@@ -28,38 +28,64 @@
...
+    stdout : child.stdout,
+    stderr: child.stderr
+  };
+
+  this.models[modelId] = modelInfo;
+  callback(null, {modelId: modelId});
+};
+
+ModelServer.prototype.addData = function(modelId, data, callback) {
+  var modelStdinStream = this.models[modelId].stdin;
```
lscheinkman: Need to validate if the process is still `connected` and the stream is still valid.
See https://nodejs.org/api/child_process.html#child_process_child_connected

unicorn/frontend/lib/ModelServer.js
```diff
@@ -28,38 +28,64 @@
...
+  modelStdinStream.write(data);  // write to model stdin
```
lscheinkman: `stdin` is a [WritableStream](https://nodejs.org/api/stream.html#stream_class_stream_writable).
Please add error handlers by subscribing to the appropriate events.

unicorn/frontend/lib/ModelServer.js
```diff
@@ -28,38 +28,64 @@
...
+  callback(null, {inputData: data});
+};
+
+ModelServer.prototype.getData = function(modelId, callback) {
```
lscheinkman: Due to the asynchronous nature of this method, i think it should be called `onData` instead of `getData`. What do you think?

unicorn/frontend/lib/ModelServer.js
```diff
@@ -28,38 +28,64 @@
...
+  var modelStdoutStream = this.models[modelId].stdout;
+  var data = modelStdoutStream.read();  // read from model stdout
```
lscheinkman: `stdout` is a [Readable](https://nodejs.org/api/stream.html#stream_class_stream_readable) stream object where the [read](https://nodejs.org/api/stream.html#stream_readable_read_size) method will only return chunks of data currently available in the buffer. If there is no data in the buffer, it will return `null`.
To avoid that situation you should only call `read` once the stream emits `readable` event. Something like this:
```javascript
modelStdoutStream.on('readable', function() {
  var data;
  while (null !== (data = readable.read())) {
    callback(null, {outputData: data});
  }
});
```

unicorn/frontend/lib/ModelServer.js
```diff
@@ -28,38 +28,64 @@
...
 };
 
 /**
  *
  */
-ModelServer.prototype.getModels = function (callback) {
+ ModelServer.prototype.getModels = function (callback) {
   // callback(error, null);
-  callback(null, { models: this.models });
+  callback(null, {models: this.models});
```
lscheinkman: Should only return `modelIds`

unicorn/frontend/lib/ModelServer.js
```diff
@@ -67,11 +93,11 @@ ModelServer.prototype.getModel = function (modelId, callback) {
 /**
  *
  */
-ModelServer.prototype.removeModel = function (modelId, callback) {
+ ModelServer.prototype.removeModel = function (modelId, callback) {
```
lscheinkman: Should kill the child process if it is still running. See https://nodejs.org/api/child_process.html#child_process_child_kill_signal

unicorn/frontend/lib/ModelServer.js
```diff
@@ -28,38 +28,64 @@
...
+  callback(null, {outputData: data});
 };
 
 /**
  *
  */
-ModelServer.prototype.getModels = function (callback) {
+ ModelServer.prototype.getModels = function (callback) {
   // callback(error, null);
-  callback(null, { models: this.models });
```
lscheinkman: It should not return `nodejs` objects via `ipc`. Try to return only javascript primitives to avoid memory leaks between the `main` process and the `renderer` process

unicorn/frontend/lib/ModelServer.js
```diff
@@ -28,38 +28,64 @@
...
+  callback(null, {models: this.models});
 };
 
 /**
  *
  */
-ModelServer.prototype.getModel = function (modelId, callback) {
+ ModelServer.prototype.getModel = function (modelId, callback) {
   // callback(error, null);
   callback(null, { model: this.models[modelId] });
```
lscheinkman: It should not return `nodejs` objects via `ipc`. Try to return only javascript primitives to avoid memory leaks between the `main` process and the `renderer` process

brev:
thanks

brev:
Great work @marionleborgne , @lscheinkman thanks for all the help and feedback. I learned a lot here.

PR #319, opened 2015-09-08
TAUR-1301 Removing git (git v1.x) as nta-git (git v2.x) package avaialble.

shantanoo:
Cc: @jcasner, @unixorn


shantanoo:
Merging for now. Will revert back in case git 1.x version is required.

unixorn:
:+1: 

We definitely want v2.x. Thanks for making the change.

PR #320, opened 2015-09-08
Remove stale CentOS 6 git package

unixorn:
We install an up-to-date version of git in our nta-git package; CentOS 6 has a very stale version, so remove it.

@jcasner @shantanoo please CR


jcasner:
:+1:

PR #321, opened 2015-09-08
Use 0.0.0.0 instead of * in supervisord.conf port setting

unixorn:
Address @oxtopus's feedback on [numenta-apps/pull/299](https://github.com/numenta/numenta-apps/pull/299)

We only use IPv4, and given that the way we're using EC2 triggers undesired behavior in supervisord let's go with a configuration that works reliably.

@oxtopus please CR
cc: @jcasner

### DIFF
--- a/taurus.metric_collectors/conf/supervisord.conf
+++ b/taurus.metric_collectors/conf/supervisord.conf
@@ -2,7 +2,7 @@
 file=%(here)s/../supervisor.sock
 
 [inet_http_server]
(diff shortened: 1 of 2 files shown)


oxtopus:
:+1: though, as it turns out, it's due to failure of `socket.gethostbyname()` for custom host name rather than using the assigned dns name.

Thanks!

unixorn:
That's what I deserve for following documentation :grimacing:

PR #325, opened 2015-09-11
ENG-26 Switch (float, int) variant in testFeedInputGetOutput back to just checking for float

vitaly-krugl:
ENG-26 Switch (float, int) variant in testFeedInputGetOutput back to just checking for float since we're now using AnomalyLikelihood class that always returns a float.


scottpurdy:
thanks!

PR #327, opened 2015-09-15
UNI-164: Add Unicorn Fluxible plugin

lscheinkman:
@brev  Please review.
See http://fluxible.io/api/plugins.html

This plugin will expose `ModelClient`, `FileClient`, `DBClient` to `FluxibleContexts`
We can limit access to `actionContext` only if you prefer.
Let me know what you think?


brev:
:+1: interesting, looks good for now, we can edit later when more time :)

PR #328, opened 2015-09-15
UNI-169 We now need to support OS X build slaves

unixorn:
Make sure the CentOS devtools don't get pulled in on non-CentOS machines

@jcasner @shantanoo please CR

### DIFF
--- a/infrastructure/saltcellar/devtools/init.sls
+++ b/infrastructure/saltcellar/devtools/init.sls
@@ -20,7 +20,8 @@
 # ----------------------------------------------------------------------
 # Formula: devtools
 #
-# Install developer tools repository and updated developer tools
+{% if grains['os_family'] == 'RedHat' %}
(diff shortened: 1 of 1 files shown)


infrastructure/saltcellar/devtools/init.sls
```diff
@@ -20,7 +20,8 @@
 # ----------------------------------------------------------------------
 # Formula: devtools
 #
-# Install developer tools repository and updated developer tools
+{% if grains['os_family'] == 'RedHat' %}
+# Install developer tools repository and updated developer tools for CentOS
```
jcasner: Should we indent everything below?  Or is that not the Salt style?

unixorn: It hasn't been in the example code I've looked at, and given that it is yaml I'd prefer to avoid possible indentation aggravation.

PR #329, opened 2015-09-16
ENG-39 Implement safe acceptance tests to execute in the taurus production/staging refresh pipelines

vitaly-krugl:
ENG-39 Implement baseline health check tests that check that all services are in the RUNNING state.

ENG-39 Share retry decorator configuration. Add license text and docstrings to health-check tests.

ENG-39 Implemented Taurus Metric Collectors deployment test resource_accessibility_test.py.

ENG-39 Implemented Taurus Engine deployment/resource_accessibility_test.py

ENG-39 Integrate deployment tests into Taurus pipeline scripts.

I tested the initilize/refresh/update taurus pipeline scripts from my dev laptop against my test ec2 instances. I tested without the `-t destructive-tests` option, as that triggers the failing `results_quality_test.py`.

### DIFF
--- a/taurus.metric_collectors/tests/deployment/health_check_test.py
+++ b/taurus.metric_collectors/tests/deployment/health_check_test.py
@@ -1,0 +1,82 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have purchased from
+# Numenta, Inc. a separate commercial license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero Public License for more details.
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""Baseline non-destructive health checks of the running Taurus Metric
+Collectors server.
+"""
+
+
+import unittest
+from urlparse import urljoin
+import xmlrpclib
+
+
+from nta.utils import error_handling
+
+
+
+# Max time to wait for service to enter RUNNING state
+_RUNNING_TIMEOUT_SEC = 5
+
+
+_METRICS_COLLECTOR_SUPERVISOR_API_URL = urljoin("http://localhost:8001", "RPC2")
+
+
+_RETRY_SERVICE_RUNNING_CHECK = error_handling.retry(
+  _RUNNING_TIMEOUT_SEC,
+  initialRetryDelaySec=1,
+  maxRetryDelaySec=1)
+
+
+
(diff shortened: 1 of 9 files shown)


vitaly-krugl:
@oxtopus Please review.

CC @scottpurdy @jcasner

taurus.metric_collectors/tests/deployment/health_check_test.py
```diff
@@ -0,0 +1,82 @@
...
+    :param str serviceName: name of service
+    :returns: State name of the service; see
+      http://supervisord.org/subprocess.html#process-states
+    :rtype: str
+    """
+    server = xmlrpclib.Server(_METRICS_COLLECTOR_SUPERVISOR_API_URL)
+
+    info = server.supervisor.getProcessInfo(serviceName)
+
+    return info["statename"]
```
oxtopus: Is it possible for `info = server.supervisor.getProcessInfo(serviceName)` to not raise an exception, but otherwise result in the `statename` key not being in `info`?  We've encountered similar such cases with other external APIs where there was a non-obvious failure that resulted in keys not being present in the return value of an API call.

For our purposes, the risk is mitigated because the test will still fail, but the reason won't be obvious.

taurus.metric_collectors/tests/deployment/resource_accessibility_test.py
```diff
@@ -0,0 +1,142 @@
+# ----------------------------------------------------------------------
```
oxtopus: At the risk of breaking from convention, I'd rather see such tests _in_ the services themselves, which they call upon startup and shutdown upon failure.  This file would become a light wrapper that calls those sanity checks.

taurus.metric_collectors/tests/deployment/health_check_test.py
vitaly-krugl: @oxtopus: I've seen your code that checks for the presence of the fields. There is not much else that we can do in these tests if the field is missing. The test methods will retry on any Exception (up to a certain time limit), so if this is transient, then a subsequent retry should get the correct value; if none of the retries succeed, then the test will fail, and rightfully so.

taurus.metric_collectors/tests/deployment/resource_accessibility_test.py
vitaly-krugl: @oxtopus: the resources might be accessible when the service launches, but then become inaccessible later. The issue might be transient and recover at a later time. I think that resource accessibility and other service health issues are better served by an integrated health reporting/monitoring system, which would require buy-in from product management.

CC @scottpurdy, @jcasner

oxtopus: I'm only advocating the logic be moved into the service itself.  Still need to regularly run the tests as planned, but the test code is just a wrapper around the service sanity checks rather than a full implementation that makes certain assumptions about the underlying implementation that may not always be true.  I don't think that ought to involve buy-in from anyone but you and me.

vitaly-krugl: I consider modification of the service logic to be outside the scope of ENG-39. My preference is to give the overall health-checking logic more thought and tackle it separately from this task.

oxtopus: I'm not advocating that we change the service logic, either, just that sanity checks added in the service rather than as separate tests.

vitaly-krugl: @oxtopus: I interpreted 

> I'd rather see such tests in the services themselves, which they call upon startup and shutdown upon failure.

in your earlier comment as advocating changes to service logic. Regardless, let's tackle further changes, if any, in this area in a separate task.

PR #335, opened 2015-09-21
UNI-169 Multi os support

unixorn:
Make motd formula work on both CentOS and OSX
* Set group ownership from root (CentOS only) to wheel, which exists on both platforms
* Get acta-diurna directly from GH instead of via an RPM

@lscheinkman, @shantanoo, @jcasner please CR


jcasner:
:+1:

PR #341, opened 2015-09-24
ENG-67 Fix race condition that stalls tweet collector

vitaly-krugl:
ENG-67 We've experienced race condition hangs when the non-empty _EngineSingleton was cloned via multiprocessing forking. `cls._engine.dispose()` didn't alleviate this problem. The fix is
to fork the worker processes before anything in the main process invokes collectorsdb.engineFactory().

ENG-67 Fix unit tests to expect AssertionError when pid doesn't match the current pid.

ENG-67 In testEngineFactorySingletonPattern: expect AssertionError from collectorsdb.engineFactory() after populating the singleton in current process and requesting the engine again from a forked process

### DIFF
--- a/taurus.metric_collectors/taurus/metric_collectors/collectorsdb/__init__.py
+++ b/taurus.metric_collectors/taurus/metric_collectors/collectorsdb/__init__.py
@@ -19,9 +19,11 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


vitaly-krugl:
Please review

CC @scottpurdy, @oxtopus

oxtopus:
LGTM :+1: Thanks!

PR #347, opened 2015-09-24
UNI-197: Limit concurrent models to 2

lscheinkman:
@brev Please review when you have a chance


brev:
@lscheinkman please merge now if you feel ok about it, I can't review until I'm done with my DB task, may be a bit sorry

brev:
@lscheinkman reviewing now

unicorn/frontend/browser/actions/AddModel.js
```diff
@@ -18,6 +18,7 @@
 // http://numenta.org/licenses/
 
 'use strict';
+import {ACTIONS} from '../lib/Constants';
```
brev: what do you think about putting the Action constants in the shared global config I created?

brev:
:+1: thanks for the IPC example in ModelClient/Server, I might borrow some of this for the DB client/server

unicorn/frontend/browser/actions/AddModel.js
lscheinkman: @brev Where is the "global config" file? I can only find the `ConfigServer.js` and `ConfigClient.js`?

brev: sure, I was just thinking of adding them to config, so it's all in a single place. no big deal either way, just 1 less file.

https://github.com/numenta/numenta-apps/blob/master/unicorn/frontend/config/default.json
https://github.com/numenta/numenta-apps/blob/master/unicorn/frontend/config/

thanks

PR #349, opened 2015-09-25
ENG-70 Return log-scaled anomaly likelihood

oxtopus:
@lscheinkman


lscheinkman:
:+1: @oxtopus Thank you

PR #351, opened 2015-09-29
UNI-169 Make salt blow away the conflicting symlinks explicitly

unixorn:
Salt won't replace a symlink with a file, so add states to clear the stale versions out of the way first.

@shantanoo, @lscheinkman, please CR.

### DIFF
--- a/infrastructure/saltcellar/motd/init.sls
+++ b/infrastructure/saltcellar/motd/init.sls
@@ -105,6 +105,16 @@ update-motd:
     - require:
       - file: acta-diurna
 
+clean-update-motd-script-source:
+  cmd.run:
+    - name: rm -f /usr/local/sbin/update-motd
+    - onlyif: test -L /usr/local/sbin/update-motd
+
+cleanup-motd-cronjob-symlink:
+  cmd.run:
+    - name: rm -f /etc/cron.daily/update-motd
+    - unless: test -L /etc/cron.daily/update-motd
+
 {% if grains['os_family'] == 'RedHat' %}
 
 motd-cronjob:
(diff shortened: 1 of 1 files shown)


infrastructure/saltcellar/motd/init.sls
```diff
@@ -95,6 +95,8 @@ update-motd:
     - group: wheel
     - mode: 0755
     - require:
+      - cmd: clean-update-motd-script-source
```
shantanoo: According to [file.managed](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html#salt.states.file.managed)
```
replace
    If this file should be replaced. If false, this command will not overwrite file contents but 
    will enforce permissions if the file exists already. Default is True.
```

Is there a bug in salt itself since it is not overwriting the file?

infrastructure/saltcellar/motd/init.sls
```diff
@@ -103,6 +105,16 @@ update-motd:
     - require:
       - file: acta-diurna
 
+clean-update-motd-script-source:
+  cmd.run:
+    - name: rm -f /usr/local/sbin/update-motd
```
shantanoo: Instead of using `rm`, would prefer using `unlink`.

infrastructure/saltcellar/motd/init.sls
```diff
@@ -95,6 +95,8 @@ update-motd:
     - group: wheel
     - mode: 0755
```
shantanoo: `s/0755/0644/`? Same for the other `motd` files.

infrastructure/saltcellar/motd/init.sls
unixorn: The AMI has the `update-motd` script in `/etc/cron.daily` and a symlink to that in `/usr/local/sbin`.  When the new formulas are applied, salt chokes on replacing files with symlinks and vice-versa.

I'll have to take a look at adding the replace clause, I overlooked that in the docs

infrastructure/saltcellar/motd/init.sls
unixorn: Why `unlink` over `rm`?

infrastructure/saltcellar/motd/init.sls
unixorn: They're executable scripts. If they don't have the executable bit turned on, the driver script will treat them as text and add their source code to `/etc/motd` instead of the output it gets from running them.

PR #357, opened 2015-10-01
Use NullHandler for now to avoid getting the unwanted unformatted warning message from logger on stderr "No handlers could be found for logger".

vitaly-krugl:


vitaly-krugl:
Hotfix: fix accidental breakage.

PR #360, opened 2015-10-01
TAUR-1301 Change the ownership for /opt/numenta/anaconda

shantanoo:
Change ownership of /opt/numenta/anaconda according to the setting in /etc/numenta/anaconda-owner. If file doesn't exist, use ec2-user:ec2-user.

Cc: @jcasner, @unixorn


unixorn:
:+1:

PR #362, opened 2015-10-02
ENG-62 Implement an additional taurus metric collectors deployment resource accessibility test

vitaly-krugl:
ENG-62 Implement an additional taurus metric collectors deployment resource accessibility test that validates the account's authorization to send emails (used for error-reporting) via the collector's SES configuration.

NOTE: This bit us during the initial deployment of the integrated stock symbol and twitter handle services.


vitaly-krugl:
CC @oxtopus

PR #365, opened 2015-10-02
UNI-187 #comment code review fixes

brev:
UNI-187 #comment code review fixes
Please see each commit msg.
@lscheinkman @marionleborgne

### DIFF
--- a/unicorn/tests/js/unit/FileServer_test.js
+++ b/unicorn/tests/js/unit/FileServer_test.js
@@ -19,9 +19,13 @@
 
 'use strict';
 
-import FileServer from '../../../frontend/lib/FileServer';
-import path from 'path';
+
 const assert = require('assert');
+
+import path from 'path';
+
+import FileServer from '../../../frontend/lib/FileServer';
+
 
 // Contents of 'fixture/file.csv'
 const EXPECTED_CONTENT =
@@ -34,7 +38,6 @@ const EXPECTED_CONTENT =
 2015-08-26T19:51:31+17:00,19
 `;
 
-
 // Expected data
 const EXPECTED_DATA = [
   {timestamp: '2015-08-26T19:46:31+17:00', metric: '21'},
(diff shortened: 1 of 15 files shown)


brev:
ready @lscheinkman, thanks. no rush.

lscheinkman:
@brev Once the test pass it should be good to go :+1:

unicorn/tests/js/unit/FileServer_test.js
```diff
@@ -47,8 +47,18 @@ const EXPECTED_DATA = [
 
 // Expected fields
 const EXPECTED_FIELDS = [
-  {name: 'timestamp', type: 'date'},
-  {name: 'metric', type: 'number'},
+  {
+    name: 'timestamp',
+    type: 'date',
+    uid: '8cc52a23d53b19cb8c2618c4118dc6b40c4ebc87',
```
lscheinkman: @brev You may need to refactor the so it can handle `uid`

lscheinkman: @brev instead of `assert.deepEqual` on [line 103](
https://github.com/numenta/numenta-apps/pull/365/files#diff-7f165428018da851cb7b84e46cd6adfcL103) you may need to just check for the presence of the `uid` and `file_uid` properties.
Something like this:

```javascript
  describe('#getFields', () => {
    it('Get fields using default options', (done) => {
      server.getFields(FILENAME, (error, fields) => {
        assert.ifError(error);
        fields.forEach((actual, idx, arr) => {
          let expected = EXPECTED_FIELDS[idx];
          assert.equal(actual.name, expected.name);
          assert.equal(actual.type, expected.type);
          // Check uid presence not value
          assert('uid' in actual);
          assert('file_uid' in actual);
        });
        done();
      });
    });
  });

```

brev:
@lscheinkman the test just happened to be passing for me every time I ran it, doh. I've tried another fix here, what do you think? Also refactored the UserError subclasses a bit more. thanks

lscheinkman:
:+1: Your version is better. thanks

PR #366, opened 2015-10-02
Grok to HTM

darian19:
I ran into some trouble posting the script I used into the comments, but I can send it if needed. When I was using my script on the repository, I made a commit every few folders just to check for any problems that might have popped up. There were a couple of files that were being ignored because of a .gitignore file, but I believe I was able to get all of those back in. I also had my script ignore files with the following extensions: '.gif', '.png', '.jar', '.so', '.zip', '.eot', '.ttf', '.woff', '.psd', '.jpg', '.swf', '.ico', '.class'. I do not think these files were being edited, but I noticed they were showing up in the comparison between repositories and were binary files, so I chose to have them ignored. The entire diff isn't showing up because the last commit changed so many files, but when you go through each commit there is an equal number of additions and deletions.


darian19:
The last commit causes that huge discrepancy in the diff, but I'm not sure why since it only renames files. Before that commit is made, the diff breaks even.

jcasner:
With over 4000 files changed, it's impossible to review this effectively.  I'm going to go ahead and merge it and we'll fix anything that breaks as a result.  

FYI @shantanoo @lscheinkman @brev @marionleborgne @unixorn @oxtopus @scottpurdy @BoltzmannBrain @subutai @cbaranski @vitaly-krugl @chetan51

PR #370, opened 2015-10-06
UNI-190 #comment Make chart fixed-width, so data does not "accordian"…

brev:
UNI-190 #comment Make chart fixed-width, so data does not "accordian" while loading. Went with 200 timepoints for now.


brev:
ready for review @marionleborgne @lscheinkman @jcasner

unicorn/frontend/browser/components/Chart.jsx
```diff
@@ -51,7 +53,13 @@ export default class Chart extends React.Component {
 
   constructor(props, context) {
     super(props, context);
+
+    // DyGraphs chart container
     this._dygraph = null;
+
+    // Chart Range finder value: For Fixed-width-chart & auto-scroll-to-right
+    this._chartXrange = 200;  // static 200px
+    // this._chartXrange = Math.floor((graphXmax - graphXmin) * 0.2); // dynamic
```
lscheinkman: Did you mean to leave this line commented out here?

marionleborgne:
@brev Just tried it. The UX is much nicer!

marionleborgne:
There is an issue (though I am not sure this is due to your change @brev) that happens when I select multiple metrics at the same time: The graphs get all scrambled like if the timestamps got shuffled. Do you get that too?

brev:
@marionleborgne I've never tried that, I'll look in the morn thanks

unicorn/frontend/browser/components/Chart.jsx
```diff
@@ -64,14 +72,19 @@ export default class Chart extends React.Component {
   componentWillUnmount() {
     if (this._dygraph) {
       this._dygraph.destroy();
-      this._dygraph = null;
+      delete this._dygraph;
     }
+    delete this._chartXrange;
```
lscheinkman: Did you mean to delete the `_chartXrange` property or just reset the value?
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete

lscheinkman:
:+1: LGTM. It looks great!

unicorn/frontend/browser/components/Chart.jsx
brev: @lscheinkman I was treating it as the same, but forgot about the context. Nice catch, thanks.

brev:
merging now, I have issues noted in email to investigate, I'll file jiras where necessary. thanks all.

unicorn/frontend/browser/components/Chart.jsx
```diff
@@ -66,12 +73,19 @@ export default class Chart extends React.Component {
       this._dygraph.destroy();
       this._dygraph = null;
     }
+    if (this._chartXrange) {
+      this._chartXrange = null;
```
lscheinkman: You probably need to initialize the value on `componentDidMount`. You can the `mount/umount` events fired by collapsing the chart on the UI

PR #379, opened 2015-10-07
ENG-34 Check all models for recent data, using simple statistical mod…

oxtopus:
…el to determine if a model appears to not be updating

This will also require updates to internal infrastructure templates to include the right dynamodb credentials.

### DIFF
--- a/taurus.monitoring/README.md
+++ b/taurus.monitoring/README.md
@@ -13,7 +13,7 @@ local changes without conflicts:
 
     cp -r conf conf-user
 
-Alternatively, the configureation files' path can be overidden using the
+Alternatively, the configuration files' path can be overidden using the
 TAURUS_MONITORS_DB_CONFIG_PATH environment variable.
 
 ### First, install `nta.utils`.  Then, to install `taurus.monitors`:
--- a/taurus.monitoring/README.md
+++ b/taurus.monitoring/README.md
@@ -13,7 +13,7 @@ local changes without conflicts:
 
     cp -r conf conf-user
 
-Alternatively, the configureation files' path can be overidden using the 
+Alternatively, the configureation files' path can be overidden using the
 TAURUS_MONITORS_DB_CONFIG_PATH environment variable.
 
 ### First, install `nta.utils`.  Then, to install `taurus.monitors`:
@@ -30,9 +30,9 @@ TAURUS_MONITORS_DB_CONFIG_PATH environment variable.
 
 ### Set up `monitorsdb` sql database connection
 
-1. The configuration in `conf/taurus-monitors-sqldb.conf` defaults to 
-host=localhost with the default user/password for mysql. To override mysql 
-host, username and password, use the following command, substituting the 
+1. The configuration in `conf/taurus-monitors-sqldb.conf` defaults to
+host=localhost with the default user/password for mysql. To override mysql
+host, username and password, use the following command, substituting the
 appropriate strings for HOST, USER and PASSWORD:
 ```
     >>> taurus-set-monitorsdb-login --host=HOST --user=USER --password=PASSWORD
--- a/taurus.monitoring/taurus/monitoring/__init__.py
+++ b/taurus.monitoring/taurus/monitoring/__init__.py
@@ -21,10 +21,10 @@
 
 # See http://peak.telecommunity.com/DevCenter/setuptools#namespace-packages
 try:
-    __import__("pkg_resources").declare_namespace(__name__)
+  __import__("pkg_resources").declare_namespace(__name__)
 except ImportError:
-    from pkgutil import extend_path
-    __path__ = extend_path(__path__, __name__)
+  from pkgutil import extend_path
+  __path__ = extend_path(__path__, __name__)
 
 
 from optparse import OptionParser
--- a/taurus.monitoring/taurus/monitoring/__init__.py
+++ b/taurus.monitoring/taurus/monitoring/__init__.py
@@ -80,7 +80,6 @@ def loadEmailParamsFromConfig(config):
 
 
 
-
 class MonitorOptionParser(OptionParser):
   """ OptionParser subclass that implements a --monitorConfPath default
   option and adds a parse_options() helper function to enforce behavior with
@@ -115,4 +114,3 @@ class MonitorOptionParser(OptionParser):
                         .format(repr(args)))
 
     return options
-
--- a/taurus.monitoring/taurus/monitoring/__init__.py
+++ b/taurus.monitoring/taurus/monitoring/__init__.py
@@ -46,13 +46,28 @@ if CONF_DIR is None:
   CONF_DIR = os.path.join(TAURUS_MONITORS_HOME, "conf")
 
 
+
 def loadConfig(options):
+  """ Load, and return a Config object given a OptionParser object
+
+  :param object options: Object having `monitorConfPath` attr representing the
+    path to a configuration file.
+  :returns: Config object
+  :rtype: nta.utils.config.Config
+  """
   confDir = os.path.dirname(options.monitorConfPath)
   confFileName = os.path.basename(options.monitorConfPath)
   return Config(confFileName, confDir)
 
 
+
 def loadEmailParamsFromConfig(config):
+  """ Extract email-related configuration directives from a Config object
+
+  :params nta.utils.config.Config config:
+  :returns: Dict having `senderAddress`, `recipients`, `awsRegion`,
+    `sesEndpoint`, `awsAccessKeyId`, and `awsSecretAccessKey` keys.
+  """
   return dict(
     senderAddress=(
       config.get("S1", "MODELS_MONITOR_EMAIL_SENDER_ADDRESS")),
@@ -67,8 +82,13 @@ def loadEmailParamsFromConfig(config):
 
 
 class MonitorOptionParser(OptionParser):
+  """ OptionParser subclass that implements a --monitorConfPath default
+  option and adds a parse_options() helper function to enforce behavior with
+  respect to positional arguments.  Monitors that accept command line arguments
+  should use this instead of the standard OptionParser class for consistency.
+  """
+
   def __init__(self, *args, **kwargs):
-    #super(self.__class__, self).__init__(*args, **kwargs)
     OptionParser.__init__(self, *args, **kwargs)
     self.add_option("--monitorConfPath",
                     help=("Specify full path to ConfigParser-compatible"
@@ -80,7 +100,14 @@ class MonitorOptionParser(OptionParser):
                           "MODELS_MONITOR_EMAIL_SES_ENDPOINT"))
 
 
-  def parse_args(self, *args, **kwargs):
+  def parse_options(self, *args, **kwargs):
+    """ Parse and return command line options.  Enforcing a common convention
+    in which positional arguments are not allowed, allowing only named
+    arguments
+
+    :returns: First result of OptionParser.parse_args()
+    :rtype: object
+    """
     (options, args) = OptionParser.parse_args(self, *args, **kwargs)
 
     if args:
--- a/taurus.monitoring/taurus/monitoring/__init__.py
+++ b/taurus.monitoring/taurus/monitoring/__init__.py
@@ -27,10 +27,12 @@ except ImportError:
     __path__ = extend_path(__path__, __name__)
 
 
+from optparse import OptionParser
 import os
 from pkg_resources import get_distribution
 
 from nta.utils import logging_support_raw
+from nta.utils.config import Config
 
 
 
@@ -42,3 +44,48 @@ TAURUS_MONITORS_HOME = os.path.abspath(distribution.location)
 CONF_DIR = os.environ.get("TAURUS_MONITORS_DB_CONFIG_PATH")
 if CONF_DIR is None:
   CONF_DIR = os.path.join(TAURUS_MONITORS_HOME, "conf")
+
+
+def loadConfig(options):
+  confDir = os.path.dirname(options.monitorConfPath)
+  confFileName = os.path.basename(options.monitorConfPath)
+  return Config(confFileName, confDir)
+
+
+def loadEmailParamsFromConfig(config):
+  return dict(
+    senderAddress=(
+      config.get("S1", "MODELS_MONITOR_EMAIL_SENDER_ADDRESS")),
+    recipients=config.get("S1", "MODELS_MONITOR_EMAIL_RECIPIENTS"),
+    awsRegion= config.get("S1", "MODELS_MONITOR_EMAIL_AWS_REGION"),
+    sesEndpoint=config.get("S1", "MODELS_MONITOR_EMAIL_SES_ENDPOINT"),
+    awsAccessKeyId=None,
+    awsSecretAccessKey=None
+  )
+
+
+
+
+class MonitorOptionParser(OptionParser):
+  def __init__(self, *args, **kwargs):
+    #super(self.__class__, self).__init__(*args, **kwargs)
+    OptionParser.__init__(self, *args, **kwargs)
+    self.add_option("--monitorConfPath",
+                    help=("Specify full path to ConfigParser-compatible"
+                          " monitor conf file, containing a [S1] section and"
+                          " the following configuration directives:\n\n"
+                          "MODELS_MONITOR_EMAIL_SENDER_ADDRESS\n"
+                          "MODELS_MONITOR_EMAIL_RECIPIENTS\n"
+                          "MODELS_MONITOR_EMAIL_AWS_REGION\n"
+                          "MODELS_MONITOR_EMAIL_SES_ENDPOINT"))
+
+
+  def parse_args(self, *args, **kwargs):
+    (options, args) = OptionParser.parse_args(self, *args, **kwargs)
+
+    if args:
+      self.parser.error("Unexpected positional arguments: {}"
+                        .format(repr(args)))
+
+    return options
+
--- a/taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
+++ b/taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
@@ -60,30 +60,38 @@ class LatencyMonitorError(Exception):
 
 class ModelLatencyChecker(MonitorDispatcher):
   parser = MonitorOptionParser(
-    usage="Usage: %prog --table=TABLE [options]",
+    usage="Usage: %prog --metricDataTable=TABLE [options]",
     description=("Monitoring script to alert on models that do not have recent"
-                 " data in public dynamodb table.  A model is considered to "
-                 "have exceeded the acceptable threshold if the time since the"
-                 "most recent sample exceeds a minimum, user-defined threshold"
-                 " (in seconds) AND such interval exceeds a multiple of the "
-                 "standard deviation for all intervals for that model during "
-                 "a fixed window of time (14 days)."))
+                 " data in public metric data dynamodb table.  A model is "
+                 "considered to have exceeded the acceptable threshold if the "
+                 "time since the most recent sample exceeds a minimum, "
+                 "user-defined threshold (in seconds) AND such interval "
+                 "exceeds a multiple of the standard deviation for all "
+                 "intervals for that model during a fixed window of time (14 "
+                 "days)."))
 
   parser.add_option("--threshold",
                     default=MIN_THRESHOLD,
                     type="int",
                     dest="threshold",
                     metavar="SECONDS",
-                    help="Default: {}".format(MIN_THRESHOLD))
-  parser.add_option("--sigma",
+                    help=("Required minimum number of seconds to pass since "
+                          "most-recent known timestamp before a model may be "
+                          "_considered_ as not having recent data "
+                          "(Default: {})").format(MIN_THRESHOLD))
+  parser.add_option("--sigmaMultipler",
                     default=SIGMA_MULTIPLIER,
                     type="int",
-                    dest="sigma",
-                    help="Default: {}".format(SIGMA_MULTIPLIER))
-  parser.add_option("--table",
+                    dest="sigmaMultipler",
+                    help=("Standard deviation multiplier. Consider 68–95–99.7 "
+                          "rule.  For example, 99.7% of values fall within 3 "
+                          "standard deviations of the mean.  Any value greater"
+                          " than 3 x stddev is considered exceptional. "
+                          "(Default: {})").format(SIGMA_MULTIPLIER))
+  parser.add_option("--metricDataTable",
                     type="string",
-                    dest="table",
-                    help="**REQUIRED dynamodb table name**")
+                    dest="metricDataTable",
+                    help="**REQUIRED metric data dynamodb table name**")
   parser.add_option("--days",
                     default=FIXED_WINDOW,
                     type="int",
@@ -105,11 +113,11 @@ class ModelLatencyChecker(MonitorDispatcher):
     self.awsSecretAccessKey = self.config.get(
       "S1", "MODELS_MONITOR_TAURUS_DYNAMODB_AWS_SECRET_ACCESS_KEY")
     self.threshold = options.threshold
-    self.sigma = options.sigma
+    self.sigmaMultipler = options.sigmaMultipler
 
-    if not options.table:
-      self.parser.error("You must specify a --table argument.")
-    self.table = options.table
+    if not options.metricDataTable:
+      self.parser.error("You must specify a --metricDataTable argument.")
+    self.metricDataTable = options.metricDataTable
 
     self.days = options.days
 
@@ -156,7 +164,7 @@ def checkAllModelLatency(monitorObj):
     aws_access_key_id=monitorObj.awsAccessKeyId,
     aws_secret_access_key=monitorObj.awsSecretAccessKey)
 
-  table = Table(monitorObj.table, connection=conn)
+  metricDataTable = Table(monitorObj.metricDataTable, connection=conn)
 
   errors = []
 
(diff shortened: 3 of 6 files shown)


oxtopus:
<a href="https://imgflip.com/i/s8rkt"></a>

@vitaly-krugl @scottpurdy @jcasner @unixorn

taurus.monitoring/README.md
```diff
@@ -13,7 +13,7 @@ local changes without conflicts:
 
     cp -r conf conf-user
 
-Alternatively, the configureation files' path can be overidden using the 
+Alternatively, the configureation files' path can be overidden using the
```
unixorn: s/configureation/configuration/

scottpurdy:
Would be nice to hav @vitaly-krugl do a review of the functional changes. Code style wise you could do a quick pass and be consistent about whitespace, etc.

One change that wasn't introduced in this PR is putting the command line parser as a class attribute while I think it makes more sense in a main function so it only gets built when running from the command line.

oxtopus:
@scottpurdy the class is the right place for the parser, IMO.  There is a subclass relationship in which additional options are added in the subclass(es).  It's not stateful, so it shouldn't be an instance attr.

taurus.monitoring/taurus/monitoring/monitor_dispatcher.py
```diff
@@ -0,0 +1,76 @@
...
+
+    :param function checkFn: The check function that raised an exception
+    :param type excType: Exception type
+    :param exception excValue: Exception value
+    :param traceback excTraceback: Exception traceback
+    """
+    pass
+
+
+  def checkAll(self):
```
vitaly-krugl: Is this framework doing the rough equivalent of unittest as far as identifying what to run and running it is concerned?

taurus.monitoring/taurus/monitoring/__init__.py
```diff
@@ -42,3 +44,48 @@
...
+    senderAddress=(
+      config.get("S1", "MODELS_MONITOR_EMAIL_SENDER_ADDRESS")),
+    recipients=config.get("S1", "MODELS_MONITOR_EMAIL_RECIPIENTS"),
+    awsRegion= config.get("S1", "MODELS_MONITOR_EMAIL_AWS_REGION"),
+    sesEndpoint=config.get("S1", "MODELS_MONITOR_EMAIL_SES_ENDPOINT"),
+    awsAccessKeyId=None,
+    awsSecretAccessKey=None
+  )
+
+
```
vitaly-krugl: too many blank lines

taurus.monitoring/taurus/monitoring/__init__.py
```diff
@@ -42,3 +44,48 @@
 CONF_DIR = os.environ.get("TAURUS_MONITORS_DB_CONFIG_PATH")
 if CONF_DIR is None:
   CONF_DIR = os.path.join(TAURUS_MONITORS_HOME, "conf")
+
+
+def loadConfig(options):
+  confDir = os.path.dirname(options.monitorConfPath)
+  confFileName = os.path.basename(options.monitorConfPath)
+  return Config(confFileName, confDir)
+
```
vitaly-krugl: Not enough blank lines; needs to be 3 between global-scope objects

taurus.monitoring/taurus/monitoring/__init__.py
```diff
@@ -42,3 +44,48 @@
...
+    awsAccessKeyId=None,
+    awsSecretAccessKey=None
+  )
+
+
+
+
+class MonitorOptionParser(OptionParser):
+  def __init__(self, *args, **kwargs):
+    #super(self.__class__, self).__init__(*args, **kwargs)
```
vitaly-krugl: remove dead code

vitaly-krugl:
Do all the changes pass pylint?

taurus.monitoring/taurus/monitoring/__init__.py
```diff
@@ -42,3 +44,48 @@
...
+
+
+def loadEmailParamsFromConfig(config):
+  return dict(
+    senderAddress=(
+      config.get("S1", "MODELS_MONITOR_EMAIL_SENDER_ADDRESS")),
+    recipients=config.get("S1", "MODELS_MONITOR_EMAIL_RECIPIENTS"),
+    awsRegion= config.get("S1", "MODELS_MONITOR_EMAIL_AWS_REGION"),
+    sesEndpoint=config.get("S1", "MODELS_MONITOR_EMAIL_SES_ENDPOINT"),
+    awsAccessKeyId=None,
```
vitaly-krugl: Why initialize awsAccessKeyId and awsSecretAccessKey at all if it merely sets them None?

taurus.monitoring/taurus/monitoring/__init__.py
```diff
@@ -42,3 +44,48 @@
...
+    awsSecretAccessKey=None
+  )
+
+
+
+
+class MonitorOptionParser(OptionParser):
```
vitaly-krugl: Need docstring to explain purpose and anticipated usage of this class.

taurus.monitoring/taurus/monitoring/__init__.py
```diff
@@ -42,3 +44,48 @@
...
+
+
+  def parse_args(self, *args, **kwargs):
+    (options, args) = OptionParser.parse_args(self, *args, **kwargs)
+
+    if args:
+      self.parser.error("Unexpected positional arguments: {}"
+                        .format(repr(args)))
+
+    return options
```
vitaly-krugl: I don't like this change to the return value signature of the builtin method.

taurus.monitoring/taurus/monitoring/__init__.py
```diff
@@ -42,3 +44,48 @@
...
+                    help=("Specify full path to ConfigParser-compatible"
+                          " monitor conf file, containing a [S1] section and"
+                          " the following configuration directives:\n\n"
+                          "MODELS_MONITOR_EMAIL_SENDER_ADDRESS\n"
+                          "MODELS_MONITOR_EMAIL_RECIPIENTS\n"
+                          "MODELS_MONITOR_EMAIL_AWS_REGION\n"
+                          "MODELS_MONITOR_EMAIL_SES_ENDPOINT"))
+
+
+  def parse_args(self, *args, **kwargs):
```
vitaly-krugl: Missing documentation of args and return value

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+from nta.utils import error_reporting
+
+from taurus.monitoring.monitor_dispatcher import MonitorDispatcher
+from taurus.monitoring import (loadConfig,
+                               loadEmailParamsFromConfig,
+                               MonitorOptionParser)
+
+
+
+MIN_THRESHOLD = 3600 # Required minimum number of seconds to pass since last
```
vitaly-krugl: Could benefit from a more descriptive name; "threshold" is too generic.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+                     # known timestamp before a model may be _considered_ as
+                     # not having recent data
+SIGMA_MULTIPLIER = 3 # Standard deviation multiplier
```
vitaly-krugl: Why is this a good value?

taurus.monitoring/taurus/monitoring/__init__.py
oxtopus: We explicitly don't want to allow positional args.  If args are detected, the program exits in an error state.  Given that, it doesn't make sense to maintain consistency with the original `OptionParser.parse_args()` method since there will never be a value returned.

That is, unless you think it makes more sense to `return options, None` or similar, knowing that the second argument will always be ignored.

vitaly-krugl: I don't think that `return options, None` is a good idea at all. However, the lack of documentation that describes the return value (that differs from the standard return value of the base class) as well as the lack of justification for not allowing positional args is not helpful.

oxtopus: Not allowing positional args is the status quo.  I'm only moving it from https://github.com/numenta/numenta-apps/pull/379/files#diff-6a2c6f52a97fa5e544a289148301e848L70 into this class so I can reuse the logic without repeating the code, or implementing an inconsistent pattern.

I'm renaming it to `parse_options()` in followup so as to avoid further confusion with `OptionParser.parse_args()`

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+  def __init__(self):
+    options = self.parser.parse_args()
+
+    self.config = loadConfig(options)
+    self.emailParams = loadEmailParamsFromConfig(self.config)
+    self.apiKey = self.config.get("S1", "MODELS_MONITOR_TAURUS_API_KEY")
+    self.modelsUrl = self.config.get("S1", "MODELS_MONITOR_TAURUS_MODELS_URL")
+    self.region = self.config.get(
```
vitaly-krugl: For readability, something like dynamodbRegion would be more descriptive; see the likes of awsAccessKeyId and awsSecretAccessKey: they would be much worse without the "aws" prefix.

taurus.monitoring/taurus/monitoring/__init__.py
oxtopus: idunno.  That's what it was previously at https://github.com/numenta/numenta-apps/pull/379/files#diff-6a2c6f52a97fa5e544a289148301e848L87

Again, I'm only shuffling code around in this case to keep DRY.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+      resourceName=repr(self),
+      message=self.formatTraceback(excType, excValue, excTraceback),
+      subjectPrefix="Model Latency Monitor",
+      params=self.emailParams
+    )
+
+
+
+@ModelLatencyChecker.registerCheck
+def checkAllModelLatency(monitorObj):
```
vitaly-krugl: Logically, this function belongs inside the `ModelLatencyChecker` class

oxtopus: The convention was established prior in the supervisor monitor.  I've refactored the implementation there so that it may be re-used here (and elsewhere).

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+    self.awsAccessKeyId = self.config.get(
+      "S1", "MODELS_MONITOR_TAURUS_DYNAMODB_AWS_ACCESS_KEY_ID")
+    self.awsSecretAccessKey = self.config.get(
+      "S1", "MODELS_MONITOR_TAURUS_DYNAMODB_AWS_SECRET_ACCESS_KEY")
+    self.threshold = options.threshold
+    self.sigma = options.sigma
+
+    if not options.table:
+      self.parser.error("You must specify a --table argument.")
+    self.table = options.table
```
vitaly-krugl: Need a more descriptive name for self.table for better code readability. We have multiple dynamodb tables in taurus.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+                    help="Default: {}".format(MIN_THRESHOLD))
+  parser.add_option("--sigma",
+                    default=SIGMA_MULTIPLIER,
+                    type="int",
+                    dest="sigma",
+                    help="Default: {}".format(SIGMA_MULTIPLIER))
+  parser.add_option("--table",
+                    type="string",
+                    dest="table",
+                    help="**REQUIRED dynamodb table name**")
```
vitaly-krugl: Which one? We have multiple dynamodb tables in Taurus.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+
+  table = Table(monitorObj.table, connection=conn)
+
+  errors = []
+
+  for model in response.json():
+    # Query recent DynamoDB metric data for each model
+    then = str(datetime.datetime.now() -
+               datetime.timedelta(days=monitorObj.days))
+    resultSet = table.query_2(uid__eq=model["uid"], timestamp__gte=then[:19])
```
vitaly-krugl: Make the timestamp encoding explicit via `datetime.strftime()`. The `str(datetime.datetime.now() ...` combined with `then[:19]` is a substandard technique.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+    raise LatencyMonitorError("Unable to query Taurus API for active models: "
+                              "Unexpected HTTP response status ({}) from {}"
+                              .format(response.status_code,
+                                      monitorObj.modelsUrl))
+  conn = boto.dynamodb2.connect_to_region(
+    monitorObj.region,
+    aws_access_key_id=monitorObj.awsAccessKeyId,
+    aws_secret_access_key=monitorObj.awsSecretAccessKey)
+
+  table = Table(monitorObj.table, connection=conn)
```
vitaly-krugl: Based on variable names used, one has to guess which specific table is being accessed.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+    resultSet = table.query_2(uid__eq=model["uid"], timestamp__gte=then[:19])
+
+    # Track all time intervals between valid (e.g. non-zero) samples
+    intervals = []
+
+    lastSampleTimestamp = None
+    for sample in resultSet:
+
+      if not sample["metric_value"]:
+        continue # Skip
```
vitaly-krugl: The `# Skip` comment is redundant

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+    self.apiKey = self.config.get("S1", "MODELS_MONITOR_TAURUS_API_KEY")
+    self.modelsUrl = self.config.get("S1", "MODELS_MONITOR_TAURUS_MODELS_URL")
+    self.region = self.config.get(
+      "S1", "MODELS_MONITOR_TAURUS_DYNAMODB_REGION")
+    self.awsAccessKeyId = self.config.get(
+      "S1", "MODELS_MONITOR_TAURUS_DYNAMODB_AWS_ACCESS_KEY_ID")
+    self.awsSecretAccessKey = self.config.get(
+      "S1", "MODELS_MONITOR_TAURUS_DYNAMODB_AWS_SECRET_ACCESS_KEY")
+    self.threshold = options.threshold
+    self.sigma = options.sigma
```
vitaly-krugl: => `self.sigmaMultiplier` for readability?

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,227 @@
...
+    if currentInterval > acceptableThreshold:
+      errors.append(_ErrorParams(model["name"],
+                                 model["uid"],
+                                 acceptableThreshold))
+
+  if errors:
+    msg = ("The following models have exceeded the acceptable threshold for "
+           "time since last timestamp in {} DynamoDB table:\n    "
+           .format(table.table_name)) + "\n    ".join(str(error)
+                                                      for error in errors)
```
shantanoo: Avoid using `string concat` using `+`, especially when `format` is already being used.

e.g.

```
msg = ("The following models have exceeded the acceptable threshold for "
       "time since last timestamp in {} DynamoDB table:\n    {}"
       .format(table.table_name, "\n    ".join(str(error))
```
                                                      for error in errors)

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+      intervals.append((timestamp - lastSampleTimestamp).total_seconds())
+      lastSampleTimestamp = timestamp
+
+    # Calculate current UTC timestamp adjusted to account for acceptable
+    # 10-minute delay in processing.
+    utcnow = datetime.datetime.utcnow() - datetime.timedelta(minutes=10)
+
+    stddev = numpy.std(intervals)
+    if stddev == float("nan") or lastSampleTimestamp is None:
+      errors.append(_ErrorParams(model["name"], model["uid"], None))
```
vitaly-krugl: Should there be a human-readable message in the error record that describes what the problem is?

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+
+      intervals.append((timestamp - lastSampleTimestamp).total_seconds())
+      lastSampleTimestamp = timestamp
+
+    # Calculate current UTC timestamp adjusted to account for acceptable
+    # 10-minute delay in processing.
+    utcnow = datetime.datetime.utcnow() - datetime.timedelta(minutes=10)
+
+    stddev = numpy.std(intervals)
+    if stddev == float("nan") or lastSampleTimestamp is None:
```
vitaly-krugl: Under which circumstances would standard deviation be NaN?

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+        continue
+
+      intervals.append((timestamp - lastSampleTimestamp).total_seconds())
+      lastSampleTimestamp = timestamp
+
+    # Calculate current UTC timestamp adjusted to account for acceptable
+    # 10-minute delay in processing.
+    utcnow = datetime.datetime.utcnow() - datetime.timedelta(minutes=10)
+
+    stddev = numpy.std(intervals)
```
vitaly-krugl: Does numpy distinguish between population standard deviation and sample standard deviation?

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+                    default=MIN_THRESHOLD,
+                    type="int",
+                    dest="threshold",
+                    metavar="SECONDS",
+                    help="Default: {}".format(MIN_THRESHOLD))
+  parser.add_option("--sigma",
+                    default=SIGMA_MULTIPLIER,
+                    type="int",
+                    dest="sigma",
+                    help="Default: {}".format(SIGMA_MULTIPLIER))
```
vitaly-krugl: I think that this help text needs to say a few more words about the value than simply showing the default value. How should someone determine an appropriate sigma multiplier? 

Do we reasonable expect anyone to override the default value? If not, let's not even expose it.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+                 " (in seconds) AND such interval exceeds a multiple of the "
+                 "standard deviation for all intervals for that model during "
+                 "a fixed window of time (14 days)."))
+
+  parser.add_option("--threshold",
+                    default=MIN_THRESHOLD,
+                    type="int",
+                    dest="threshold",
+                    metavar="SECONDS",
+                    help="Default: {}".format(MIN_THRESHOLD))
```
vitaly-krugl: Need better help text

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+    self.emailParams = loadEmailParamsFromConfig(self.config)
+    self.apiKey = self.config.get("S1", "MODELS_MONITOR_TAURUS_API_KEY")
+    self.modelsUrl = self.config.get("S1", "MODELS_MONITOR_TAURUS_MODELS_URL")
+    self.region = self.config.get(
+      "S1", "MODELS_MONITOR_TAURUS_DYNAMODB_REGION")
+    self.awsAccessKeyId = self.config.get(
+      "S1", "MODELS_MONITOR_TAURUS_DYNAMODB_AWS_ACCESS_KEY_ID")
+    self.awsSecretAccessKey = self.config.get(
+      "S1", "MODELS_MONITOR_TAURUS_DYNAMODB_AWS_SECRET_ACCESS_KEY")
+    self.threshold = options.threshold
```
vitaly-krugl: Need a better name for readability

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+    utcnow = datetime.datetime.utcnow() - datetime.timedelta(minutes=10)
+
+    stddev = numpy.std(intervals)
+    if stddev == float("nan") or lastSampleTimestamp is None:
+      errors.append(_ErrorParams(model["name"], model["uid"], None))
+      continue
+
+    # Fabricate a hypothetical interval representing the amount of time since
+    # the most recent valid timestamp
+    currentInterval = (utcnow - lastSampleTimestamp).total_seconds()
```
vitaly-krugl: Could this yield a negative number? How would that impact the calculations?

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
vitaly-krugl: Have you run this against the staging or production environment? What are some example stddev values for "healthy" stock metrics?

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+
+    # Only consider intervals that are more than N sigma AND above an arbitrary
+    # minimum threshold.  More frequent companies will have a lower stddev and
+    # therefore required a higher, if artifical, threshold to avoid too many
+    # false positives
+    acceptableThreshold = max(monitorObj.threshold, monitorObj.sigma * stddev)
```
vitaly-krugl: In this statement, is is confusing to see "sigma" multiplied by stddev, which is also commonly known as sigma (σ)

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+                                 model["uid"],
+                                 acceptableThreshold))
+
+  if errors:
+    msg = ("The following models have exceeded the acceptable threshold for "
+           "time since last timestamp in {} DynamoDB table:\n    "
+           .format(table.table_name)) + "\n    ".join(str(error)
+                                                      for error in errors)
+    raise LatencyMonitorError(msg)
+
```
vitaly-krugl: Need 3 lines between global objects

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -0,0 +1,223 @@
...
+
+def main():
+  ModelLatencyChecker().checkAll()
+
```
vitaly-krugl: Need 3 lines between global objects

vitaly-krugl:
@oxtopus: I am done with the code review, and there is feedback

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
oxtopus: Good catch.  Yes, it can yield a negative number.  It won't affect the calculations negatively, however.  i.e. if `lastSampleTimestamp` is more recent than the adjusted `utcnow`, it won't result in an error notification.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
oxtopus: I have, and it's across the board and depends on the frequency of twitter updates.  Obviously, stock is higher because it regularly goes dark for several hours.  Here are some examples:

```
TWITTER.TWEET.HANDLE.SPG.VOLUME 7958.04244184
XIGNITE.TRI.VOLUME 13386.7870253
XIGNITE.PCP.VOLUME 13386.7870253
TWITTER.TWEET.HANDLE.KMB.VOLUME 6677.86373017
TWITTER.TWEET.HANDLE.CVX.VOLUME 313.300877667
```

oxtopus: re: 
> Does numpy distinguish between population standard deviation and sample standard deviation?

`numpy.std()` does not, afaik.  However, we're only querying a sub sample of the total population.  To that effect, we're using a sample standard deviation.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
oxtopus: The heuristic is described in the description for the command.  I've since renamed to `sigmaMultiplier` for consistency.  I don't think we need to repeat documentation.  Sigma is a common, well understood concept in statistics.

re:
> Do we reasonable expect anyone to override the default value? If not, let's not even expose it. 

We may decide that the default of 3 results in too many false positives.  In which case, we may want to experiment with different values.  Allowing a CLI override makes it so you can experiment without a code change.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
oxtopus: Use of `threshold` is covered in description.  As with other options, we want to experiment with different values since the default values I've chosen are based on anecdotal observations.

PR #382, opened 2015-10-07
Proposed unit tests for the retry decorator.

SaganBolliger:
Would like feedback on on proposed unit tests.
https://jira.numenta.com/browse/ENG-50
@scottpurdy?

### DIFF
--- a/nta.utils/tests/unit/error_handling_test.py
+++ b/nta.utils/tests/unit/error_handling_test.py
@@ -26,7 +26,7 @@ import os
 import Queue
 import unittest
 
-from mock import patch, call
+from mock import patch, call, Mock
 
 from nta.utils import error_handling
 from nta.utils.logging_support_raw import LoggingSupport
@@ -127,22 +127,14 @@ class ErrorHandlingUtilsTest(unittest.TestCase):
       timeoutSec=0, initialRetryDelaySec=0.2,
       maxRetryDelaySec=10)
 
-    fcnExecCounter = [0]
+    testFunction = Mock(side_effect=Exception("Test exception"),
+                        __name__="testFunction")
 
-    @_retry
-    def testFunction():
-      fcnExecCounter[0] += 1
-      raise Exception("Test exception")
-
-    osExitCodeQ = Queue.Queue()
-    with patch.object(os, "_exit", autospec=True,
-                      side_effect=osExitCodeQ.put):
-
-      with self.assertRaises(Exception):
-        testFunction()
+    with self.assertRaises(Exception):
+      _retry(testFunction)()
 
     self.assertEqual(mockSleep.mock_calls, [])
-    self.assertEqual(fcnExecCounter[0], 1)
+    testFunction.assert_called_once_with()
 
 
   @patch("time.sleep")
@@ -157,29 +149,23 @@ class ErrorHandlingUtilsTest(unittest.TestCase):
       timeoutSec=30, initialRetryDelaySec=2,
       maxRetryDelaySec=10)
 
-    class _CounterContainer(object):
-      fcnExecCounter = 0
+    testFunction = Mock(side_effect=Exception("Test exception"),
+                        __name__="testFunction")
 
-    @_retry
-    def testFunction():
-      _CounterContainer.fcnExecCounter += 1
-      raise Exception("Test exception")
-
-    with patch.object(os, "_exit", autospec=True):
-      with self.assertRaises(Exception):
-        testFunction()
+    with self.assertRaises(Exception):
+      _retry(testFunction)()
 
     self.assertEqual(mockSleep.mock_calls, [call(2), call(4), call(8),
                                             call(10), call(10)])
 
     # Currently fails due to ENG-78
-    self.assertEqual(_CounterContainer.fcnExecCounter, 6)
+    self.assertEqual(testFunction.call_count, 6)
 
 
   @patch("time.sleep")
   @patch("time.time")
   def testRetryRetryExceptionIncluded(self, mockTime, mockSleep):
-    """ Test that retry is triggered if raised exeception is in
+    """ Test that retry is triggered if raised exception is in
     retryExceptions """
 
     self.mockSleepTime(mockTime, mockSleep)
@@ -198,9 +184,8 @@ class ErrorHandlingUtilsTest(unittest.TestCase):
     def testFunction():
       raise TestChildException("Test exception")
 
-    with patch.object(os, "_exit", autospec=True):
-      with self.assertRaises(Exception):
-        testFunction()
+    with self.assertRaises(Exception):
+      testFunction()
 
     self.assertEqual(mockSleep.call_count, 1)
 
@@ -227,9 +212,8 @@ class ErrorHandlingUtilsTest(unittest.TestCase):
     def testFunction():
       raise TestExceptionB("Test exception")
 
-    with patch.object(os, "_exit", autospec=True):
-      with self.assertRaises(TestExceptionB):
-        testFunction()
+    with self.assertRaises(TestExceptionB):
+      testFunction()
 
     self.assertEqual(mockSleep.call_count, 0)
 
@@ -259,9 +243,8 @@ class ErrorHandlingUtilsTest(unittest.TestCase):
     def testFunctionTrue():
       raise TestChildException("Test exception")
 
-    with patch.object(os, "_exit", autospec=True):
-      with self.assertRaises(TestChildException):
-        testFunctionTrue()
+    with self.assertRaises(TestChildException):
+      testFunctionTrue()
 
     self.assertEqual(mockSleep.call_count, 1)
 
@@ -278,9 +261,8 @@ class ErrorHandlingUtilsTest(unittest.TestCase):
     def testFunctionFalse():
       raise TestChildException("Test exception")
 
-    with patch.object(os, "_exit", autospec=True):
-      with self.assertRaises(TestChildException):
-        testFunctionFalse()
+    with self.assertRaises(TestChildException):
+      testFunctionFalse()
 
     self.assertEqual(mockSleep.call_count, 0)
 
--- a/nta.utils/tests/unit/error_handling_test.py
+++ b/nta.utils/tests/unit/error_handling_test.py
@@ -25,8 +25,6 @@
 import os
 import Queue
 import unittest
-import logging
-import time
 
 from mock import patch, call
 
@@ -40,8 +38,8 @@ def setUpModule():
 
 class ErrorHandlingUtilsTest(unittest.TestCase):
   """ Unit tests for the error-handling utilities """
-  
-  
+
+
   def testAbortProgramOnAnyExceptionWithoutException(self):
 
     @error_handling.abortProgramOnAnyException(exitCode=2)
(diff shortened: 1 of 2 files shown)


nta.utils/tests/unit/error_handling_test.py
```diff
@@ -95,6 +95,37 @@ def doSomething(*_args, **_kwargs):
...
+    pass
+
+
+  def testRetryMaxRetryDelaySecOverrides(self):
+    """ Test that maxRetryDelaySec overrides default behavior for
+    2 * initialRetryDelaySec retry delay """
+    pass
+
+
+  def testRetryRetryExceptionIncludedExcluded(self):
```
scottpurdy: maybe break this into `testRetryRetryExceptionIncluded` and `testRetryRetryExceptionExcluded` for the two different cases

scottpurdy:
These look like good tests to implement!

scottpurdy:
@vitaly-krugl may have suggestions for other things to check.

SaganBolliger:
Here are the implemented unit tests. Tests won't pass until #399 is merged.

One concern: the code

```
    accumulatedTime = [0]  
    def testTime():
      return accumulatedTime[0]
    
    def testSleep(duration):
      accumulatedTime[0] += duration  
    
    mockTime.side_effect = testTime
    mockSleep.side_effect = testSleep
```

repeats in each unit test. It would be nice to factor this out, but not what the best way is. @scottpurdy?

nta.utils/tests/unit/error_handling_test.py
```diff
@@ -94,6 +101,215 @@ def doSomething(*_args, **_kwargs):
       exitCode = osExitCodeQ.get_nowait()
       self.assertEqual(exitCode, 2)
 
+  @patch('time.sleep')
```
scottpurdy: use double quotes per style guide:
https://github.com/numenta/nupic/wiki/Python-Style-Guide#quotes

nta.utils/tests/unit/error_handling_test.py
```diff
@@ -94,6 +101,215 @@ def doSomething(*_args, **_kwargs):
...
+    def testTime():
+      return accumulatedTime[0]
+    
+    def testSleep(duration):
+      accumulatedTime[0] += duration  
+    
+    mockTime.side_effect = testTime
+    mockSleep.side_effect = testSleep
+    
+    _retry = error_handling.retry(timeoutSec=0, initialRetryDelaySec=0.2,
```
scottpurdy: Fix indentation. See [PEP-8](https://www.python.org/dev/peps/pep-0008/#indentation) examples. Also, you should run pylint on code before you submit. There is a pylintrc in the root of the repo that should get picked up automatically.

nta.utils/tests/unit/error_handling_test.py
```diff
@@ -94,6 +101,215 @@ def doSomething(*_args, **_kwargs):
...
+
+  @patch('time.sleep')
+  @patch('time.time')
+  def testRetryWaitsInitialRetryDelaySec(self, mockTime, mockSleep):
+    """ Test that first retry delay is initialRetryDelaySec and subsequent 
+    retry delays are geometrically doubling up to maxRetryDelaySec  """
+    
+    accumulatedTime = [0]
+    
+    def testTime():
```
scottpurdy: This seems overly complicated. Why not have mockTime and mockSleep return exact values? You can assert afterwards that `testSleep` is called with the durations that you expect.

scottpurdy:
With regards to repeated code, you can always put it into a different method that each test method calls. Only the methods that start with "test" will be treated as tests. But the non-test method will have to return anything you want to use in the calling method.

nta.utils/tests/unit/error_handling_test.py
```diff
@@ -25,16 +25,23 @@
...
+import logging
+import time
 
-from mock import patch
+from mock import patch, call
 
 from nta.utils import error_handling
 
+from nta.utils.logging_support_raw import LoggingSupport
+
```
oxtopus: Need three blank lines here, and the blank line above `from nta.utils.logging_support_raw import LoggingSupport` may be removed, too.

nta.utils/tests/unit/error_handling_test.py
SaganBolliger: @scottpurdy Do you mean the hanging indent on L121? If so, this is fixed.

nta.utils/tests/unit/error_handling_test.py
SaganBolliger: @scottpurdy Same logic, but refactored as per our discussion.

nta.utils/tests/unit/error_handling_test.py
SaganBolliger: Fixed.

nta.utils/tests/unit/error_handling_test.py
```diff
@@ -94,6 +101,215 @@ def doSomething(*_args, **_kwargs):
       exitCode = osExitCodeQ.get_nowait()
       self.assertEqual(exitCode, 2)
 
+  @patch('time.sleep')
+  @patch('time.time')
+  def testRetryNoRetries(self, mockTime, mockSleep):
+    """ Test that when timeoutSec == 0, function is executed exactly once 
+    with no retries, and raises an exception on failure. """    
+    
+    accumulatedTime = [0]
```
oxtopus: Why the `[0]` convention here, and elsewhere?  It seems unnecessary.

SaganBolliger: @oxtopus It's necessary because in order for the closure to have the side effect of increasing accumulatedTime, it must close on an object that can be modified, in this case a list object. Closing on a variable and then changing the variable doesn't work. See the latest commit that refactors the mocking code and introduces a dummy class for this purpose instead of a list object for clarity.

nta.utils/tests/unit/error_handling_test.py
```diff
@@ -95,6 +101,189 @@ def doSomething(*_args, **_kwargs):
...
+    """ Test that when timeoutSec == 0, function is executed exactly once
+    with no retries, and raises an exception on failure. """
+
+    self.mockSleepTime(mockTime, mockSleep)
+
+    _retry = error_handling.retry(
+      timeoutSec=0, initialRetryDelaySec=0.2,
+      maxRetryDelaySec=10)
+
+    fcnExecCounter = [0]
```
SaganBolliger: @scottpurdy Do you want me to do the dummy class thing here too instead of using the list object? This is the only point in which I count function executions.

oxtopus: Mock objects compute statistics like that automatically.  You may not need to implement anything.

SaganBolliger: Hmm, but this isn't a mock object. This is used target function ```testFunction()
``` that is being decorated to count how many times it is executed.

Comment (scottpurdy) [inline nta.utils/tests/unit/error_handling_test.py]: If still needed then yes, I'd use a class attribute instead of a list variable.

Comment (SaganBolliger) [inline nta.utils/tests/unit/error_handling_test.py]: Ok fixed.

Comment (oxtopus) [inline nta.utils/tests/unit/error_handling_test.py]: `testFunction()` can be a mock object with `Exception("Test exception")` as a side effect:

```
testFunction = Mock(side_effect=Exception("Test exception"))
with self.assertRaises(Exception):
  _retry(testFunction)()

mockObj.assert_called_once_with()
```

Comment (oxtopus) [inline nta.utils/tests/unit/error_handling_test.py]: You could also do something like this:

```
@_retry
def testFunction():
  raise Exception("foo")

mockedTestFunction = Mock(side_effect=testFunction)
with self.assertRaises(Exception):
  mockedTestFunction()

mockedTestFunction.assert_called_once_with()
```

Comment (SaganBolliger) [inline nta.utils/tests/unit/error_handling_test.py]: @oxtopus Sticking with the first solution variant we worked out.

File: nta.utils/tests/unit/error_handling_test.py [hunk 8]
```diff
@@ -95,6 +101,171 @@ def doSomething(*_args, **_kwargs):
...
+      timeoutSec=0, initialRetryDelaySec=0.2,
+      maxRetryDelaySec=10)
+
+    testFunction = Mock(side_effect=Exception("Test exception"),
+                        __name__="testFunction")
+
+    with self.assertRaises(Exception):
+      _retry(testFunction)()
+
+    self.assertEqual(mockSleep.mock_calls, [])
```
oxtopus: `self.assertFalse(mockSleep.called)` or, you can set a side effect to `mockSleep` that raises an exception if called, forcing the function to fail if it's called.

SaganBolliger: Changed in latest commit.

nta.utils/tests/unit/error_handling_test.py
```diff
@@ -95,6 +101,171 @@ def doSomething(*_args, **_kwargs):
       self.assertEqual(exitCode, 2)
 
 
+  def mockSleepTime(self, mockTime, mockSleep):
```
vitaly-krugl: Need a docstring that describes what it does and its args and their types

nta.utils/tests/unit/error_handling_test.py
```diff
@@ -95,6 +101,171 @@ def doSomething(*_args, **_kwargs):
...
+      return _TimeContainer.accumulatedTime
+
+    def testSleep(duration):
+      _TimeContainer.accumulatedTime += duration
+
+    mockTime.side_effect = testTime
+    mockSleep.side_effect = testSleep
+
+
+  @patch("time.sleep")
```
vitaly-krugl: MUST use `autospec=True` arg (preferred) in all patches and Mocks. autospec should work in all cases. In some cases, autospec doesn't work, and we fall back to using `spec_set` instead.

@SaganBolliger: read https://jira.numenta.com/browse/ENG-3 for an explanation.

vitaly-krugl:
@SaganBolliger: I noted some issues, but I am not done with the review yet, so please hold off on merging it.

nta.utils/tests/unit/error_handling_test.py
```diff
@@ -95,6 +101,171 @@ def doSomething(*_args, **_kwargs):
...
+    """ Test that when timeoutSec == 0, function is executed exactly once
+    with no retries, and raises an exception on failure. """
+
+    self.mockSleepTime(mockTime, mockSleep)
+
+    _retry = error_handling.retry(
+      timeoutSec=0, initialRetryDelaySec=0.2,
+      maxRetryDelaySec=10)
+
+    testFunction = Mock(side_effect=Exception("Test exception"),
```
vitaly-krugl: It's a bad idea to use `Exception` class, because if something else goes wrong (bug in retry, e.g., `KeyError`), that exception will also be based on `Exception`.

Instead, define and use your own exception class in this method: `class MyExcpetion(Exception)`

nta.utils/tests/unit/error_handling_test.py
SaganBolliger: Done

nta.utils/tests/unit/error_handling_test.py
SaganBolliger: Done

nta.utils/tests/unit/error_handling_test.py
```diff
@@ -95,6 +108,166 @@ def doSomething(*_args, **_kwargs):
...
+
+    self.assertFalse(mockSleep.called)
+    testFunction.assert_called_once_with()
+
+
+  @patch("time.sleep", autospec=True)
+  @patch("time.time", autospec=True)
+  def testRetryWaitsInitialRetryDelaySec(self, mockTime, mockSleep):
+    """ Test that first retry delay is initialRetryDelaySec and subsequent
+    retry delays are geometrically doubling up to maxRetryDelaySec  """
```
vitaly-krugl: For a multiline docstring, I think that it's our convention to place the trailing `"""` on a line all by itself. @scottpurdy, did I get this right?

nta.utils/tests/unit/error_handling_test.py
```diff
@@ -95,6 +108,166 @@ def doSomething(*_args, **_kwargs):
...
+    def testSleep(duration):
+      _TimeContainer.accumulatedTime += duration
+
+    mockTime.side_effect = testTime
+    mockSleep.side_effect = testSleep
+
+
+  @patch("time.sleep", autospec=True)
+  @patch("time.time", autospec=True)
+  def testRetryNoRetries(self, mockTime, mockSleep):
```
vitaly-krugl: I would rename this to something like `testRetryNoTimeForRetries` to make it clear why there are no retries. Otherwise, success could be another reason for no retries.

nta.utils/tests/unit/error_handling_test.py
scottpurdy: Yes. You can see examples in [PEP-257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) but I don't usually get too picky. I'd write this doc string like this:

```Test that delay times are correct.```

The details of what the delays should be are already documented in the function under test and should be obvious from the test code. Similarly, the parameters are explained by the decorators. So I don't think you need any explanation beyond the simple description of what is intended to be tested by this test method.

vitaly-krugl:
Need additional tests:

1. Decorated function receives the intended args (and no others) and that it's return value is not inhibited by the decorator.
2. initial call succeeds; make sure it didn't get retried; make sure the expected result was returned.
3. initial call(s) fail, but a subsequent call succeeds; make sure there was the expected number of retries; make sure the expected result was returned.

vitaly-krugl:
Given that there is going to be a fair number of tests for the retry decorator, lets place its tests in a separate test class; e.g., `RetryDecoratorTest`

vitaly-krugl:
@SaganBolliger, I am done with the review. There is additional feedback.

SaganBolliger:
@vitaly-krugl I believe I've implemented all the suggested changes.

Didn't check return value in new unit test 2 testNoRetryIfCallSucceeds since that is already tested in new unit test 1 testReturnsExpectedWithExpectedArgs.

vitaly-krugl:
LGTM

PR #395, opened 2015-10-08
Wait for services to stop

oxtopus:
This adds an extra step to check that there are no "STOPPING" or "RUNNING" supervisor services before finally shutting down.

### DIFF
--- a/taurus/pipeline/scripts/update-remote-taurus-servers.sh
+++ b/taurus/pipeline/scripts/update-remote-taurus-servers.sh
@@ -225,12 +225,24 @@ pushd "${REPOPATH}"
   ssh -v -t ${SSH_ARGS} "${TAURUS_SERVER_USER}"@"${TAURUS_SERVER_HOST}" \
     "cd /opt/numenta/products/taurus &&
      if [ -f taurus-supervisord.pid ]; then
+       supervisorctl --serverurl http://localhost:9001 stop all
+       let attempts=1
+       supervisorctl status | grep -E \"RUNNING|STOPPING\"
+       while [ $? -eq 0  ] && [ $attempts -le 5 ]; do
+         sleep 1
+         supervisorctl status | grep -E \"RUNNING|STOPPING\";
+         let \"attempts+=1\"
+       done
+       supervisorctl status | grep -E \"RUNNING|STOPPING\"
+       if [ $? -eq 0 ]; then
+         echo \"Unable to stop all processes...\"
+         exit;
+       fi
        supervisorctl --serverurl http://localhost:9001 shutdown
      fi  &&
      if [ -f /var/run/nginx.pid ]; then
        sudo /usr/sbin/nginx -p . -c conf/nginx-taurus.conf -s stop;
      fi"
-
 
   # Sync git histories with taurus server for current HEAD
   ${GIT} push --force \
--- a/taurus/pipeline/scripts/update-remote-taurus-servers.sh
+++ b/taurus/pipeline/scripts/update-remote-taurus-servers.sh
@@ -227,13 +227,13 @@ pushd "${REPOPATH}"
      if [ -f taurus-supervisord.pid ]; then
        supervisorctl --serverurl http://localhost:9001 stop all
        let attempts=1
-       supervisorctl status | grep -E \"RUNNING|STOPPING\"
+       supervisorctl --serverurl http://localhost:9001 status | grep -E \"RUNNING|STOPPING\"
        while [ \$? -eq 0  ] && [ \$attempts -le 5 ]; do
          sleep 1
-         supervisorctl status | grep -E \"RUNNING|STOPPING\";
+         supervisorctl --serverurl http://localhost:9001 status | grep -E \"RUNNING|STOPPING\";
          let \"attempts+=1\"
        done
-       supervisorctl status | grep -E \"RUNNING|STOPPING\"
+       supervisorctl --serverurl http://localhost:9001 status | grep -E \"RUNNING|STOPPING\"
        if [ \$? -eq 0 ]; then
          echo \"Unable to stop all processes...\"
          exit;
--- a/taurus/pipeline/scripts/update-remote-taurus-servers.sh
+++ b/taurus/pipeline/scripts/update-remote-taurus-servers.sh
@@ -228,13 +228,13 @@ pushd "${REPOPATH}"
        supervisorctl --serverurl http://localhost:9001 stop all
        let attempts=1
        supervisorctl status | grep -E \"RUNNING|STOPPING\"
-       while [ $? -eq 0  ] && [ $attempts -le 5 ]; do
+       while [ \$? -eq 0  ] && [ \$attempts -le 5 ]; do
          sleep 1
          supervisorctl status | grep -E \"RUNNING|STOPPING\";
          let \"attempts+=1\"
        done
        supervisorctl status | grep -E \"RUNNING|STOPPING\"
-       if [ $? -eq 0 ]; then
+       if [ \$? -eq 0 ]; then
          echo \"Unable to stop all processes...\"
          exit;
        fi


taurus/pipeline/scripts/update-remote-taurus-servers.sh
```diff
@@ -225,13 +225,25 @@ pushd "${REPOPATH}"
   ssh -v -t ${SSH_ARGS} "${TAURUS_SERVER_USER}"@"${TAURUS_SERVER_HOST}" \
     "cd /opt/numenta/products/taurus &&
      if [ -f taurus-supervisord.pid ]; then
+       supervisorctl --serverurl http://localhost:9001 stop all
```
vitaly-krugl: Please follow the recommendation at https://github.com/numenta/numenta-apps/pull/264/files#diff-7f881ff7a3ad0fa954565b8f6efa2d05R378 regarding
> These files were meant to mirror the instructions to the user in the readme as much as possible. I don't think a user would ever implement such a loop in their console. ...

That feedback makes a lot of sense.

taurus/pipeline/scripts/update-remote-taurus-servers.sh
```diff
@@ -225,13 +225,25 @@ pushd "${REPOPATH}"
   ssh -v -t ${SSH_ARGS} "${TAURUS_SERVER_USER}"@"${TAURUS_SERVER_HOST}" \
     "cd /opt/numenta/products/taurus &&
      if [ -f taurus-supervisord.pid ]; then
```
vitaly-krugl: If supervisor died unexpectedly, the pid file would still be around, but supervisord is inaccessible.

taurus/pipeline/scripts/update-remote-taurus-servers.sh
```diff
@@ -225,13 +225,25 @@ pushd "${REPOPATH}"
...
+       supervisorctl --serverurl http://localhost:9001 stop all
+       let attempts=1
+       supervisorctl status | grep -E \"RUNNING|STOPPING\"
```
vitaly-krugl: We pass `--serverurl http://localhost:9001` arg explicitly in other calls to supervisorctl in this and related scripts

taurus/pipeline/scripts/update-remote-taurus-servers.sh
```diff
@@ -225,13 +225,25 @@ pushd "${REPOPATH}"
...
+       while [ \$? -eq 0  ] && [ \$attempts -le 5 ]; do
+         sleep 1
+         supervisorctl status | grep -E \"RUNNING|STOPPING\";
```
vitaly-krugl: We pass `--serverurl http://localhost:9001` arg explicitly in other calls to supervisorctl in this and related scripts

taurus/pipeline/scripts/update-remote-taurus-servers.sh
```diff
@@ -225,13 +225,25 @@ pushd "${REPOPATH}"
...
+         let \"attempts+=1\"
+       done
+       supervisorctl status | grep -E \"RUNNING|STOPPING\"
```
vitaly-krugl: We pass `--serverurl http://localhost:9001` arg explicitly in other calls to supervisorctl in this and related scripts

vitaly-krugl:
The refresh-remote-taurus-servers.sh script also needs this logic as referenced in https://jira.numenta.com/browse/ENG-76

vitaly-krugl:
Done with code review, please see my feedback

taurus/pipeline/scripts/update-remote-taurus-servers.sh
oxtopus: Understood.  However,  something has to be done _before_ the code is updated remotely -- this code is checked out locally in the build environment before issuing the remote calls on the server.  I have a separate change that I held back on that more or less implements a more robust and documented version in the form of a nta.utils console script.  Once that's deployed, then we can back off on this change.

In the meantime, I don't see a simpler way to implement such a check while preventing an error in the remote instance, which is what we're trying to avoid in the first place.

oxtopus:
I disagree re: refresh.  There's no intention of gracefully preserving any models between updates in the refresh pipeline.  This change is needed to address only failures encountered in the graceful updates.

taurus/pipeline/scripts/update-remote-taurus-servers.sh
oxtopus: In the interim, if we run into a situation like that (unlikely), it will require manual remediation per design.  No need to address here and now.

oxtopus:
Merging this for now.  Once this is deployed, I'll submit https://github.com/numenta/numenta-apps/compare/master...oxtopus:ENG-76-wait-for-services-console-script for review in which a console script is added that obviates the issue of:

> These files were meant to mirror the instructions to the user in the readme as much as possible. I don't think a user would ever implement such a loop in their console. ...

taurus/pipeline/scripts/update-remote-taurus-servers.sh
vitaly-krugl: I would hate to think that manual remediation is the preferred approach.

oxtopus: It's part of the design.

oxtopus: @vitaly-krugl Following up to:

> I would hate to think that manual remediation is the preferred approach

Please see https://github.com/numenta/numenta-apps/tree/master/taurus/pipeline/scripts#compute-resources-are-ephemeral in which the README states:

>  Little attempt is made to recover an instance from a failed state, either from the result of a failure in the update process, tests, or intermittent failures. If something goes wrong, a failed pair of instances should be replaced with a fresh pair of instances, or the failed instance remediated manually.

Since it's clear you disagree, please propose a change to the wording.

vitaly-krugl: @oxtopus, all that I was suggesting was a little defensive coding to avoid ending up with a failed server or failed continuous integration and other pipelines. As things stand now, we have wasted plenty of time and effort investigating and recovering from failed pipelines, such as the recent incidents with failed models as well as the one where 3 out of 5 attempts to update staging and production taurus servers resulted in failure, all due to lack of a little defensive coding.

We don't need a convention or guidelines for implementing robust code; it should be self-evident.

CC @scottpurdy

vitaly-krugl: @scottpurdy, as far as I can tell, there may be several problems here.

1. In the following snippet, `let \"attempts+=1\"` overwrites the value of `$?` from `supervisorctl --serverurl http://localhost:9001 status | grep -E \"RUNNING|STOPPING\"`, so the control loop is checking the wrong return status.

 ```
       while [ \$? -eq 0  ] && [ \$attempts -le 5 ]; do
         sleep 1
         supervisorctl --serverurl http://localhost:9001 status | grep -E \"RUNNING|STOPPING\";
         let \"attempts+=1\"
       done
```

2. After calling `supervisorctl --serverurl http://localhost:9001 shutdown`, need to wait for supervisor to shut down, to avoid race condition whereby you get the error "port xyz is in use" when trying to restart supervisor in the subsequent steps.

 ```
supervisorctl --serverurl http://localhost:9001 shutdown
```

3. Checking taurus-supervisord.pid  and stopping services independently of shutting down supervisor may be unnecessary. It would probably suffice to simply issue the shtudown call to supervisor and then wait for supervisor to become unresponsive, using supervisor RPC interface.

scottpurdy: > It would probably suffice to simply issue the shtudown call to supervisor and then wait for supervisor to become unresponsive, using supervisor RPC interface.

@oxtopus - Does that seem reasonable?

vitaly-krugl: In case the shutdown wait logic is implemented in bash, using the current `stop all` wait loop as a template, we want to make sure that the issues identified above are not re-introduced in the new loop.

PR #398, opened 2015-10-09
Eng 76 install console script but dont use it yet

oxtopus:
Merge and deploy this before #396!

### DIFF
--- a/nta.utils/nta/utils/supervisor_utils.py
+++ b/nta.utils/nta/utils/supervisor_utils.py
@@ -1,0 +1,48 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have purchased from
+# Numenta, Inc. a separate commercial license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+from urlparse import urljoin
+import xmlrpclib
+
+
+
+class SupervisorClient(xmlrpclib.Server):
+  def __init__(self, serverUrl):
+    """
+    Supervisord xmlrpclib.Server wrapper
+
+    :param str serverUrl: Supervisor API (e.g. http://127.0.0.1:9001)
+
+    Usage::
+
+        from nta.utils.supervisor_utils import SupervisorClient
+
+        client = SupervisorClient("http://127.0.0.1:9001")
+
+        server.supervisor.getState()
+
+        if state.get("statename") != "RUNNING":
+          raise Exception("Supervisord is not running")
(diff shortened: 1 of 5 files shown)


nta.utils/nta/utils/supervisor_utils.py
```diff
@@ -0,0 +1,48 @@
+# ----------------------------------------------------------------------
```
vitaly-krugl: @oxtopus: it seems like it might be better to fold supervisord_state into supervisor_utils, instead of having the two files.

nta.utils/nta/utils/tools/supervisord_state.py
```diff
@@ -112,3 +111,56 @@ def waitForRunningStateMain():
   else:
     sys.exit("Timed out waiting for supervisord RUNNING state from %s" % (
       args.supervisorApiUrl,))
+
+
+
+def waitForAllToStop(timeout=60,
```
vitaly-krugl: @oxtopus: is there some way that a console script stub can pass any of these parameters to this function? If not, then it's confusing to have these parameters here.

nta.utils/nta/utils/tools/supervisord_state.py
```diff
@@ -112,3 +111,56 @@ def waitForRunningStateMain():
...
+               .format(args.supervisorApiUrl, repr(ex)))
+
+    return []
+
+  processes = getAllProcessInfo()
+
+  now = time.time()
+
+  while not all(p.get("statename") in stoppedStates for p in processes):
+    if time.time() > (now + args.timeout):
```
vitaly-krugl: @oxtopus: I found a bug here: if timeout is passed on the command line, then it will be of type `str`, and the addtion of args.timeout will throw an exception.

nta.utils/nta/utils/tools/supervisord_state.py
```diff
@@ -112,3 +111,56 @@ def waitForRunningStateMain():
...
+               "\n    ".join("{} (pid {}) {} (status {})"
+                             .format(p.get("name"),
+                                     p.get("pid"),
+                                     p.get("statename"),
+                                     p.get("state"))
+                             for p in processes))
+
+    print >> sys.stderr, ("Sleeping for {} second{}..."
+                          .format(args.sleep, "s" if args.sleep != 1 else ""))
+    time.sleep(sleep)
```
vitaly-krugl: This is a bug, too. It should be getting the value from `args.sleep`. Also, make sure that arg parser converts the value to float, otherwise it will be string.

nta.utils/nta/utils/tools/supervisord_state.py
```diff
@@ -112,3 +111,56 @@ def waitForRunningStateMain():
   else:
     sys.exit("Timed out waiting for supervisord RUNNING state from %s" % (
       args.supervisorApiUrl,))
+
+
+
+def waitForAllToStop(timeout=60,
+                     sleep=1,
+                     stoppedStates=set(["STOPPED", "FATAL", "ERROR"])):
```
vitaly-krugl: pylint usually doesn't like it when a mutable object is used as the default value of an arg.

vitaly-krugl:
@oxtopus: I am done with the code review; there are a few issues

vitaly-krugl:
I feel bad that we don't have tests for these. I didn't create tests for waitForRunningStateMain either.

nta.utils/nta/utils/tools/supervisord_state.py
oxtopus: The function signature is a perfectly acceptable place to define defaults like these.

nta.utils/nta/utils/tools/supervisord_state.py
oxtopus: Indeed, pylint raised a warning.  Fixed in https://github.com/numenta/numenta-apps/pull/407 by using `frozenset`.

PR #402, opened 2015-10-09
Make sure that /usr/local/git/bin is in PATH for cronjobs

unixorn:
We stopped installing upstream git, so ensure we have our git in the PATH

@jcasner @shantanoo, please CR


jcasner:
:+1:

PR #403, opened 2015-10-09
ENG-74 Incorporate a timestamp and increase traceback "tail" size in … … …htmengine's failed model and inference command results that are dispatched to anomaly service.

vitaly-krugl:
ENG-74 Incorporate a timestamp and increase traceback "tail" size in htmengine's failed model and inference command results that are dispatched to anomaly service. This should help with diagnosis.


vitaly-krugl:
CC @oxtopus

PR #404, opened 2015-10-12
Install using revised python prefix as opposed to split package install location and script dir

oxtopus:
It's just simpler to have to specify a single prefix rather than two paths.  In practice, the two paths always share a common prefix anyway.

Requires `install-prefix` companion branch in internal infrastructure repository.

### DIFF
--- a/htm.it/README.rst
+++ b/htm.it/README.rst
@@ -26,11 +26,9 @@ automatically with htm-it.
 
 ::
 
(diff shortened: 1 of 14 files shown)


jcasner:
Missing a couple places...

* infrastructure/ami-tools/packer-scripts/configure-htm-it-pipeline-centos6.5-ami
* infrastructure/ami-tools/packer-scripts/configure-htm-it-pipeline-centos7-ami

jcasner:
:+1:

unixorn:
:+1:

PR #407, opened 2015-10-13
ENG-76 Fixup type-related bugs discovered in review

oxtopus:
Thanks, @vitaly-krugl !

### DIFF
--- a/nta.utils/nta/utils/tools/supervisord_state.py
+++ b/nta.utils/nta/utils/tools/supervisord_state.py
@@ -116,7 +116,7 @@ def waitForRunningStateMain():
 
 def waitForAllToStop(timeout=60,
                      sleep=1,
-                     stoppedStates=set(["STOPPED", "FATAL", "ERROR"])):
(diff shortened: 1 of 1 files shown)


nta.utils/nta/utils/tools/supervisord_state.py
```diff
@@ -116,7 +116,7 @@ def waitForRunningStateMain():
 
 def waitForAllToStop(timeout=60,
```
vitaly-krugl: @oxtopus: this function is a console script entry point; it should not have parameters; it depends on arg parser and may call sys.exit(), so its not compatible with use as a simple function. Please remove the parameters and specify the default values directly as parameters in `parser.add_argument()` calls instead or turn them into constants. It's misleading the way you have it right now.

oxtopus: _"Fixed"_ in https://github.com/numenta/numenta-apps/pull/409

vitaly-krugl: Thanks, @oxtopus

PR #408, opened 2015-10-13
UNI-200: Add DatabaseServer unit tests

lscheinkman:
@brev Please review.
Added `DatabaseServer.js` unit tests, switched to `leveldown` backend and fixed some bugs found by test.


unicorn/frontend/lib/DatabaseServer.js
```diff
@@ -165,19 +172,26 @@ DatabaseServer.prototype.getMetricDatas = function(query, callback) {
         results.push(result);
       }
 
-      // sort by rowid
+      // sort by uid, metric_uid, rowid
       results.sort((a, b) => {
```
brev: Would you mind switching to longer and more descriptive var names?  `prev` and `next` is what I've gone with thanks. (this is one of the new lint rules, 3 chars min I think).

brev:
:+1: very exciting, thanks. cc @marionleborgne

PR #411, opened 2015-10-14
TAUR-964 Updated metrics config, automated cleanup of deprecated companies/models and promotion of metrics to models

vitaly-krugl:
TAUR-964 Updated taurus metrics.csv and re-generated metrics.json.
TAUR-964 Implemented delete_company.py tool.

TAUR-964 Implemented metric_utils.deleteMetricsFromEngineBySymbol().
TAUR-964 Implemented Unit and Integration tests for deleteMetricsFromEngineBySymbol.
TAUR-964 metric_utils.deleteMetric() now raises MetricNotFound on HTTP error 404.
TAUR-964 Implemented unit tests for metric_utils.deleteMetric().

TAUR=964 Added Console script entry point for taurus-delete-company.

TAUR-964 Cleaned up some legacy pylint warnings.

ENG-43 Fix race condition in testCreateAllModels metric collectors integration test.

TAUR-964 In delete_company.py, log whether xignite_security row was deleted.

TAUR-964 Removed code related to the old TAUR-1327 issue from resymbol_metrics after updating TAUR-1327 description such that the code becomes unnecessary.

TAUR-964 Integrated garbage collection of old tweets into twitter_direct_agent metric collector.

TAUR-964 Implemented company_maintenance_agent.py, a service responsible for cleanup of deprecated companies/models and for promoting unmonitored metrics to models.

TAUR-964 Misc. pylint and other cleanup.

TAUR-964 Impelemented unit tests for company_maintenance_agent.
TAUR-964 Implemented new metric_utils functions and tests: getMetricNamesFromConfig, filterCompanyMetricNamesBySymbol.
TAUR-964 Added support for metric name filtering to metric_utils.createAllModels and tests.

TAUR-964 Implemented unit and integration tests for metric_utils.ComapanyDeleter.
TAUR-964 Implemented nta.utils.metric_test_utils.configureTimeAndSleepMocks() (borrowed from Sagan's work in retry decorator test)

AUR-964 Implemented unit test for nta.utils.test_utils.time_test_utils


vitaly-krugl:
Please review

CC @scottpurdy @oxtopus

nta.utils/nta/utils/test_utils/time_test_utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+"""Test utilities concerning the builtin time module"""
+
+
+import threading
+threading.Lock()
+
+
+
+def configureTimeAndSleepMocks(timeMock, sleepMock):
```
vitaly-krugl: @SaganBolliger: I borrowed this from your work on the retry decorator test, so that it can be used in other tests, too. I updated your tests in this PR to use the shared version as well.

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
```diff
@@ -0,0 +1,235 @@
...
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""This service maintains consistency between the collectors' metrics
+configuration and the models as well as the buffered data in metric collectors
+repository. It also promotes unmonitored metrics that have reached a certain
+metric data quantity threshold to models.
+
+Upon start-up: The script checks its mode and exists with 0 return code if the
```
oxtopus: `exists` -> `exits`

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
```diff
@@ -0,0 +1,235 @@
...
+
+  :param list args: the equivalent of sys.argv[1:]
+
+  """
+
+  helpString = (
+    "%prog\n\n"
+    "{doc}".format(doc=__main__.__doc__))
+
+  parser = OptionParser(helpString)
```
oxtopus: Obligatory "_new_ code that parses command line args should use `argparse`"

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
```diff
@@ -0,0 +1,235 @@
...
+
+  _, remainingArgs = parser.parse_args(args=args)
+  if remainingArgs:
+    parser.error("Unexpected positional args: %r" % (remainingArgs,))
+
+
+
+@collectorsdb.retryOnTransientErrors
+def _queryCachedCompanySymbols():
+  """# Get the cached security symbols from the xignite_security table
```
oxtopus: Remove the `#`

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
```diff
@@ -0,0 +1,235 @@
...
+
+  :returns: A set of stock symbols from the xignite_security table
+  :rtype: set
+  """
+  engine = collectorsdb.engineFactory()
+
+  return set(
+    row.symbol for row in
+    engine.execute(sql.select([schema.xigniteSecurity.c.symbol])).fetchall())
```
oxtopus: Should be able to use `distinct()` to construct a query that dedupes.

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -335,19 +440,298 @@ def deleteMetric(host, apiKey, metricName):
...
+  # XIGNTE.NEWS.AMZN.VOLUME
+  return tuple(
+    name for name in metricNames
+    if all(name.split(".")) and
+    len(name.split(".")) >= 3 and
+    name.split(".")[-2] == tickerSymbol)
+
+
+
+class CompanyDeleter(object):
```
oxtopus: My first reaction when I see a class name that ends with "er" is to assume that it requires an _instance_.  It's counterintuitive (to me) that this class only exposes class methods because of the name.  Meanwhile, since there's no mutable state and I don't believe that subclassing `CompanyDeleter` is an expected pattern, `CompanyDeleter` is just being used as a container for utility code.  IMO, this is better suited as a module, without invoking any OOP.

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
```diff
@@ -0,0 +1,235 @@
...
+      tickerSymbols=deprecatedSymbols,
+      engineServer=_TAURUS_HTM_SERVER,
+      engineApiKey=_TAURUS_API_KEY,
+      warnAboutDestructiveAction=False)
+  else:
+    g_log.info("There were no deprecated companies to remove")
+
+
+
+def _determineMetricsReadyForPromotion(metricsConfig, allCustomMetrics):
```
oxtopus: `determine` in the name is superfluous.  It distracts from the intent of the function, which _returns_ an iterable of metric names that are ready for promotion.

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
```diff
@@ -0,0 +1,235 @@
...
+
+  :param dict metricsConfig: Metric configuration object that defines all
+    instances and metrics for all data collectors; as returned by
+    `metric_utils.getMetricsConfiguration()`
+
+  :param iterable allCustomMetrics: Custom metric info dicts from
+    Taurus Engine as returned by `metric_utils.getAllCustomMetrics()`
+
+  :returns: Names of of metrics that need to be promoted to models
+  :rtype: iterable
```
oxtopus: It specifically returns a `tuple()`.

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
```diff
@@ -0,0 +1,235 @@
...
+  """
+  configuredMetricNames = set(
+    metric_utils.getMetricNamesFromConfig(metricsConfig))
+
+  # Compile a set of unmonitored metrics that are in our current metrics
+  # configuration and are ready for promoting to models
+  return tuple(
```
oxtopus: re: earlier comment about return type.  With a documented return type of `iterable`, I would expect the return to be:

```
return (
  obj["name"] for obj in allCustomMetrics
  if obj["status"] == 0 and obj["name"] in configuredMetricNames and
    obj["last_rowid"] >= _NUM_METRIC_DATA_ROWS_THRESHOLD
)
```

taurus.metric_collectors/taurus/metric_collectors/delete_company.py
```diff
@@ -0,0 +1,143 @@
...
+    "--engine-apikey",
+    required=True,
+    dest="engineApiKey",
+    metavar="TAURUS_ENGINE_API_KEY",
+    help="API Key of Taurus HTM Engine")
+
+  args = parser.parse_args()
+
+  if not args.tickerSymbols:
+    msg = ("Missing or empty company stock ticker symbols")
```
oxtopus: No need for parens.

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -87,12 +97,50 @@ class RetriesExceededError(Exception):
...
+
+
+class WarningPromptTimeout(Exception):
+  """The warning prompt about the destructive action timed out"""
+  pass
+
+
+
+class FlusherMetricNotFound(Exception):
+  """Inidicates that the the wait for Taurus Engine metric data path flusher
```
oxtopus: `the the` -> `the`

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -87,12 +97,50 @@ class RetriesExceededError(Exception):
...
+
 g_log = logging.getLogger("metric_collectors.metric_utils")
 
 
 
 # Retry decorator for specific `requests` errors
-def _retry_on_requests_errors(timeoutSec=10, pauseSec=0.2):
+def _retryOnRequestsErrors(timeoutSec=10, pauseSec=0.2):
+
+  def retryFilter(exc, _args, _kwargs):
```
oxtopus: Should this not be `def retryFilter(exc, *_args, **_kwargs):`?

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -226,30 +299,63 @@ def createCustomHtmModel(host,
...
   """
-  metricsConfiguration = getMetricsConfiguration()
+  metricsConfig = getMetricsConfiguration()
+
+  configuredMetricNames = set(getMetricNamesFromConfig(metricsConfig))
+
+  if onlyMetricNames is not None:
+    # Validate onlyMetricNames and convert it to set
+
+    if not len(onlyMetricNames):
```
oxtopus: `if not onlyMetricNames:` would satisfy the condition, too.

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -226,30 +299,63 @@ def createCustomHtmModel(host,
...
+      raise ValueError("onlyMetricNames is empty")
+
+    asSet = set(onlyMetricNames)
+    if len(asSet) != len(onlyMetricNames):
+      raise ValueError("onlyMetricNames contains duplicates")
```
oxtopus: What is risk of duplicates?  Is the check and exception necessary?

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -335,19 +440,298 @@ def deleteMetric(host, apiKey, metricName):
...
+
+
+def filterCompanyMetricNamesBySymbol(metricNames, tickerSymbol):
+  """Filter company metric names by stock symbol
+
+  :param sequence metricNames: custom metric names
+  :param str tickerSymbol: Stock symbol; only company metric names matching
+    this stock symbol will be returned.
+
+  :returns: sequence of company metric names mathing `tickerSymbol`
```
oxtopus: `mathing` -> `matching`

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -335,19 +440,298 @@ def deleteMetric(host, apiKey, metricName):
...
+  :rtype: sequence
+  """
+  tickerSymbol = tickerSymbol.upper()
+
+  # Examples company metric names:
+  # TWITTER.TWEET.HANDLE.AET.VOLUME
+  # XIGNITE.TWC.CLOSINGPRICE
+  # XIGNITE.AMZN.VOLUME
+  # XIGNTE.NEWS.AMZN.VOLUME
```
oxtopus: `XIGNTE.NEWS.AMZN.VOLUME` -> `XIGNITE.NEWS.AMZN.VOLUME`

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -335,19 +440,298 @@ def deleteMetric(host, apiKey, metricName):
...
+  :param sequence metricNames: custom metric names
+  :param str tickerSymbol: Stock symbol; only company metric names matching
+    this stock symbol will be returned.
+
+  :returns: sequence of company metric names mathing `tickerSymbol`
+  :rtype: sequence
+  """
+  tickerSymbol = tickerSymbol.upper()
+
+  # Examples company metric names:
```
oxtopus: `Examples` -> `Example`

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -335,19 +440,298 @@ def deleteMetric(host, apiKey, metricName):
...
+      warnAboutDestructiveAction is False
+
+    :raises WarningPromptTimeout: if warning prompt timed out
+    :raises UserAbortedOperation: if user chose to abort the operation
+    :raises FlusherMetricNotFound:
+    """
+    tickerSymbols = tuple(symbol.upper() for symbol in tickerSymbols)
+
+    # Check for duplicate symbols
+    repeatedSymbols = set(sym for sym in tickerSymbols
```
oxtopus: This, and the next few lines could be expressed simpler:

```python
if any(tickerSymbols.count(sym) > 1 for sym in tickerSymbols):
  raise ValueError(...)
```

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -335,19 +440,298 @@ def deleteMetric(host, apiKey, metricName):
...
+
+    # Set will be handier going forward
+    tickerSymbols = set(tickerSymbols)
+
+    if warnAboutDestructiveAction:
+      cls._warnAboutDestructiveAction(timeout=warningTimeout,
+                                      tickerSymbols=tickerSymbols,
+                                      engineServer=engineServer)
+
+    # If any of the the ticker symbols still appear in the collector's metrics
```
oxtopus: `the the` -> `the`

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -335,19 +440,298 @@ def deleteMetric(host, apiKey, metricName):
...
+    # config, abort the operation as a precautionary measure.
+    allSymbols = set(security[0].upper() for security in
+                     getAllMetricSecurities())
+
+    problemSymbols = tickerSymbols.intersection(allSymbols)
```
oxtopus: Given use of set operators elsewhere, I would use `&` for consistency rather than `set.intersection()`

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -335,19 +440,298 @@ def deleteMetric(host, apiKey, metricName):
...
+      # symbols presently depends on aging of the rows (e.g., twitter tables).
+      with collectorsdb.engineFactory().begin() as conn:
+        numDeleted = (
+          conn.execute(
+            collectorsdb.schema.xigniteSecurity  # pylint: disable=E1120
+            .delete()
+            .where(collectorsdb.schema.xigniteSecurity.c.symbol == symbol))
+          ).rowcount
+
+        assert 0 <= numDeleted <= 1, (
```
oxtopus: `numDeleted` is always an integer, and allowable values are explicitly `0` or `1`.  Better to be more explicit in the assertion:  `assert numDeleted in (0, 1):`

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -335,19 +440,298 @@ def deleteMetric(host, apiKey, metricName):
...
+              expectedAnswer=expectedAnswer,
+              timeout=timeout))
+
+    timerExpiredQ = Queue.Queue()
+
+    def onTimerExpiration():
+      timerExpiredQ.put(1)
+      # NOTE: thread.interrupt_main() doesn't unblock raw_input, so we use
+      # SIGINT instead
+      os.kill(os.getpid(), signal.SIGINT)
```
oxtopus: Good call on the timed auto-abort!

taurus.metric_collectors/taurus/metric_collectors/twitterdirect/twitter_direct_agent.py
```diff
@@ -1348,6 +1350,29 @@ def _queryTweetVolumes(self, aggDatetime, metrics):
 
 
 
+class _TweetGarbageCollector(object):
+  """Garbage collector for old tweets in the twitter_tweets table"""
+
+  # Tweets oder than this many days will be deleted from twitter_tweets table
```
oxtopus: `oder` -> `older`

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
```diff
@@ -0,0 +1,235 @@
...
+
+def main():
+  """ NOTE: main may be used as "console script" entry point by setuptools
+  """
+  logging_support.LoggingSupport.initService()
+
+  try:
+
+    try:
+      _parseArgs(sys.argv[1:])
```
oxtopus: Consider renaming `_parseArgs()` to something like `_assertNoArgs`, or something more indicative of the intent.  `_parseArgs` is misleading if it's only being used to validate that the user didn't pass anything.

taurus.metric_collectors/taurus/metric_collectors/twitterdirect/twitter_direct_agent.py
vitaly-krugl: Well, the old tweets have this "oder" about them :P

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
```diff
@@ -0,0 +1,235 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero Public License for more details.
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""This service maintains consistency between the collectors' metrics
```
oxtopus: Perhaps include some details about _how_ to configure the opmode, since this doesn't take any user input.

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
vitaly-krugl: `_parseArgs` is still being used to parse the command-line, looking for options such as `--help` and `-h`, and provide help as their main function here at the moment.

taurus.metric_collectors/tests/unit/metric_utils_test.py
```diff
@@ -528,6 +720,341 @@ def testMetricDataBatchWrite(self, MessageBusConnectorMock):
...
+
+    # Execute and validate
+    with self.assertRaises(metric_utils.FlusherMetricNotFound):
+      metric_utils.CompanyDeleter._waitForFlusherAndGarbageCollect(
+        engineServer="host",
+        engineApiKey="apikey",
+        flusherMetricName=flusherMetricName)
+
+
+  @patch("__builtin__.raw_input", autospec=True)
```
scottpurdy: I haven't seen this before. Why do you need to have the `__builtin__.` prefix?

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def _purgeDeprecatedCompanies():
+  """Purge cached data and Taurus Engine metrics/models corresponding to symbols
+  that are in xignite_security table, but not in metrics configuration.
+  """
+  activeCompanySymbols = set(security[0] for security in
+                             metric_utils.getAllMetricSecurities())
+
+  cachedCompanySymbols = _queryCachedCompanySymbols()
```
oxtopus: I would prefer to see the `set()` conversion here, rather than in the return value of `_queryCachedCompanySymbols()` since there's an expectation for _how_ the return value will be used.

taurus.metric_collectors/tests/unit/metric_utils_test.py
vitaly-krugl: @scottpurdy: `__builtin__` is the magic module where builtins reside. It doesn't work by itself. Look for `Where Is open?` in http://www.ichimonji10.name/blog/6/. In python 3.xx this gets renamed to `builtins` and the same technique is used there - look for `builtins.open` in https://docs.python.org/3/library/unittest.mock-examples.html

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -335,19 +440,298 @@ def deleteMetric(host, apiKey, metricName):
...
+  _DATA_PATH_FLUSH_TIMEOUT_SEC = 300
+
+  # Prefix of metric name used for flushing Taurus Engine's metric data path;
+  # see code comments in `CompanyDeleter.deleteCompanies()` for more info.
+  _DATA_PATH_FLUSHER_METRIC_PREFIX = (  # pylint: disable=C0103
+    ".delete_metric_flusher_")
+
+
+  @classmethod
+  def deleteCompanies(cls,
```
oxtopus: There is a lot of knowledge about multiple disparate components encoded in this utility.  I think it would be better if the function were decomposed in such a way that the service-specific delete operations are implemented in the services themselves, and this function is only responsible for propagating the `deleteCompanies()` call to each of the services, which implement the service-specific delete logic.

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
```diff
@@ -335,19 +440,298 @@ def deleteMetric(host, apiKey, metricName):
...
+    if not found:
+      raise FlusherMetricNotFound("Still waiting for data path flusher metric "
+                                  "{metric}".format(metric=flusherMetricName))
+
+
+  @classmethod
+  def _flushTaurusEngineMetricDataPath(cls, engineServer, engineApiKey):
+    """Flush Taurus Engine's metric data path.
+
+    There is no formal mechanism for this in htmengine, so we're going to flush
```
oxtopus: Sending a dummy metric sample in this fashion, adding additional code and consuming resources seems a bit hacky, and instead the addressing the issue in `htmengine` seems like a good idea.

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
```diff
@@ -0,0 +1,235 @@
...
+  metricsConfig = metric_utils.getMetricsConfiguration()
+
+  readyMetricNames = _determineMetricsReadyForPromotion(
+    metricsConfig=metricsConfig,
+    allCustomMetrics=metric_utils.getAllCustomMetrics(
+      _TAURUS_HTM_SERVER,
+      _TAURUS_API_KEY))
+
+  if not readyMetricNames:
+    g_log.info("There are no metrics that are ready for promotion at this time")
```
oxtopus: Creating metrics are a relatively rare operation relative to the lifetime of the service.  As a result, log messages such as this will fill up the logs.  Seems better to log that metrics are created, as opposed to when not.

oxtopus:
Done w/ review for now.

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
vitaly-krugl: makes sense, done.

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
vitaly-krugl: they are already unique, so no need to dedupe; however, I made the recommended change in the user of this function.

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: Regarding
> CompanyDeleter is just being used as a container for utility code. IMO, this is better suited as a module, without invoking any OOP

Good call!. Done

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: @oxtopus, regarding
> There is a lot of knowledge about multiple disparate components encoded in this utility

After https://jira.numenta.com/browse/ENG-83 and https://jira.numenta.com/browse/ENG-82, this should become generic for across all collectors services, so I am going to leave the code as is (other then moving it out of metric_utils)

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
vitaly-krugl: > It specifically returns a tuple().

@oxtopus What if it's not important or desirable to commit to to a specific return type, what's wrong with documenting a generic pseudo-type instead, such as sequence?

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
vitaly-krugl: > re: earlier comment about return type. With a documented return type of iterable, I would expect ...

@oxtopus, your example

```
return (
  obj["name"] for obj in allCustomMetrics
  if obj["status"] == 0 and obj["name"] in configuredMetricNames and
    obj["last_rowid"] >= _NUM_METRIC_DATA_ROWS_THRESHOLD
)
```
is an iterator, but not an iterable. You can cycle through an iterator only once, but an iterable may be cycled multiple times.

taurus.metric_collectors/taurus/metric_collectors/delete_company.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: No, the callback gets a copy of the args and kwargs that were passed to the decorated function in a sequence and map that are the 2nd and 3rd args of the callback.

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
oxtopus: Calling it a `sequence` is better.

vitaly-krugl: --> sequence

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: @oxtopus: Duplicates here would indicate a bug in user code. I am going to remove this check and let the user code fend for itself via its own test code.

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: > This, and the next few lines could be expressed simpler ...
@oxtopus, the code below uses repeatedSymbols in the exception message.

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: good point

taurus.metric_collectors/taurus/metric_collectors/twitterdirect/twitter_direct_agent.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
vitaly-krugl: done

taurus.metric_collectors/taurus/metric_collectors/metric_utils.py
vitaly-krugl: Regarding
> Sending a dummy metric sample in this fashion, adding additional code and consuming resources seems a bit hacky, and instead the addressing the issue in htmengine seems like a good idea.

@oxtopus, I would be happy to brainstorm a possible solution together with you.

taurus.metric_collectors/taurus/metric_collectors/common_services/company_maintenance_agent.py
vitaly-krugl: `metric_utils.createAllModels` already logs when metrics are created. I agree with your feedback, so we're just going to remove the unwanted logging.

vitaly-krugl:
@oxtopus, thanks for the thorough feedback. I addressed it in https://github.com/vitaly-krugl/numenta-apps/commit/2ada07f7b372506471568c567a510cd36a3777b7. Would you like to give the changes a once-over?

scottpurdy:
There are a couple spots where "Heinz" is spelled "Heiz" erroneously I believe.

vitaly-krugl:
Good catch, @scottpurdy! It looks like we got some bad data from Taylor in the csv file. The one that she uploaded had the same spelling mistake.

vitaly-krugl:
@scottpurdy, I just pushed a fix for "Heiz" to this PR. FYI

PR #417, opened 2015-10-16
UNI-200: Add aggregation options

lscheinkman:
@brev, @marionleborgne  Please review


unicorn/tests/js/unit/FileServer_test.js
```diff
@@ -158,23 +164,55 @@ describe('FileServer', () => {
         }
       });
     });
+
+    it('should get aggregated data', (done) => {
+      let options = {
+        limit: 1,
+        aggregation: {
+          'timefield': 'timestamp',
```
brev: lint will complain about quoted keys

unicorn/frontend/lib/TimeAggregator.js
```diff
@@ -0,0 +1,134 @@
...
+};
+
+TimeAggregator.prototype._flush = function(done) {
+  this._publish();
+  done();
+};
+
+TimeAggregator.prototype._reset = function() {
+  let value;
+  switch (this._function) {
```
brev: lint may complain about switch statements, but I'd say change the lint rule. I prefer `switch`/`case` sometimes too.

unicorn/frontend/lib/TimeAggregator.js
```diff
@@ -0,0 +1,134 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+// See the GNU Affero Public License for more details.
+//
+// You should have received a copy of the GNU Affero Public License
+// along with this program.  If not, see http://www.gnu.org/licenses.
+//
+// http://numenta.org/licenses/
+var util = require('util');
```
brev: does this need to be `var`?

brev:
:+1: much better method names, thanks

unicorn/frontend/lib/TimeAggregator.js
lscheinkman: Actually the current lint rules complained about the indentation. I am ok with either one as long as we are consistent.
Once you merge the new lint rules I will fix on a separate PR.

unicorn/frontend/lib/TimeAggregator.js
lscheinkman: @brev Unfortunately I cannot use ES6 when inheriting from `node stream.Transform`. I've tried different ways and this was the only way it worked. I will revisit once update `node` to a newer version, hopefully they will fix it

PR #419, opened 2015-10-16
Ensure that the nupic build script works for both 0.3.4 and earlier and 0.3.5 and later

jcasner:
@scottpurdy please review


infrastructure/infrastructure/utilities/nupic/build_commands.py
```diff
@@ -97,6 +97,32 @@ def isReleaseVersion(nupicBranch, nupicSha):
 
 
 
+def isVersionGreaterOrEqual(testVersion, comparisonVersion):
```
scottpurdy: You can use pkg_resources for this:
https://pythonhosted.org/setuptools/pkg_resources.html#parsing-utilities

scottpurdy:
Done with review

infrastructure/infrastructure/utilities/nupic/build_commands.py
jcasner: This is while building the package, so I need to read from the VERSION file - I don't have a package to read yet

PR #420, opened 2015-10-16
Taurus mobile ios

lscheinkman:
@jcasner Please review

CC: @cbaranski @dm1973


jcasner:
@lscheinkman if you're happy with the work that @dm1973 has done, then this looks good to me.

PR #423, opened 2015-10-20
Removed E1101 and E1103 from the disable list in pylintrc, because th…

vitaly-krugl:
Removed E1101 and E1103 from the disable list in pylintrc, because this was hiding erroneous member accesses. It's understood that removing those error numbers will result in errors on dynamic members, and those will need to be disabled in code on a case-by-case basis.
Removed deprecated setting include-ids.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -62,11 +62,6 @@ load-plugins=
 # multiple time (only on the command line, not in the configuration file where
 # it should appear only once).
 #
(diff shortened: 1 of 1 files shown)


vitaly-krugl:
Discussed with @scottpurdy. Merging . . .

CC @oxtopus

scottpurdy:
:+1:

PR #427, opened 2015-10-20
TAUR-1538 Update rabbitmq version & infrastructure module changes

unixorn:
## Infrastructure:
Changed our requirements to be `>=`, not `==`. We had run into an issue where taurus specified an exact version of 3.11 for PyYAML and infrastructure was specifying exactly 3.10.

The infrastructure module needs to be less picky so that it won't conflict with our other modules.

## RabbitMQ
Deal with upstream update, 3.5.4 is no longer available.

TODO: [DEVOPS-31](https://jira.numenta.com/browse/DEVOPS-31) - Update the carbonite repo with all the rpms we're currently using so we don't have to upgrade packages just because upstream did.

@jcasner please CR

### DIFF
--- a/infrastructure/requirements.txt
+++ b/infrastructure/requirements.txt
@@ -1,4 +1,4 @@
-boto>=2.38.0
-fabric>=1.9.0
+boto==2.38.0
+fabric==1.9.0
 grokcli
-PyYAML>=3.10
+PyYAML==3.11
--- a/infrastructure/requirements.txt
+++ b/infrastructure/requirements.txt
@@ -1,4 +1,4 @@
-boto==2.38.0
-fabric==1.9.0
+boto>=2.38.0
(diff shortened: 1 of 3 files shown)


infrastructure/requirements.txt
```diff
@@ -1,4 +1,4 @@
-boto==2.38.0
-fabric==1.9.0
+boto>=2.38.0
```
jcasner: why are we unlocking version numbers?

jcasner:
Fix the infrastructure/requirements.txt versions, then :+1: :ship: it

infrastructure/requirements.txt
unixorn: You're right; we should keep them specific, it is better practice to make sure all the requirements.txt files in numenta-apps keep their shared requirements in sync.

PR #430, opened 2015-10-21
DEVOPS-43 Add helper script to publish python modules from jenkins

unixorn:
@jcasner, this will let us have jenkins publish the infrastructure module to our pypi when it gets changed.


infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,200 @@
...
+
+def _parseCLIArgs():
+  """
+  Parse the command line arguments
+
+  :returns: Parsed command line arguments
+  :rtype: argparse.Namespace
+  """
+  # Set up our defaults from the environment
+  workspaceD = os.environ.get("WORKSPACE", ".")
```
jcasner: there's a function that does this in `infrastructure.utilities.jenkins` I think

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,200 @@
...
+  repoDirectory = "numenta-apps"
+  fork=os.environ.get("FORK", "git@github.com:Numenta/numenta-apps.git")
```
jcasner: spaces around `=`

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,200 @@
...
+  commitish=os.environ.get("COMMITISH", "master")
```
jcasner: spaces around `=`

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,200 @@
...
+  pypiServer=os.environ.get("PYPI_SERVER", "numenta-pypi")
```
jcasner: spaces around `=`

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,200 @@
...
+  """
+  Creates/updates the git repo in workspace.
+
+  :param str workspace: path to workspace directory
+  :param str repoDirectory: directory name of the git repository
+  :param str fork: Which fork to check out
+  :param str commitish: What commit/branch to check out
+  :param logger: An initialized logger
+  """
+  assert os.path.isdir(workspace), ("workspace %s is not a directory" %
```
jcasner: why only assert workspace and not the other params?

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,200 @@
...
+                                    workspace)
+  fullRepoPath = os.path.join(workspace, repoDirectory)
+  if os.path.exists(fullRepoPath):
+    logger.info("Updating git repository in %s", repoDirectory)
+    with changeToWorkingDir(fullRepoPath):
+      logger.debug("Cleaning git repository...")
+      runWithOutput(command=("git", "clean", "-fd"), logger=logger)
```
jcasner: do we have a command in `git.py` for this?  If not, let's add it there...

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,200 @@
...
+    with changeToWorkingDir(workspace):
+      git.clone(logger=logger,
+                gitURL=fork,
+                directory=repoDirectory)
+
+
+  with changeToWorkingDir(fullRepoPath):
+    logger.debug("Setting origin remote to %s", fork)
+    runWithOutput(logger=logger,
+                  command=("git", "remote", "set-url", "origin", fork))
```
jcasner: are you sure we want to change the `origin` remote?

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,200 @@
...
+  :param bool dryrun: Whether to upload to our pypi server
+  """
+  setupPath = os.path.join(moduleDirectory, "setup.py")
+  assert os.path.isfile(setupPath), "Could not find setup.py at %s" % setupPath
+
+  buildArgs = ["python", "setup.py",
+               "sdist", "bdist_wheel", "bdist", "bdist_egg"]
+
+  # Only upload if building on jenkins unless --force-upload in CLI arguments
+  if ("JENKINS_HOME" in os.environ) or forceUpload:
```
jcasner: i don't think you need the parens here - have you run pylint?

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,200 @@
...
+    buildArgs = buildArgs + ["upload", "-r", pypiServer]
```
jcasner: `buildArgs.extend(["upload", "-r", pypiServer])`

jcasner:
@unixorn I'm done with this round of review

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
unixorn: I want the jenkins job to be able to specify a specific fork to use.

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
unixorn: Pylint came back clean on this file

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
unixorn: At this point in the script, there isn't a logger to pass to `jenkins.getWorkspace`

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
unixorn: Added more asserts for the other parameters

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
jcasner: Changing the origin remote url is probably the wrong way to do this...  Instead, just change the pull remote to be the URL.  For example, instead of `git pull jcasner master` do `git pull git@github.com:jcasner/numenta-apps.git master`

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
jcasner: Then this line is probably at the wrong place in this file...

unixorn: I need WORKSPACE to generate the paths that are going to be the default values for the command line arguments.

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
jcasner:
```python
$ python
>>> forceUpload = True
>>> import os
>>> if "JENKINS_HOME" in os.environ or forceUpload:
...   print "True"
...
True
```

Parens are superfluous - pylint should have caught them and the missing spaces around the `=` sign earlier... I'll talk to Scott about it

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
jcasner: You can set this after initializing the logger if the value is `None`.  It's a better approach to ensure consistency across implementations.

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
unixorn: I didn't realize I could pull without using a defined remote. Definitely a better approach.

unixorn:
Ready for another CR pass, @jcasner

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
jcasner: This still needs to be moved

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
jcasner: You should probably also have a separate assert for `workspace` to ensure it's a `basestring`.  Otherwise, you could get a `TypeError` instead of an `AssertionError`:

```python
>>> import os
>>> workspace = True
>>> assert os.path.isdir(workspace)
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/genericpath.py", line 49, in isdir
    st = os.stat(s)
TypeError: coercing to Unicode: need string or buffer, bool found
>>> workspace = "~"
>>> assert os.path.isdir(workspace)
AssertionError
```

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,212 @@
...
+def setupWorkspace(workspace, repoDirectory, fork, commitish, logger):
+  """
+  Creates/updates the git repo in workspace.
+
+  :param str workspace: path to workspace directory
+  :param str repoDirectory: directory name of the git repository
+  :param str fork: Which fork to check out
+  :param str commitish: What commit/branch to check out
+  :param logger: An initialized logger
+  """
```
jcasner: `:raises AssertionError:`

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,212 @@
...
+def buildAndUploadModule(logger, moduleDirectory,
+                         pypiServer, forceUpload=False):
+  """
+  Builds and uploads the a python module to our pypi server
+
+  :param logger: An initialized logger
+  :param str moduleDirectory: path to directory containing setup.py
+  :param str pypiServer: which pypi server to upload to.
+  :param bool dryrun: Whether to upload to our pypi server
+  """
```
jcasner: `:raises AssertionError:`

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,212 @@
...
+    logger.debug("Cloning new git repository from %s...", fork)
+    with changeToWorkingDir(workspace):
+      git.clone(logger=logger,
+                gitURL=fork,
+                directory=repoDirectory)
+
+
+  with changeToWorkingDir(fullRepoPath):
+    logger.debug("Pulling from %s", fork)
+    runWithOutput(command=("git", "pull", fork), logger=logger)
```
jcasner: can we add a `pull` function to the `git` utility please?

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,212 @@
...
+  assert isinstance(pypiServer, basestring), (
+    "pypiServer is %r instead of string" % pypiServer)
+  assert isinstance(forceUpload, bool), (
+    "forceUpload is %r instead of bool" % forceUpload)
+
+  setupPath = os.path.join(moduleDirectory, "setup.py")
+  assert os.path.isfile(setupPath), "Could not find setup.py at %s" % setupPath
+
+  buildArgs = ["python", "setup.py",
+               "sdist", "bdist_wheel", "bdist", "bdist_egg"]
```
jcasner: Do we really need a source distribution?  Seems like we really just need `bdist_wheel` and `bdist_egg`, but not a big deal either way...

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
```diff
@@ -0,0 +1,212 @@
...
+
+  # Only upload if building on jenkins unless --force-upload in CLI arguments
+  if forceUpload or "JENKINS_HOME" in os.environ:
+    buildArgs.extend(["upload", "-r", pypiServer])
+  else:
+    logger.debug("Not being run by jenkins, skipping pypi upload")
+
+  with changeToWorkingDir(moduleDirectory):
+    logger.info("Building in %s", moduleDirectory)
+    logger.debug("Running: %s", " ".join(buildArgs))
```
jcasner: if you pass the logger and env vars to `runWithOutput` it will print the command and env vars as a debug statement

jcasner:
@unixorn done with this round of review

infrastructure/pipelines/infrastructure-pypipeline/publish-python-module
unixorn: Doing it more for completion than anything else. We don't need it now but it might be handy later.

unixorn:
Ready for another round, @jcasner

infrastructure/infrastructure/utilities/cli.py
```diff
@@ -55,7 +55,8 @@ def executeCommand(command, env=None, logger=None):
     if env is None:
       env = os.environ
     if logger is not None:
-      diagnostics.printEnv(env, logger)
+      if "STIFLE_ENVIRONMENT_CONSOLE_SPAM" not in os.environ:
```
jcasner: :-1: 

`diagnostics.printEnv()` prints at the `debug` level.  I don't like having 2 different debug levels available.  Let's take this out.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -445,6 +444,27 @@ def fetch(repository, refspec, logger):
   return executeCommand(command=command, logger=logger)
 
 
+def pull(arguments, logger):
+  """
+  Do a git pull.
+  :param str arguments: Any extra arguments you want to pass to git pull
```
jcasner: Give some examples here.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -445,6 +444,27 @@ def fetch(repository, refspec, logger):
...
+  :param logger: An initialized logger object
+  """
+  if arguments:
```
jcasner: arguments is a required field, so no need for `if arguments`.  Otherwise, you need to make `arguments=None` in the definition

jcasner:
@unixorn ready for another pass

infrastructure/infrastructure/utilities/cli.py
unixorn: The default behavior remains the same. This change just allows people to not have to wade through the environment for every single external command they run, which is really really annoying when I'm not changing the environment at all. There are so many lines of environment variable output that I don't get to see two real lines in a normal sized terminal window at the same time.

I made the behavior change be triggered by an environment variable so that no one else has to use the trimmed output unless they want to and no one will accidentally trigger the behavior.

jcasner: I would just remove this section entirely... If you call `runWithOutput` or `executeCommand` with a logger, then it will call `diagnostics.printEnv` anyhow, which is why you're getting the double output.  It only makes sense to include it right before running the command...

unixorn: I don't understand why you feel so strongly that it is not ok to have an option to suppress environment output. I don't need to see the environment variable dump every single time a command is run and actively don't want to. Every time `cli.executeCommand` or `cli.runWithOutput` is called I see literally 94 extra lines of environment output I don't want or need which makes it difficult to pick out the information I am looking for.

Scripts like the jenkins pipelines would continue to print the environment to the console, this would only suppress the output for users that go out of their way to set an environment variable.

No one has to use it if they don't want it and the default behavior makes working with utilities.cli annoying for me every time I run a script during development.

Reverting.

infrastructure/infrastructure/utilities/git.py
```diff
@@ -445,6 +444,34 @@ def fetch(repository, refspec, logger):
...
+            isinstance(arguments, list)), (
+              "arguments must be a string or list, but is %r" % arguments)
+  assert logger
+  gitCommand = ["git", "pull"]
+  if arguments:
+    if isinstance(arguments, list) or isinstance(arguments, tuple):
+      gitCommand.extend(arguments)
+    if isinstance(arguments, basestring):
+      arguments.extend(str.split(arguments))
+  logger.debug("* Running %s", gitCommand)
```
jcasner: `executeCommand` already logs the command at the debug level.

jcasner:
Remove the last superfluous line and then :+1:

PR #431, opened 2015-10-21
ENG-40 Parse replacement parameters using newer argparse, as opposed …

oxtopus:
…to deprecated optparse.  Apply replacements using newly added `_recurseDictAndReplace()` function

This adds a new optional `--replaceParam` command line argument to the unicorn model  runner.  For example, to encode timestamps as day of week rather than time of day, as well as enable verbose output in the encoder:

```
python unicorn_backend/model_runner.py \
  --model "1" \
  --stats '{"max": 10, "min": 0}'  \
  --replaceParam modelConfig/modelParams/sensorParams/verbosity 1 \
  --replaceParam modelConfig/modelParams/sensorParams/encoders/c0_dayOfWeek '{"dayOfWeek": [21, 1], "type": "DateEncoder", "fieldname": "c0", "name": "c0"}' \
  --replaceParam modelConfig/modelParams/sensorParams/encoders/c0_timeOfDay null
```

Using the dummy data emitter with the params above yields the following output:

```
RecordSensor got data: {'_timestamp': datetime.datetime(2015, 8, 4, 0, 55, 11), '_sequenceId': 0, '_category': [None], 'c0': datetime.datetime(2015, 8, 4, 0, 55, 11), '_timestampRecordIdx': None, 'c1': 835.93679, '_reset': 0}
sensor: day of week                                                                                                                                         | c1                                                                                                                                                                                                                                                                                                                                                                                                               |
sensor: ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
      0: array('c', '...........*********************...................................................................................................................') | array('c', '.........................................*........................*.............*.............*.....*..................**............................*.....................*..............................................*................*.............................*................*.................*...........*.*.*..............*.*............*..............*......................................') |
     nz: (42) [ 11  12  13  14  15  16  17  18  19  20  21  22  23  24  25  26  27  28
  29  30  31 188 213 227 241 247 266 267 296 318 365 382 412 429 447 459
 461 463 478 480 493 508]
  encIn: c0.day of week:1.00, c1:835.94
[0, 0.0301029996658834]
```

### DIFF
--- a/unicorn/backend/unicorn_backend/model_runner.py
+++ b/unicorn/backend/unicorn_backend/model_runner.py
@@ -28,7 +28,7 @@ import sys
 
 
 def _scriptIsFrozen():
-  """ Returns True if all of the modules are built-in to the interpreter by 
+  """ Returns True if all of the modules are built-in to the interpreter by
   cxFreeze, for example.
   """
   return hasattr(sys, "frozen")
@@ -44,7 +44,7 @@ if _scriptIsFrozen():
 from datetime import datetime
 import json
 import logging
-from optparse import OptionParser
+from argparse import ArgumentParser
 import traceback
 
 import validictory
@@ -63,6 +63,12 @@ g_log = logging.getLogger(__name__)
 
 
 
+_REPLACE_PATH_SEPARATOR = "/" # Path separator between components in
+                              # replacement parameters.
+                              # e.g. modelConfig/modelParams/clParams
+
+
+
 class _CommandLineArgError(Exception):
   """ Error parsing command-line options """
   pass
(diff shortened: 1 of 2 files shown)


oxtopus:
integration test is currently failing for me.  Hold off until further notice...

oxtopus:
Ok, Fixed!  Integration tests pass now.

unicorn/backend/unicorn_backend/model_runner.py
```diff
@@ -282,9 +347,7 @@ def main():
   g_log.addHandler(logging.NullHandler())
   try:
 
-    options = _parseArgs()
-
-    _ModelRunner(modelId=options.modelId, stats=options.stats).run()
+    _ModelRunner(**vars(_parseArgs())).run()
```
scottpurdy: I'd really prefer to pull out the keys explicitly. Yes, more typing. But way more explicit.

oxtopus: I tend to agree, but the original attempt was excessive.  `_parseArgs()` returns an instance of `_Options` (see https://github.com/oxtopus/numenta-apps/blob/ENG-40-model-params/unicorn/backend/unicorn_backend/model_runner.py#L78-L94).  The sole purpose for which is to be a container for arguments.

If we were to get rid of the `_Options` implementation, for sure, I'd say don't do the kwarg expansion.  But given the `_Options` implementation, leaving it there only adds maintenance overhead.

scottpurdy: I see. The problem with _Options is that the code you pass it into has to understand the _Options object format. I think it would be cleaner [using values as the boundary](https://www.destroyallsoftware.com/talks/boundaries).

That said, this isn't new code so may be outside the scope of this change. Up to you.

PR #432, opened 2015-10-21
ENG-84 Add new console script entry point to wait for the supervisor …

oxtopus:
…api is not available, indicating that supervisord has stopped, simplify logic in taurus initialize/refresh/update scripts

### DIFF
--- a/nta.utils/nta/utils/tools/supervisord_state.py
+++ b/nta.utils/nta/utils/tools/supervisord_state.py
@@ -82,7 +82,7 @@ def getStateMain():
 def waitForRunningStateMain():
   """Console script entry point: waits for supervisord to enter RUNNING state;
   sets non-zero exit code on timeout or other non-recoverable error; prints
-  status and error messages to STDERR.
+  status to STDOUT and error messages to STDERR.
   """
 
   parser = ArgumentParser(
@@ -107,7 +107,7 @@ def waitForRunningStateMain():
       break
 
     if i <= maxWaitCycles:
-      print >> sys.stderr, "Waiting for supervisord to enter RUNNING state..."
+      print "Waiting for supervisord to enter RUNNING state..."
       time.sleep(5)
   else:
     sys.exit("Timed out waiting for supervisord RUNNING state from %s" % (
@@ -140,7 +140,7 @@ def waitForStoppedStateMain():
       break
 
     if i <= maxWaitCycles:
-      print >> sys.stderr, "Waiting for supervisord to stop..."
+      print "Waiting for supervisord to stop..."
       time.sleep(5)
   else:
     sys.exit("Timed out waiting for supervisord to stop")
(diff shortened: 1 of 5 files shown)


nta.utils/nta/utils/tools/supervisord_state.py
```diff
@@ -113,6 +114,37 @@ def waitForRunningStateMain():
...
+    try:
+      getSupervisordState(args.supervisorApiUrl)
+    except (socket.error, xmlrpclib.Fault):
+      # API is no longer available at specified supervisorApiUrl.  It is safe
+      # to assume that IF the supervisordApiUrl is correct, that supervisor is
+      # not running.  In which case, break out of loop and return.
+      break
+
+    if i <= maxWaitCycles:
+      print >> sys.stderr, "Waiting for supervisord to stop..."
```
scottpurdy: I don't know if this should go to stderr since it isn't necessarily a problem. It is only a problem if we get to `sys.exit` below, right?

scottpurdy:
Done reviewing.

oxtopus:
@scottpurdy I updated it to print to stderr only for errors.  Merging now.  Thanks!

nta.utils/nta/utils/tools/supervisord_state.py
vitaly-krugl: @scottpurdy: I recall that we had discussions and agreed upon the convention of using stderr for status/diagnostic messages based on prior discussions. Related to this, python console logger by default logs to stderr, and everything goes to stderr: DEBUG, INFO, etc. - not just ERROR and CRITICAL. stdout is typically used to output information that was specifically requested of a script whose purpose it is to provide that information. This includes various utilities, such as `sort`, `cut`, etc.

scottpurdy: Ok I think I remember that. Should we be using logging with initTool then to enforce consistency and restrict print statements to the expected/requested console output?

vitaly-krugl: I personally prefer to use logging for tools like this, partially because of the timestamps that on occasion are helpful for debugging. I am okay with console output, too, as long as we output status and diagnostic messages to stderr. Because existing code tends to set precendent as well as getting reused, this also helps to defend our code base from subtle errors, such as unexpected status/diagnostic output corrupting legitimate stdout data streams.

vitaly-krugl: @scottpurdy: here is an example that demonstrates the use of stderr also for prompting in a standard unix utility on MacOSX:

The sequence below shows that the prompt "override r--r--r--  vkruglikov/staff for test-delete-prompt//read-only-file?" is emitted to stderr (not stdout):
```
vkruglikov@vkruglikovs-MacBook-Pro:~$ mkdir test-delete-prompt
vkruglikov@vkruglikovs-MacBook-Pro:~$ touch test-delete-prompt/read-only-file
vkruglikov@vkruglikovs-MacBook-Pro:~$ chmod 0444 test-delete-prompt/read-only-file 
vkruglikov@vkruglikovs-MacBook-Pro:~$ rm -r test-delete-prompt/
override r--r--r--  vkruglikov/staff for test-delete-prompt//read-only-file? ^C
vkruglikov@vkruglikovs-MacBook-Pro:~$ rm -r test-delete-prompt/ 2>/dev/null
^C
vkruglikov@vkruglikovs-MacBook-Pro:~$
```

PR #412, opened 2015-10-14
ENG-51 Use nta.utils retry decorator in numeta-apps

SaganBolliger:
Fixes https://jira.numenta.com/browse/ENG-51

PR #433, opened 2015-10-21
ENG-51 Replace use of nupic.support.decorators.retry with retry decorator from nta.utils

SaganBolliger:
Fixes https://jira.numenta.com/browse/ENG-51
This is the next PR for this issue after #412 had to be reverted by #415 

This passes all tests for htm.it and all but one test for taurus, which also fails on master, as described in: https://jira.numenta.com/browse/TAUR-1578

@scottpurdy This is ready for review.


oxtopus:
:+1: Thanks!

PR #434, opened 2015-10-21
Use built-in 'valid-jsdoc' rule to validate JSDoc instead of `jsdoc` lint plugin

lscheinkman:
@brev The current version of `jsdoc` is causing Atom editor to hang when `linter-eslint` plugin is enable.
For now use built-in 'valid-jsdoc' rule to validate JSDoc and disable `jsdoc` lint plugin until a more stable version is available.


brev:
:+1:

PR #435, opened 2015-10-22
ENG-27 unicorn backend unit tests

oxtopus:
Unit tests invoked w/ `python setup.py test`.  

e.g.:

```
$ python setup.py test
running test
running egg_info
writing requirements to unicorn_backend.egg-info/requires.txt
writing unicorn_backend.egg-info/PKG-INFO
writing top-level names to unicorn_backend.egg-info/top_level.txt
writing dependency_links to unicorn_backend.egg-info/dependency_links.txt
reading manifest file 'unicorn_backend.egg-info/SOURCES.txt'
writing manifest file 'unicorn_backend.egg-info/SOURCES.txt'
running build_ext
============================================================================== test session starts ===============================================================================
platform darwin -- Python 2.7.10 -- pytest-2.5.1
plugins: cov, xdist
collected 7 items 

../tests/py/integration/model_runner_test.py .....
../tests/py/unit/model_runner_test.py ..

============================================================================ 7 passed in 3.89 seconds ============================================================================
```

cc @vitaly-krugl


unicorn/backend/README.md
```diff
@@ -13,6 +13,10 @@ To generate mock data and feed it to the model runner, run:
  python unicorn_backend/mock_data_generator.py | python unicorn_backend/model_runner.py --model "1" --stats '{"max": 10, "min": 0}'
 ```
 
-This will output the model output to stdout. 
+This will output the model output to stdout.
```
scottpurdy: `This will write the model output...`

unicorn/backend/requirements.txt
```diff
@@ -3,7 +3,3 @@ validictory==0.9.1
 
 # NuPIC HTM
 nupic==0.3.4
-
-# required by tests
-mock==1.0.1
```
scottpurdy: Are these not used at all or implicitly picked up elsewhere?

unicorn/backend/setup.py
```diff
@@ -23,11 +29,38 @@
...
 setup(
   name = "unicorn_backend",
   description = "Unicorn backend",
   packages = find_packages(),
   include_package_data=True,
   install_requires = installRequires,
-  dependency_links = dependencyLinks
+  dependency_links = dependencyLinks,
+  tests_require=["mock==1.0.1", "nta.utils>=0.0.0"],
+  cmdclass = {"test": PyTest},
```
scottpurdy: Not a bad way to integrate unit tests.

unicorn/backend/requirements.txt
scottpurdy: Nvm, I see in setup.py.

unicorn/backend/setup.py
```diff
@@ -23,11 +29,38 @@
...
+
 setup(
   name = "unicorn_backend",
   description = "Unicorn backend",
   packages = find_packages(),
   include_package_data=True,
   install_requires = installRequires,
-  dependency_links = dependencyLinks
+  dependency_links = dependencyLinks,
+  tests_require=["mock==1.0.1", "nta.utils>=0.0.0"],
```
scottpurdy: This seems risky because if one of these is needed for the actual code (not just tests) then we won't catch it since the only way to test is to include these. But if we installed the package without tests then it would be broken and we can't test it...

unicorn/tests/py/unit/model_runner_test.py
```diff
@@ -0,0 +1,94 @@
...
+  @staticmethod
+  @patch("sys.stdout.write", autospec=True)
+  @patch("sys.stdout.flush", autospec=True)
+  @patch("sys.stdin.readline",
+         side_effect=["[1438649711, 835.93679]\n"],
+         autospec=True)
+  def testModelRunnerAtBoundaries(readlineMock, stdoutFlushMock,
+                                  stdoutWriteMock):
+    """ _ModelRunner().run() reads from STDIN and writes to STDOUT
+    """
```
scottpurdy: You can end the doc string on first line for one line doc strings. At least that is what I do based on the examples in [PEP-257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings).

unicorn/tests/py/unit/model_runner_test.py
```diff
@@ -0,0 +1,94 @@
...
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""Unit test of the unicorn_backend.model_runner module"""
+
+import logging
+from mock import patch
```
scottpurdy: mock isn't a standard library package so I usually put it in the third-party import section.

scottpurdy:
Done reviewing. I like the unit test. I am a bit worried about introducing new test entry points and test-specific dependencies. I don't want to take too much time but perhaps you could send an email proposal to relevant people for discussion? I actually quite like the `python setup.py test` option but it doesn't really handle unit vs. integration tests and wouldn't allow running the tests from an installed package (as opposed to source).

oxtopus:
Thanks, @scottpurdy.  See latest in which I've fixed handling of command line arguments.  By default `python setup.py test` runs all tests, but you can use `-a` to append arguments to `pytest`.  For example, `python setup.py test -a unit` runs unit tests only, `python setup.py test -a integration` runs integration only.  The full api for py.test is available this way.  For example, to run only a specific test using the py.test `-k` argument:

```
$ python setup.py test -a "-k testModelRunnerAtBoundaries"
running test
running egg_info
writing requirements to unicorn_backend.egg-info/requires.txt
writing unicorn_backend.egg-info/PKG-INFO
writing top-level names to unicorn_backend.egg-info/top_level.txt
writing dependency_links to unicorn_backend.egg-info/dependency_links.txt
reading manifest file 'unicorn_backend.egg-info/SOURCES.txt'
writing manifest file 'unicorn_backend.egg-info/SOURCES.txt'
running build_ext
============================================================================== test session starts ===============================================================================
platform darwin -- Python 2.7.10 -- pytest-2.5.1
plugins: cov, xdist
collected 7 items 

unit/model_runner_test.py .

============================================================= 6 tests deselected by '-ktestModelRunnerAtBoundaries' ==============================================================
===================================================================== 1 passed, 6 deselected in 0.67 seconds =====================================================================
```

unicorn/backend/setup.py
oxtopus: Doing this was the result of discussion among @vitaly-krugl and @marionleborgne as a result of `nta.utils` only being needed for tests, but causing problems during build when it wasn't needed.

scottpurdy:
@oxtopus - We should still discuss this way of doing tests. Defaulting to all tests, for instance, could be misleading in places like NuPIC where integration tests require MySQL, or Taurus where they could have a detrimental effect on a running system. We don't need to hold this change up though, I will follow up with relevant discussions.

PR #442, opened 2015-10-27
UNI-215: Add File Details

lscheinkman:
@brev Please review

CC: @marionleborgne


marionleborgne:
@lscheinkman a couple of high level comments:

1.If there are no header rows (meaning the first row are not strings only) then we should probably name the columns for the users with `Metric 1`,  `Metric 2`, etc... I tested it with a csv file that had no headers and the checkboxes to create models automatically look weird. The metric names are the first values in the CSV. So my checkboxes look like that
```
[x] 4.0
[x] 3.1
```

2. I don't think we should assume that the timestamp is the first column always. Right now, it seems that we assume that the timestamp is columns[0] so we remove this metric from the list of metrics right away. Instead, we should probably check that we have at least one column with a few datetime values. And then we can say that this is the timestamp column. What do you think @lscheinkman ?

3. Finally, it looks like the `File Size` field in the UI is editable. It gets highlighted when I mouse over it. Are you planning to remove this later?

I'll continue looking at the PR. Thanks!

lscheinkman:
@marionleborgne 
Thank you for the review, here are some of the answers to the feedback.

> 1 - If there are no header rows (meaning the first row are not strings only) then we should probably name the columns for the users with Metric 1, Metric 2, etc..

There is no easy way to guess whether the first row is header or data. This is just a convention. Some applications use a flag when they import the data and let the user decide. We can try to guess but it is not as reliable as asking the user.

> 2 - I don't think we should assume that the timestamp is the first column always.

We are not assuming timestamp is the first column but we are assuming there is only **one** timestamp field. See https://github.com/lscheinkman/numenta-apps/blob/UNI-215/unicorn/frontend/browser/components/FileDetails.jsx#L196-L198

> 3 - Finally, it looks like the File Size field in the UI is editable

This is a bug, it should not be editable.

brev:
:+1: LGTM!

PR #443, opened 2015-10-28
DEVOPS-19 Create base infrastructure AMI using Centos 7

jcasner:


jcasner:
@scottpurdy this is still a work in progress but I wanted to make sure to get something into github before I go...  Hopefully will finish this before EOD tomorrow but either way, you've got a good start...

jcasner:
@scottpurdy merged this early because it isn't getting pulled from salt-master currently.

PR #446, opened 2015-10-28
ENG-85 Emit ERROR-level log message when model_runner process completes with non-0 exit status

vitaly-krugl:


vitaly-krugl:
CC @scottpurdy

htmengine/htmengine/model_swapper/slot_agent.py
```diff
@@ -349,8 +350,9 @@ def stopModel(doCallback):
           "%r: {TAG:SWAP.SA.MODEL.STOP.PENDING} modelState=%r",
           self, modelState)
         modelState.modelExitStatus = modelState.modelRunner.stopGracefully()
-        self._logger.info("%r: {TAG:SWAP.SA.MODEL.STOP.DONE} modelState=%r",
-                          self, modelState)
+        self._logger.log(
+          logging.INFO if modelState.modelExitStatus == 0 else logging.ERROR,
```
scottpurdy: I think it would be easier to read as two separate logging calls inside an if/else. If you don't like that, can you at least pull out `logging.INFO if modelState.modelExitStatus == 0 else logging.ERROR` onto a line before:

```
logLevel = logging.INFO if modelState.modelExitStatus == 0 else logging.ERROR
self._logger.log(logLevel, ...)
```

But I'd prefer if/else blocks with separate info/error logging statements since that seems easier to read to me.

vitaly-krugl:
@scottpurdy's feedback is addressed via #449

PR #447, opened 2015-10-28
Update pipeline_controller to work with the new pipeline names

jcasner:
@vitaly-krugl this will fix the issue you noticed


vitaly-krugl:
:thumbsup:

PR #449, opened 2015-10-29
ENG-85 Address Scott's feedback regarding log-level logic in SlotAgent._runEventLoop.stopModel closure. Add unit test for failed model runner.

vitaly-krugl:
@scottpurdy, please review. This addresses your feedback from #446


scottpurdy:
:+1:

PR #455, opened 2015-11-03
fix test location calculation

marionleborgne:
@oxtopus @scottpurdy @rhyolight 

Fixing the error that happened with #454 when running `python setup.py test`


oxtopus:
:+1:

PR #456, opened 2015-11-04
Update to 'fluxible:1.0.1', 'react:0.14.2' and 'material-ui:0.13.1

lscheinkman:
@brev Need to update to new version of material-ui which requires new version of react and other dependencies.


brev:
:+1:  LGTM thanks, glad to have that done

PR #459, opened 2015-11-06
ENG-95 Override Taurus Engine's REST API Key in Taurus pipelines

vitaly-krugl:
ENG-95 Impelement taurus-set-api-key console script for overriding Taurus Engine's REST API Key.

ENG-95 Updated Taurus pipeline scripts to override Taurus Engine's REST API Key using the new tool.

### DIFF
--- a/taurus/conf/application.conf
+++ b/taurus/conf/application.conf
@@ -40,7 +40,6 @@ plaintext_port = 2003
 queue_name = taurus.metric.custom.data
 
 [security]
-# TAUR-1065 Remove this from the conf file and replace with something else (env var??)
 apikey = taurus
 
 [anomaly_likelihood]
--- a/taurus/taurus/engine/set_api_key.py
+++ b/taurus/taurus/engine/set_api_key.py
@@ -1,0 +1,87 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have purchased from
+# Numenta, Inc. a separate commercial license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero Public License for more details.
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""Applies the given Taurus REST API key as override for application.conf."""
+
+
+import logging
+import __main__
+import argparse
+
+from nta.utils.config import Config
+
+import taurus.engine
+from taurus.engine import logging_support
+
+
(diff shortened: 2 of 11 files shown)


taurus/conf/application.conf
```diff
@@ -40,7 +40,6 @@ plaintext_port = 2003
 queue_name = taurus.metric.custom.data
 
 [security]
-# TAUR-1065 Remove this from the conf file and replace with something else (env var??)
```
vitaly-krugl: TAUR-1065 was closed

vitaly-krugl:
Merging as hotfix to expedite resolution of failure on staging.

Please review

CC @scottpurdy, @oxtopus

taurus/taurus/engine/set_api_key.py
```diff
@@ -0,0 +1,87 @@
...
+
+g_log = logging.getLogger(__name__)
+
+
+
+def main():
+  logging_support.LoggingSupport().initTool()
+
+
+  parser = argparse.ArgumentParser(description=__main__.__doc__)
```
oxtopus: Better to use `__doc__` alone as opposed to `__main__.__doc__`, as `__main__` may change, depending on how the code is run, and `__main__.__doc__` will refer to the docstring of the `__main__` in which this code is executed (and not this module).

taurus/taurus/engine/set_api_key.py
```diff
@@ -0,0 +1,87 @@
...
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""Applies the given Taurus REST API key as override for application.conf."""
+
+
+import logging
+import __main__
```
oxtopus: See my other comment about `__main__`, you shouldn't need to import `__main__`.

taurus/taurus/engine/set_api_key.py
vitaly-krugl: Good to know. Fixed.

taurus/taurus/engine/set_api_key.py
vitaly-krugl: removing import of __main__

vitaly-krugl:
@oxtopus: thanks, I addressed your feedback (including from email) via PR #463

vitaly-krugl:
@oxtopus, are you done with this code review, and are the fixes in #463 sufficient?

oxtopus:
I'm done.

taurus/taurus/engine/set_api_key.py
```diff
@@ -0,0 +1,87 @@
...
+
+  if not args.apikey:
+    msg = "Missing or empty api key"
+    g_log.error(msg)
+    parser.error(msg)
+
+
+  conf = taurus.engine.config
+
+  assert conf.has_section("security"), (
```
scottpurdy: This seems like an important check to always perform so I'd change this to an explicit if/raise. The `assert` statement is only checked when `__debug__` is set to true. Same for below.

vitaly-krugl: @scottpurdy, the implementation is based on the assumption that the conf file must contain that section + option. The asserts are here to verify this assumption. If you thing that this intended scenario warrants an explicit check, I can get it in.

scottpurdy: Oh I see, this is the configuration that ships with Taurus (possibly with some overrides), not some user-defined configuration? Fine to leave as an assert then.

vitaly-krugl: @scottpurdy, that's correct.

PR #461, opened 2015-11-06
ENG-97 Work around tweet deletion lock timeout by purging in small increments of rows

vitaly-krugl:


vitaly-krugl:
This is for a P2 issue - please review. CC @scottpurdy @oxtopus

taurus.metric_collectors/taurus/metric_collectors/twitterdirect/purge_old_tweets.py
```diff
@@ -70,29 +76,140 @@ def _parseArgs():
...
+  while totalDeleted < estimate:
+    # NOTE: we're dealing with a couple of issues here:
+    #
+    # 1. sqlalchemy core doesn't support LIMIT in delete statements, so we can't
+    #    use delete directly with LIMIT and ORDER BY
+    # 2. MySql (5.6.21) doesn't support LIMIT & IN subqueries: "This version of
+    #    MySQL doesn't yet support 'LIMIT & IN/ALL/ANY/SOME subquery"
+    #
+    # So, we're going to stick with sqlalchemy, and break the operation into two
+    # queries: get the candidate uids, then delete rows with those uids
```
vitaly-krugl: Reviewer F.Y.I.: Since this is not a mission-critical utility (performance-wise), I chose the benefit of sharing `selectionPredicate` between the estimation operation and the deletion operations (following the D.R.Y. principle) over using creating a separate hard-coded "DELETE ... ORDER BY ... LIMIT" query with its own version of the predicate.

taurus.metric_collectors/taurus/metric_collectors/twitterdirect/purge_old_tweets.py
```diff
@@ -70,29 +76,140 @@ def _parseArgs():
...
+
+  sqlEngine = collectorsdb.engineFactory()
+
+  selectionPredicate = (
     twitterTweetsSchema.c.created_at <
-    sqlalchemy.func.date_sub(
-      sqlalchemy.func.utc_timestamp(),
-      sqlalchemy.text("INTERVAL %i DAY" % (thresholdDays,)))
+    sql.func.date_sub(sql.func.utc_timestamp(),
+                      sql.text("INTERVAL %i DAY" % (thresholdDays,)))
```
oxtopus: This should use `str.format()` re: recent(ish) discussions

taurus.metric_collectors/taurus/metric_collectors/twitterdirect/purge_old_tweets.py
```diff
@@ -70,29 +76,140 @@ def _parseArgs():
 
 
 
-@collectorsdb.retryOnTransientErrors
 def purgeOldTweets(thresholdDays):
```
oxtopus: The entire process could be simplified by breaking the delete query up in such a way as to limit the delete to specific date intervals, and then iterating through those date intervals from oldest to newest.  To be extra careful, if we run into a timeout, we could reduce the intervals and retry automatically.

oxtopus:
This looks good to me, albeit a bit more complex than it needs to be.  No need to go back and simplify, however.

vitaly-krugl:
Thanks for the feedback @oxtopus. I also considered doing it by date intervals, but that solution has it's own complexities as you noted in your alternative suggestion
> To be extra careful, if we run into a timeout, we could reduce the intervals and retry automatically.
If I were to do it over, I would probably just use a native MySQL query directly, and bite the bullet on duplicating the predicate's construction.

taurus.metric_collectors/taurus/metric_collectors/twitterdirect/purge_old_tweets.py
vitaly-krugl: @oxtopus: for consistency, I went ahead and converted the use of string `%` operators in purge_old_tweets.py with `.format()`

PR #466, opened 2015-11-10
Fix ESDocs and lint issues

lscheinkman:
@brev Fixed the docs so it passes lint and generate docs for actions


brev:
:+1: great thanks

PR #480, opened 2015-11-23
Unicorn bump npm packages

brev:
Numenta-apps Unicorn
* Bump npm packages
** Skips upgrades to:
*** Node: should be on 4.2.2, but not sure what to do on https://github.com/NumentaCorp/numenta-private-salt-infrastructure/pull/320
*** Babel: Babel 6 missing decorator support still
*** Electron-prebuilt: `ipc` deprecation
* fix lints
* remove old Web mentions
* etc.


brev:
@lscheinkman @marionleborgne please review, thanks.

unicorn/js/browser/components/Chart.jsx
```diff
@@ -162,6 +162,12 @@ export default class Chart extends React.Component {
 
   /**
    * DyGrpahs Chart range finder change/zoom callback function
+   * @method
+   * @param {Number} rangeXmin - Minimum X value of chart Range Finder
+   * @param {Number} rangeXmax - Maximum X value of chart Range Finder
+   * @param {Array} [yRanges] - Extra Y value data (unused currently)
+   * @private
+   * @this Chart
```
lscheinkman: There is no need for the extra `jsdoc` tags. We are using `esdoc` and it supports `ES6` classes. See https://esdoc.org/feature.html#es6-class  and https://esdoc.org/tags.html

unicorn/js/browser/components/FileList.jsx
```diff
@@ -125,12 +136,12 @@ export default class FileList extends React.Component {
     }
   }
 
-  _handleMetricContextMenu(modelId, filename, timestampField, metric, ev, action) {
+  _handleMetricContextMenu(modelId, filename, tsField, metric, event, action) {
```
lscheinkman: Why rename `timestampField` to `tsField`?

unicorn/js/browser/components/FileList.jsx
```diff
@@ -44,14 +52,17 @@ const {
 const DIALOG_STRINGS = {
   model: {
     title: 'Delete Model',
-    message: 'Deleting this model will delete the associated model results. Are you sure you want to delete this model?'
+    message: `Deleting this model will delete the associated model results.
+              Are you sure you want to delete this model?`
```
lscheinkman: This may add extra spaces to the string.

PR #482, opened 2015-11-23
DEVOPS-63 Limit stock metrics to market hours

oxtopus:


taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -80,6 +86,110 @@ class LatencyMonitorError(TaurusMonitorError):
...
 
 
+def isStockModel(model):
+  """ Returns true if model is a stock model.  False if type is not recognized
+  as either "StockPrice" or "StockVolume".
+  :param dict model: Model dict
+  :returns: True if metric type is one of "StockPrice", "StockVolume", False
+    otherwise
+  """
+  return (model
```
vitaly-krugl: Bummer that pure statistics didn't work out. It looked so clean without the metric-specific stuff.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -80,6 +86,110 @@ class LatencyMonitorError(TaurusMonitorError):
...
+    datetime.date(2020, 9, 7),
+    datetime.date(2020, 11, 26),
+    datetime.date(2020, 12, 25)
+  }
+
+  if marketLocalTime.date() in marketClosureHolidays:
+    # market-local date is a known market holiday
+    return True
+
+  if (marketLocalTime.time() < datetime.time(10, 30) or
```
vitaly-krugl: I am confused by the numbers and the comments. I thought that our New York stock exchanges operate 9:30 AM – 4:00 PM.

oxtopus: You are correct.  If we, by chance rant the latency monitor right at, or close to 9:30, there's a chance that we could get a false positive because data hasn't fully made it through to dynamodb, yet.  Market data is delayed, too, further complicating the issue.  I purposefully generaously offset the window to account for such delays.  Is there something I can do to change the comment inside the block to make that more clear?

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
oxtopus: We could do an ensemble of multiple techniques that account for the different distributions ;)

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
vitaly-krugl: @oxtopus: If you included a mention of the market delay + dynmodb latency in the comment, that should make it clear.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -220,7 +330,7 @@ def getMetricData(self, metricUid):
       http://boto.readthedocs.org/en/latest/dynamodb2_tut.html#the-resultset)
     """
     # Query recent DynamoDB metric data for each model
-    now = datetime.datetime.utcnow()
+    now = datetime.datetime.now(_UTC_TZ)
```
vitaly-krugl: For my own understanding, does dynamodb expect the timezone offset in `gte` queries? Does it treat a timestamp without the offset as UTC?

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
```diff
@@ -278,6 +395,7 @@ def checkAllModelLatency(self):
                  # no data at all!  No point in calculating stddev.
 
       stddev = numpy.nanstd(intervals)
```
vitaly-krugl: For stock models, this incorporates weekends and holidays, while the new logic is only trying to apply this calculation during market hours. Is this what you intended? If you intended it, please add a comment that explains what's going on and why it's okay.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
oxtopus: In one case we get, for example, '2015-11-10 23:33:29+00:00', and in the other '2015-11-10 23:33:29'.  I don't think it handles timestamps at all, but instead treats them as strings.  In which case, I will need to format the timestamp to strip the timezone component so that `gte` does work properly.

taurus.monitoring/tests/unit/model_latency_monitor_test.py
```diff
@@ -103,11 +107,96 @@ def query2SideEffect(uid__eq, timestamp__gte):
...
+    def query2SideEffect(uid__eq, timestamp__gte):
+      return METRIC_DATA_BY_ID[uid__eq]
+
+    tableMock.return_value = (
+      Mock(query_2=Mock(side_effect=query2SideEffect,
+                        __name__=str(id(query2SideEffect))))
+    )
+
+    # This should not raise an exception
+    ModelLatencyChecker().checkAllModelLatency()
```
vitaly-krugl: What gives us confidence that the lack of exception here is because the logic recognized now as a market holiday and not due to another bug in logic?

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
vitaly-krugl: :)

vitaly-krugl:
@oxtopus: I am done with the review. Please take a look at the outstanding feedback regarding inclusion of weekends and holidays in stddev of stock models, question in the holidays test, and clarifying a code comment.

taurus.monitoring/taurus/monitoring/latency_monitor/model_latency_monitor.py
oxtopus: It takes into consideration both the acceptable minimum threshold and the computed threshold.  If we stop getting data for a specific metric, we'll still catch it.  I'll try to add _something_ that explains this in a comment.

oxtopus:
Thanks.  I believe I've addressed all of the feedback.

PR #493, opened 2015-12-01
Testing - add VERBOSE to environment when calling make

scottpurdy:


rhyolight:
By analyzing the blame information on this pull request, we identified @shantanoo, @akhilaananthram and @rhyolight to be potential reviewers

oxtopus:
:+1:

infrastructure/infrastructure/utilities/nupic/build_commands.py
```diff
@@ -295,7 +295,9 @@ def buildNuPICCore(env, nupicCoreSha, logger, buildWorkspace, nupicVersion):
                            pythonLibDir=libdir,
                            pythonIncludeDir=includeDir),
                       env=env, logger=logger)
-        runWithOutput("VERBOSE=1 make -j 1", env=env, logger=logger)
+        newEnv = dict(env)
+        newEnv["VERBOSE"] = "1"
+        runWithOutput("make -j 1", env=newEnv, logger=logger)
         runWithOutput("make install", env=env, logger=logger)
```
oxtopus: I don't think it would make a difference.  `runWithOutput("VERBOSE=1 make -j 1", env=env, logger=logger)` is equivalent to what @scottpurdy has done here.

scottpurdy: @vitaly-krugl - good question but `make install` just copies things from the build location to the install location so I don't think I'd learn anything from it and I'm not sure if the output would actually be different. That said, I was planning on reverting these changes after we fix the problem. Do you guys think we should keep the verbose output for future debugging?

PR #512, opened 2015-12-11
ENG-58 Implemented metric data garbage collector and integrated it into taurus engine.

vitaly-krugl:
CC @oxtopus @scottpurdy


vitaly-krugl:
CC @oxtopus @scottpurdy

taurus/conf/supervisord.conf
```diff
@@ -83,9 +83,20 @@ stdout_logfile=logs/dynamodb_service.log
 redirect_stderr=true
 stopsignal=INT
 
+;*************** METRIC DATA GARBAGE COLLECTOR **************
+[program:metric-data-garbage-collector]
+command=python -m htmengine.runtime.metric_garbage_collector --threshold-days=90
```
vitaly-krugl: Note: we integrate it here instead of `htmengine/conf/supervisord-base.conf`, because the garbage collector logic is based on metric data timestamps, which is not suitable for HTM-IT that allows custom metrics with potentially very old timestamps. If metric data rows had creation timestamps, then it might be desirable to integrate it via `htmengine/conf/supervisord-base.conf`. See docstring in `htmengine.runtime.metric_garbage_collector` for more details.

oxtopus:
:+1: LGTM, Thanks!

PR #514, opened 2015-12-13
UNI-289: Load raw data from file system if not available from database

lscheinkman:
@marionleborgne This should fix the missing raw data chart


marionleborgne:
Yes. Just tested it on my machine. Thanks @lscheinkman. Feel free to close the associated JIRA.

PR #518, opened 2015-12-14
ENG-98 Fix regression in import of ModelCheckpointMgr.

vitaly-krugl:
CC @oxtopus

### DIFF
--- a/taurus/pipeline/scripts/refresh-remote-taurus-servers.sh
+++ b/taurus/pipeline/scripts/refresh-remote-taurus-servers.sh
@@ -331,7 +331,7 @@ pushd "${REPOPATH}"
      sudo rabbitmqctl stop_app &&
      sudo rabbitmqctl reset &&
      sudo rabbitmqctl start_app &&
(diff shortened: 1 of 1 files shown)


taurus/pipeline/scripts/refresh-remote-taurus-servers.sh
```diff
@@ -331,7 +331,7 @@ pushd "${REPOPATH}"
      sudo rabbitmqctl stop_app &&
      sudo rabbitmqctl reset &&
      sudo rabbitmqctl start_app &&
-     python -c 'from htmengine.model_checkpoint_mgr import ModelCheckpointMgr; ModelCheckpointMgr.removeAll()' &&
+     python -c 'from htmengine.model_checkpoint_mgr.model_checkpoint_mgr import ModelCheckpointMgr; ModelCheckpointMgr.removeAll()' &&
```
oxtopus: For consistency with other pipeline operations, this should be wrapped in a console script entry point.

PR #521, opened 2015-12-14
IMBU-49 Update README

oxtopus:
@BoltzmannBrain Also note that I removed `imbu/engine/imbu_tp.json` as it is not being used.


BoltzmannBrain:
Please add the config json back in because we'll be using it very soon to run HTM models.
I left one nitpick, but :+1:

PR #525, opened 2015-12-16
ENG-98 Implemented Taurus refresh CI pipeline's reset of Taurus Engine's message bus resources, repository, and model checkpoints via distutils command extension reset_all_data

vitaly-krugl:
CC @scottpurdy, @oxtopus 

ENG-98 Implemented Taurus refresh CI pipeline's reset of Taurus Engine's message bus resources, repository, and model checkpoints via distutils command extension reset_all_data.

ENG-98 Remove unnecessary db migration from refresh pipeline, since taurus-reset-collectorsdb already performs all the necessary migrations.

ENG-98 Removed obsolete script pipeline/scripts/build-taurus.


taurus/pipeline/scripts/refresh-remote-taurus-servers.sh
```diff
@@ -375,8 +367,6 @@ pushd "${REPOPATH}"
         --host=${RABBITMQ_HOST} \
         --user=${RABBITMQ_USER} \
         --password=${RABBITMQ_PASSWD} &&
-     cd /opt/numenta/products/taurus.metric_collectors/taurus/metric_collectors/collectorsdb &&
-     python migrate.py &&
```
vitaly-krugl: @oxtopus: `python setup.py reset_all_data` performs migration as part of one of its operations. Do you think that we need an explicit `python migrate.py` here as well? Similarly in `taurus.metric_collectors` update.

taurus/setup.py
```diff
@@ -1,9 +1,55 @@
...
+      "Invalid suppress_prompt_and_obliterate: "
+      "expected 0 or 1, but got {}".format(self.suppress_prompt_and_obliterate))
+
+
+  def run(self):
+    # NOTE: we execute this via subprocess in order to enable certain setup
+    # options (e.g., --help-commands) to be executed without dependency on
+    # certain environment variables (e.g., `APPLICATION_CONFIG_PATH`).
+    cmd = [
+      "python",
```
oxtopus: There have been issues in the past in which the arbitrarily defined "python" command is used to execute a command with unexpected results.  It's better to ensure that the same "python" command that invoked setup.py is invoked here -- i.e. use `sys.executable` in place of "python".

vitaly-krugl:
@oxtopus, are you done with the review of this PR?

taurus/setup.py
vitaly-krugl: Good point, I'll fix it - thx

vitaly-krugl: Thanks, @oxtopus. I fixed this via PR #541

PR #531, opened 2015-12-17
ENG-84 Fix race condition in taurus update pipeline's server stop/start logic

vitaly-krugl:
CC @scottpurdy, @oxtopus

ENG-84 Cleaned up server stop/start logic in update-remote-taurus-servers.sh.

ENG-84 Added necessary nta-wait-for-supervisord-stopped to eliminate stop/start race condition.

ENG-84 In nta.utils.tools.supervisord_state.waitForStoppedStateMain, wait only for socket.error to conclude that supervisord is not available in order to eliminate race condition with re-starting logic.
ENG-84 Removed unnecessary supervisord shutdown of metric collector before running resource accessibility test, since the server was already shut down at the top of the script.

### DIFF
--- a/taurus/pipeline/scripts/refresh-remote-taurus-servers.sh
+++ b/taurus/pipeline/scripts/refresh-remote-taurus-servers.sh
@@ -214,7 +214,7 @@ pushd "${REPOPATH}"
 
   fi
 
-  # Shutdown services
+  # Stop taurus metric collector server
   ssh -v -t ${SSH_ARGS} "${TAURUS_COLLECTOR_USER}"@"${TAURUS_COLLECTOR_HOST}" \
     "cd /opt/numenta/products/taurus.metric_collectors &&
      if [ -f supervisord.pid ]; then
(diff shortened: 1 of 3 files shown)


taurus/pipeline/scripts/refresh-remote-taurus-servers.sh
```diff
@@ -368,8 +369,6 @@ pushd "${REPOPATH}"
         --user=${RABBITMQ_USER} \
         --password=${RABBITMQ_PASSWD} &&
      cd /opt/numenta/products/taurus.metric_collectors &&
-     supervisorctl --serverurl http://localhost:8001 shutdown &&
```
vitaly-krugl: Already shut it down at the beginning of the script, so no need to shut it down again.

taurus/pipeline/scripts/update-remote-taurus-servers.sh
```diff
@@ -339,16 +338,9 @@ pushd "${REPOPATH}"
      cd /opt/numenta/products/taurus/taurus/engine/repository &&
      python migrate.py &&
      cd /opt/numenta/products/taurus &&
-     if [ -f /var/run/nginx.pid ]; then
-       sudo /usr/sbin/nginx -p . -c conf/nginx-taurus.conf -s reload
```
vitaly-krugl: Don't want the reloads, since the script shuts down these servers at the beginning

scottpurdy:
:+1:

vitaly-krugl:
Thx!

PR #540, opened 2015-12-20
Re-enable load/save

oxtopus:
As it turns out, the problem seemed to be the placement of `model.numLabels = 0`

This appears to be working now.


BoltzmannBrain:
:+1:

PR #544, opened 2015-12-26
Initialize NLP model after attempt at loading

BoltzmannBrain:
When creating a model, the runner would first initialize a model instance and then attempt loading a saved model. When loading a saved model, this initialization step is unnecessary. To avoid this I've added a `loadModel()` method and reordered the logic.

@oxtopus please review. I ran into issues adding `cPickle` to the requirements.txt.

### DIFF
--- a/imbu/engine/fluent_api.py
+++ b/imbu/engine/fluent_api.py
@@ -354,8 +354,6 @@ def createModel(modelName, modelFactory):
 
 
 
-
-
 @postfork
 @thread
 def setupModelWorkers():
--- a/imbu/engine/fluent_api.py
+++ b/imbu/engine/fluent_api.py
@@ -26,7 +26,7 @@ import logging
 from multiprocessing import JoinableQueue, Process
 import os
 import pkg_resources
-import cPickle as pkl
+import cPickle as pickle
 from threading import Thread
 import traceback
 import web
@@ -111,7 +111,7 @@ class ModelProcess(Process):
         elif isinstance(obj, SaveModelTask):
           output = self.modelObj.saveModel()
         elif isinstance(obj, LoadModelTask):
-          output = loadModel(obj.modelDir)  # TODO: Fix this modelDir scope mess!
+          output = loadModel(obj.modelDir)
           self.modelObj = output
 
         # Blocking put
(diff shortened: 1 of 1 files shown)


imbu/engine/fluent_api.py
```diff
@@ -110,7 +111,7 @@ def run(self):
         elif isinstance(obj, SaveModelTask):
           output = self.modelObj.saveModel()
         elif isinstance(obj, LoadModelTask):
-          output = self.modelObj.loadModel(modelDir=obj.modelDir) # TODO: Fix this modelDir scope mess!
+          output = loadModel(obj.modelDir)  # TODO: Fix this modelDir scope mess!
```
oxtopus: I think this pr addresses the scope issue I'm referring to in the to-do comment.  I think we can remove the comment.

imbu/engine/fluent_api.py
```diff
@@ -26,6 +26,7 @@
 from multiprocessing import JoinableQueue, Process
 import os
 import pkg_resources
+import cPickle as pkl
```
oxtopus: import cPickle as `pickle`, rather than `pkl`.  Otherwise, not necessary to rename.

oxtopus:
Re: cPickle and requirements.txt, it's not necessary.  cPickle is built-in.

oxtopus:
Lgtm aside from some minor feedback.  Thanks!

oxtopus:
Upon further inspection, this isn't working as expected.  `loadModel()` added here is not a complete translation of https://github.com/numenta/nupic.research/blob/d8a9992dcd23b14c58e23873c1e3419c6d25e27e/htmresearch/frameworks/nlp/classification_model.py#L113-L124

The new `loadModel()` passes the directory verbatim, whereas the old one appended "model.pkl".  As a result, the new implementation always fails, causing the models to retrain rather than load.

I'm going to follow up with a fix, but the more I think about it, the less I like the new `loadModel()` function external to the class implementation added here.  The implementation details need to be abstracted in the class definition, even if it's just a static or class method.

oxtopus:
See https://github.com/numenta/numenta-apps/pull/547 for solution

PR #551, opened 2016-01-06
TAUR-1792: Account for intervals at the end of the year where fromDay could be greater than toDay

lscheinkman:


lscheinkman:
CC: @oxtopus

PR #554, opened 2016-01-07
Imbu runs NLP models with new token-level scheme

BoltzmannBrain:
* Models are trained by feeding them one token at a time.
* Also removed the CioWindows model for now; it'll likely be replaced with a model that uses NuPIC's Network API.

@oxtopus please review.

### DIFF
--- a/imbu/engine/fluent_api.py
+++ b/imbu/engine/fluent_api.py
@@ -39,9 +39,10 @@ from htmresearch.frameworks.nlp.classify_fingerprint import (
 # from htmresearch.frameworks.nlp.classify_htm import ClassificationModelHTM
 from htmresearch.frameworks.nlp.classify_keywords import (
   ClassificationModelKeywords)
-from htmresearch.frameworks.nlp.classify_windows import (
-  ClassificationModelWindows)
+#from htmresearch.frameworks.nlp.classify_windows import (
+#  ClassificationModelWindows)
 from htmresearch.support.csv_helper import readCSV
+from htmresearch.support.text_preprocess import TextPreprocess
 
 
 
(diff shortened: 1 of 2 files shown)


imbu/engine/fluent_api.py
```diff
@@ -332,6 +340,23 @@ def createModel(modelName, modelFactory):
   g_models[modelName] = modelProxy
 
 
+def trainModelWithText(model, trainingData):
+  """ Train the given model on trainingData.
+  This is (essentially) the same training method as in the research repo's
+  imbu_runner.py.
+  """
+  TP = TextPreprocess()
```
oxtopus: Per naming convention, the capitalized `TP` implies that it's a class.  Meanwhile, `TP` is commonly used elsewhere to refer to "Temporal Pooler".  Consider naming to something more descriptive, say, `testPreprocessor`.

imbu/engine/fluent_api.py
```diff
@@ -332,6 +340,23 @@ def createModel(modelName, modelFactory):
...
+  for seqId, (text, _, _) in enumerate(trainingData.values()):
+    textTokens = TP.tokenize(text)  # TODO: use model's tokenization method instead
+    lastToken = len(textTokens) - 1
+    for i, token in enumerate(textTokens):
+      # use the sequence's ID as the category label
+      model.trainText(token,
+                      [seqId],
+                      sequenceId=seqId,
+                      reset=int(i==lastToken))
```
oxtopus: Why is this param an `int` and not `bool`?

BoltzmannBrain: The NLP models expect `int` resets b/c this is how resets are flagged in NuPIC's network API.

imbu/engine/fluent_api.py
BoltzmannBrain: yes thank you

PR #561, opened 2016-01-18
Catch up

BoltzmannBrain:
@oxtopus @lscheinkman 
For now, require that we load models (from engine/<model name>). These changes (besides those in the UI) are temporary fixes to catch up to the nupic.research refactoring. This will be replaced by [IMBU-70](https://jira.numenta.com/browse/IMBU-70).


oxtopus:
:+1:

PR #566, opened 2016-01-20
Switch haigha and pytz to new releases

vitaly-krugl:
CC @oxtopus @scottpurdy 
Switch haigha amqp client library source from vitaly's private branch to official 0.9.0 release, now that vitaly's haigha PRs are in the official release.

Update pytz version number to the current latest 2015.7 (from 2015.2)


scottpurdy:
:+1:

PR #571, opened 2016-01-23
[UNI-299] Implementation of paramFinder and paramFinder runner

ywcui1990:
@marionleborgne @vitaly-krugl 

A first version of paramFinder class (for future nupic use) and the paramFinder runner (for unicorn)

I am still working on the tests. But any early feedback on this implementation is appreciated!


marionleborgne:
:+1: I ran `param_finder_runner.py` on the hotgym data and it gave me a windows size of 1.5 hour for aggregation with the "mean" operation. 
Adding a unit test to validate that the param_finder returns the right params will be good indeed. Let me know when you've added that.

unicorn/py/unicorn_backend/param_finder_runner.py
```diff
@@ -0,0 +1,163 @@
...
+               "valueIndex")
+
+
+  def __init__(self,
+               fileName,
+               rowOffset=1,
+               timestampIndex=0,
+               valueIndex=1):
+    """
+    :param str modelId: model identifier
```
marionleborgne: This is the old doc string from `model_runner.py` :smile: Please update the param names and descriptions.

marionleborgne:
@ywcui1990 Below is the JSON returned by the param finder. I have a couple of questions:
1. Why is `predictionSteps` set to 1? We only want to do anomaly detection
2. Why is `predictedField` set to `c1`? I ran the param finder against a CSV where the metric name is `attendeeCount` not `c1`. Are you sure the field name gets updated correctly? I see a couple of `c1` in the JSON returned, instead of the actual metric name. Also, @vitaly-krugl how do we handle the case where a CSV file without headers is passed (in that case, what is the name of the predicted field? is it `c1`?)


```
         {
           "aggInfo":{  
               "windowSize":5270,
               "func":"mean"
            },
            "modelInfo":{  
               "inferenceArgs":{  
                  "predictionSteps":[  
                     1
                  ],
                  "predictedField":"c1",
                  "inputPredictedField":"auto"
               },
               "modelConfig":{  
                  "aggregationInfo":{  
                     "seconds":0,
                     "fields":[  

                     ],
                     "months":0,
                     "days":0,
                     "years":0,
                     "hours":0,
                     "microseconds":0,
                     "weeks":0,
                     "minutes":0,
                     "milliseconds":0
                  },
                  "model":"CLA",
                  "version":1,
                  "predictAheadTime":null,
                  "modelParams":{  
                     "sensorParams":{  
                        "verbosity":0,
                        "encoders":{  
                           "c0_dayOfWeek":{  
                              "dayOfWeek":[  
                                 21,
                                 3
                              ],
                              "type":"DateEncoder",
                              "fieldname":"c0",
                              "name":"c0"
                           },
                           "c0_timeOfDay":null,
                           "c1":{  
                              "name":"c1",
                              "resolution":3.2857343580170375,
                              "seed":42,
                              "fieldname":"c1",
                              "type":"RandomDistributedScalarEncoder"
                           },
                           "c0_weekend":null
                        },
                        "sensorAutoReset":null
                     },
                     "clEnable":false,
                     "spParams":{  
                        "columnCount":2048,
                        "spVerbosity":0,
                        "maxBoost":1.0,
                        "spatialImp":"cpp",
                        "inputWidth":0,
                        "synPermInactiveDec":0.0005,
                        "synPermConnected":0.1,
                        "synPermActiveInc":0.0015,
                        "seed":1956,
                        "numActiveColumnsPerInhArea":40,
                        "globalInhibition":1,
                        "potentialPct":0.8
                     },
                     "trainSPNetOnlyIfRequested":false,
                     "clParams":{  
                        "alpha":0.035828933612158,
                        "clVerbosity":0,
                        "steps":"1",
                        "regionName":"CLAClassifierRegion"
                     },
                     "tpParams":{  
                        "columnCount":2048,
                        "activationThreshold":13,
                        "pamLength":3,
                        "cellsPerColumn":32,
                        "permanenceInc":0.1,
                        "minThreshold":10,
                        "verbosity":0,
                        "maxSynapsesPerSegment":32,
                        "outputType":"normal",
                        "globalDecay":0.0,
                        "initialPerm":0.21,
                        "permanenceDec":0.1,
                        "seed":1960,
                        "maxAge":0,
                        "newSynapseCount":20,
                        "maxSegmentsPerCell":128,
                        "temporalImp":"cpp",
                        "inputWidth":2048
                     },
                     "anomalyParams":{  
                        "anomalyCacheRecords":null,
                        "autoDetectThreshold":null,
                        "autoDetectWaitRecords":5030
                     },
                     "spEnable":true,
                     "inferenceType":"TemporalAnomaly",
                     "tpEnable":true
                  }
               }
            }
         }
```

marionleborgne:
@vitaly-krugl based on our last discussion about the timestamp format, shouldn't we have one additional command line arg that allows us to pass the timestamp format string to the `param_finder_runner`?

vitaly-krugl:
@marionleborgne, regarding
> @vitaly-krugl could you remind me how do we handle the case where a CSV file without headers is passed (in that case, what is the name of the predicted field if there is no header row? is it `c1`?)

Actually, "c0" and "c1" (or "timestamp" and "value") are better than arbitrary column names from the csv file, which might contain spaces and other unsavory characters. What we name things on the backend side has no bearing on the computation, and removing user-related nuances for the list of problems to deal with is a good thing.

vitaly-krugl:
@marionleborgne, regarding
> @vitaly-krugl based on our last discussion about the timestamp format, shouldn't we have one additional command line arg that allows us to pass the timestamp format string to the `param_finder_runner`?

Yes, I think so. It's called `datetimeFormat` and is described in the "ParamFinder Command-line Args" section of [Unicorn Front End/Back End Interface (phase 2)](https://docs.google.com/document/d/1XZYGFCIoK7NarAl-DzCqluaTHac1PYFZQaEiu0r2rhs)

vitaly-krugl:
@marionleborgne, good catch regarding `timestampFieldName` and `valueFieldName`. Those need to be implemented per the "ParamFinder Command-line Args" section of [Unicorn Front End/Back End Interface (phase 2)](https://docs.google.com/document/d/1XZYGFCIoK7NarAl-DzCqluaTHac1PYFZQaEiu0r2rhs)

vitaly-krugl:
@marionleborgne and @ywcui1990, regarding Marion's question
>  Why is predictionSteps set to 1? We only want to do anomaly detection

see https://github.com/numenta/nupic/blob/master/src/nupic/frameworks/opf/common_models/anomaly_params_random_encoder/best_single_metric_anomaly_params.json, which is a template for the configuration parameters returned by `nupic.frameworks.opf.common_models.cluster_params.getScalarMetricWithTimeOfDayAnomalyParams()`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+3. suggest data aggregation type (SUM or AVG)
+"""
+
+import csv
+import dateutil.parser
+import numpy
+from nupic.frameworks.opf.common_models.cluster_params import (
+  getScalarMetricWithTimeOfDayAnomalyParams)
+
+_modeFromNameDict = {
```
vitaly-krugl: If it's intended to be a constant, then make it all-caps, separated by underscores: `_NAME_TO_MODE_MAP`.

If it's intended to be a module-level global variable, then prefix it with `g`: `g_nameToModelMap`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""
+Implements param finder, which automatically
+1. identify appropriate data aggregation window
+2. suggest whether to use TimeOfDay encoder
+3. suggest data aggregation type (SUM or AVG)
```
vitaly-krugl: "sum" or "mean"

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+"""
```
vitaly-krugl: Thank's for putting a description in the docstring. This is helpful.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+
+import csv
+import dateutil.parser
+import numpy
+from nupic.frameworks.opf.common_models.cluster_params import (
```
vitaly-krugl: Need whitespace between groups of imports. 0ne line to separate builtin from external imports in this case

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+  if len(a) == 0:
+    raise ValueError('a cannot be empty')
+  if len(v) == 0:
+    raise ValueError('v cannot be empty')
+  mode = _modeFromName(mode)
+  return numpy.core.multiarray.correlate(a, v[::-1], mode)
+
+
+
+def _modeFromName(mode):
```
vitaly-krugl: This seems like unnecessary complexity. Instead of `_modeFromNameDict`, just define model-level constants, and pass those directly to `_convolve()` without the need to map. E.g.,

```
_CORRELATION_MODE_VALID = 0
_CORRELATION_MODE_SAME = 1
_CORRELATION_MODE_FULL = 2

File: unicorn/py/unicorn_backend/param_finder.py [hunk 7]
```diff
@@ -0,0 +1,536 @@
...
+
+_modeFromNameDict = {
+  'v': 0,
+  's': 1,
+  'f': 2
+}
+
+
+
+def _convolve(a, v, mode='full'):
```
vitaly-krugl: Short names like `a` and `v` are frowned upon in python.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+  models the effect of a linear time-invariant system on a signal [1]_.  In
+  probability theory, the sum of two independent random variables is
+  distributed according to the convolution of their individual
+  distributions.
+
+  If `v` is longer than `a`, the arrays are swapped before computation.
+
+  Parameters
+  ----------
+  :param a : (N,) array_like
```
vitaly-krugl: Should be no space between variable name and `:`

e.g., `:param a:`

vitaly-krugl:
Need to run the two scripts through pylint using pylintrc from `numenta-apps/pylintrc` for numenta-apps code and `nupic/pylintrc` for nupic code and fix findings.

vitaly-krugl:
Need unit tests. See https://github.com/numenta/numenta-apps/tree/master/unicorn/py/unicorn_backend/tests

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+
+  References
+  ----------
+  .. [1] Wikipedia, "Convolution", http://en.wikipedia.org/wiki/Convolution.
+
+  """
+  a, v = numpy.array(a, ndmin=1), numpy.array(v, ndmin=1)
+  if len(v) > len(a):
+    a, v = v, a
+  if len(a) == 0:
```
vitaly-krugl: It's probably easier to understand, if you place the `==` checks before the `>` block, even though the result is the same in this function. The idea is that you want to perform arg validation before the core logic of the function.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+  -------
+  :return out : ndarray
+      Discrete, linear convolution of `a` and `v`.
+
+  References
+  ----------
+  .. [1] Wikipedia, "Convolution", http://en.wikipedia.org/wiki/Convolution.
+
+  """
+  a, v = numpy.array(a, ndmin=1), numpy.array(v, ndmin=1)
```
vitaly-krugl: The code is more readable if you use individual statements on individual lines; there is no need to bundle them like this, making it more difficult to understand. The following is more readable:

```
a = numpy.array(a, ndmin=1)
v = numpy.array(v, ndmin=1)
```

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+  if len(v) > len(a):
+    a, v = v, a
```
vitaly-krugl: Need to use whitespace between code blocks for readability.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+
+
+def _modeFromName(mode):
+  if isinstance(mode, basestring):
+    return _modeFromNameDict[mode.lower()[0]]
+  return mode
+
+
+
+def _rickerWavelet(points, a):
```
vitaly-krugl: Here and elsewhere, the short arg and variable names like `a` are not conventional in python. Please use more descriptive names for code readability. Pylint doesn't like them either.

vitaly-krugl:
Function documentation is great overall, but need to follow conventions, which differ for nupic (Doxygen) and numenta-apps (Sphinx) code.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+
+class paramFinder(object):
+  def __init__(self,
+               fileName,
+               rowOffset=1,
+               timestampIndex=0,
+               valueIndex=1):
+    """
+    :param fileName: str, path to input csv file
+    :param rowOffset: int, index of first data row in csv
```
vitaly-krugl: add "zero-based"

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+    output[ind, :] = _convolve(data, waveletData,
+                               mode='same')
+  return output
+
+
+
+class paramFinder(object):
+  def __init__(self,
+               fileName,
+               rowOffset=1,
```
vitaly-krugl: Let's not have default values for any of these parameters. There are no really good, intuitive defaults for these, so it's better to force users to specify them explicitly. There are several reasons for this.

1. CSV files contains metric data used elsewhere in nupic (e.g., https://github.com/numenta/nupic/blob/master/src/nupic/data/file_record_stream.py) actually expect 3 header rows (not one).
2. There is no convention that timestamp should be in column 0 and value in column 1.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+  output = numpy.zeros([len(widths), len(data)])
+  for ind, width in enumerate(widths):
+    waveletData = wavelet(min(10 * width, len(data)), width)
+    output[ind, :] = _convolve(data, waveletData,
+                               mode='same')
+  return output
+
+
+
+class paramFinder(object):
```
vitaly-krugl: By convention, class names should be mixed case, starting with upper case; so, this should be `ParamFinder`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+               timestampIndex=0,
+               valueIndex=1):
+    """
+    :param fileName: str, path to input csv file
+    :param rowOffset: int, index of first data row in csv
+    :param timestampIndex: int, column index of the timeStamp
+    :param valueIndex: int, column index of the value
+    """
+
+    self._thresh = 0.2
```
vitaly-krugl: Use a more descriptive name and/or add a comment that describes what this threshold applies to.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+          "func": A string representation of the aggregation function (string)
+            the following values are supported: "sum", "mean"
+    """
+    (timeStamp, value) = self.readCSVFiles(self._fileName,
+                                           self._rowOffset,
+                                           self._timestampIndex,
+                                           self._valueIndex)
+    numDataPts = len(value)
+
+    (self.medianSamplingInterval,
```
vitaly-krugl: Here and elsewhere in this PR: define all instance member variables in `__init__` first. If there is no meaningful value for them in `__init__`, then simply assign `None`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+
+    (timeStamp, value) = self.resampleData(timeStamp,
+                                           value,
+                                           self.medianSamplingInterval)
+
+    (self._cwtVar, self._timeScale) = self.calculateContinuousWaveletTransform(
+      self.medianSamplingInterval, value)
+
+    self.suggestedSamplingInterval = self.determineAggregationWindow(
+      self._timeScale,
```
vitaly-krugl: When more than 1-2 args, pass then as keyword args. Error-prone and often not readable otherwise.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+                                type='DateEncoder',
+                                dayOfWeek=(21, 3))
+    else:
+      modelParams['modelConfig']['modelParams']['sensorParams']['encoders'] \
+        ['c0_dayOfWeek'] = None
+    return modelParams
+
+
+  @staticmethod
+  def readCSVFiles(fileName,
```
vitaly-krugl: readCSVFiles ==> readCSVFile (only reads one file)

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+    self._rowOffset = rowOffset
+    self._timestampIndex = timestampIndex
+    self._valueIndex = valueIndex
+
+
+  def run(self):
+    """
+    Run paramFinder on input csv file
+
+    :return: JSON object with the following properties
```
vitaly-krugl: return doc is incomplete, missing modelInfo

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+    """
+
+    self._thresh = 0.2
+    self._fileName = fileName
+    self._rowOffset = rowOffset
+    self._timestampIndex = timestampIndex
+    self._valueIndex = valueIndex
+
+
+  def run(self):
```
vitaly-krugl: Based on what this does, I think it might be better off as a standalone function, and no class is needed. If I understand it correctly, it would be simpler for the user to just call a single function instead of constructing an instance of `ParamFinder` and calling `run`; e.g., `def findModelParams(fileName, rowOffset, timestampIndex, valueIndex)`). You can still subdivide the work among private functions and pass the necessary parameters by value.

This implementation is probably better served by functional programming than by OO programming; for example, `getAggregationFunction ()` depends on `self.medianAbsoluteDevSamplingInterval` and other instance members being set up by something else, so it's confusing to follow the code, and not clear in what order functions can legally be called.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+
+        if (timeScale[leftLocalMin] < weekPeriod < timeScale[
+          rightLocalMin]
+            and cwtVarAtWeekPeriod > localMaxValue * 0.5):
+          useDayOfWeek = True
+
+    return useTimeOfDay, useDayOfWeek
+
+
+  def getAggregationFunction(self):
```
vitaly-krugl: This and other private methods in this class need to use the naming convention for private functions, using a leading underscore. It would be inappropriate to call this as a public API, because it depends on `run()` to set up `self.medianAbsoluteDevSamplingInterval` and other instance variables.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+            the following values are supported: "sum", "mean"
+    """
+    (timeStamp, value) = self.readCSVFiles(self._fileName,
+                                           self._rowOffset,
+                                           self._timestampIndex,
+                                           self._valueIndex)
+    numDataPts = len(value)
+
+    (self.medianSamplingInterval,
+     self.medianAbsoluteDevSamplingInterval) = self.getMedianSamplingInterval(
```
vitaly-krugl: It seems that `self.medianAbsoluteDevSamplingInterval` and other instance members here should use the naming convention for private members (leading underscrore)

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+    """
+    Read csv data file, the data file must have two columns
+    that contains time stamps and data values
+
+    :param fileName: str, path to input csv file
+    :param rowOffset: int, index of first data row in csv
+    :param timestampIndex: int, column index of the timeStamp
+    :param valueIndex: int, column index of the value
+    :return: timeStamps: numpy array of time stamps
+             values: numpy array of data values
```
vitaly-krugl: The return value needs to make it clear that a two-tuple is returned; for example: 

A two-tuple (timestamps, values), where timestamps is a numpy array of `datetime64` time stamps, and values is a numpy array of `float32` data values as

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+        timeStamp = dateutil.parser.parse(row[timestampIndex])
+        timeStamps.append(numpy.datetime64(timeStamp))
+        values.append(row[valueIndex])
+
+        numRow += 1
+        if numRow >= maxRowNumber:
+          break
+
+      timeStamps = numpy.array(timeStamps, dtype='datetime64[s]')
+      values = numpy.array(values, dtype='float32')
```
vitaly-krugl: Is float32 sufficient? This needs to be consistent with the maximum value size supported by nupic models for scalar values in metric data.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+
+    with open(fileName, 'r') as csvFile:
+      fileReader = csv.reader(csvFile)
+      for _ in xrange(rowOffset):
+        fileReader.next()  # skip header line
+
+      timeStamps = []
+      values = []
+
+      maxRowNumber = 20000
```
vitaly-krugl: `maxRowNumber` should instead be defined as a class-level or module-level constant. Among other things, this would make it possible to patch in unit tests.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+    waveletData = wavelet(min(10 * width, len(data)), width)
+    output[ind, :] = _convolve(data, waveletData,
+                               mode='same')
+  return output
+
+
+
+class paramFinder(object):
+  def __init__(self,
+               fileName,
```
vitaly-krugl: Expecting a CSV file in an API like this limits its flexibility. This means that anyone who needs to use this functionality needs to create a CSV file first in order to use this functionality. If you anticipate that this API would not need to scan an entire file (regardless of size), then it's more flexible to expose a constant that lets the users know the desired numbers of rows, and then accept the info as a sequence of (timestamp, value) two-tuples, and do away with all the other args here.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+    Run paramFinder on input csv file
+
+    :return: JSON object with the following properties
+      "aggInfo" aggregation information, JSON null if no aggregation is needed
+        otherwise, a JSON object contains the folowing properties
+          "windowSize": aggregation window size in seconds (integer)
+          "func": A string representation of the aggregation function (string)
+            the following values are supported: "sum", "mean"
+    """
+    (timeStamp, value) = self.readCSVFiles(self._fileName,
```
vitaly-krugl: These vars should be plural `timeStamps` and `values`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+          break
+
+      timeStamps = numpy.array(timeStamps, dtype='datetime64[s]')
+      values = numpy.array(values, dtype='float32')
+
+      return timeStamps, values
+
+
+  @staticmethod
+  def resampleData(timeStamp, value, newSamplingInterval):
```
vitaly-krugl: Based on use, `totalDuration = (timeStamp[-1] - timeStamp[0])`, it appears that `timeStamp` is actually an array of timestamps, so need a more appropriate name for the `timeStamp` arg and also update param documentation.

Same for `value` ==> `values` and update param documentation.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+      newTimestamp[sampleI] = timeStamp[0] + sampleI * newSamplingInterval
+
+    newValue = numpy.interp((newTimestamp - timeStamp[0]).astype('float32'),
+                            (timeStamp - timeStamp[0]).astype('float32'), value)
+
+    return newTimestamp, newValue
+
+
+  @staticmethod
+  def calculateContinuousWaveletTransform(samplingInterval, value):
```
vitaly-krugl: `value` ==> `values` and update docstring

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+
+    # variance of wavelet power
+    cwtVar = numpy.var(numpy.abs(cwtmatr), axis=1)
+    cwtVar = cwtVar / numpy.sum(cwtVar)
+
+    return cwtVar, timeScale
+
+
+  @staticmethod
+  def getMedianSamplingInterval(timeStamp):
```
vitaly-krugl: `timeStamp` ==> `timeStamps` and update docstring

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+             medianAbsoluteDev, median absolute deviation of sampling interval
+    """
+    if timeStamp.dtype != numpy.dtype('<M8[s]'):
+      timeStamp = timeStamp.astype('datetime64[s]')
+
+    samplingIntervals = numpy.diff(timeStamp)
+
+    medianSamplingInterval = numpy.median(samplingIntervals)
+    medianAbsoluteDev = numpy.median(numpy.abs(
+      samplingIntervals - medianSamplingInterval))
```
vitaly-krugl: pylint probably won't like this type of hanging indentation. The following would be better:

```
medianAbsoluteDev = numpy.median(
  numpy.abs(samplingIntervals - medianSamplingInterval))
```
```

File: unicorn/py/unicorn_backend/param_finder.py [hunk 33]
```diff
@@ -0,0 +1,536 @@
...
+    return cwtVar, timeScale
+
+
+  @staticmethod
+  def getMedianSamplingInterval(timeStamp):
+    """
+    calculate median and median absolute deviation of sampling interval
+
+    :param timeStamp: a numpy array of sampling times
+    :return: medianSamplingInterval, numpy timedelta64 format in unit of seconds
```
vitaly-krugl: Start by saying that it's a two-tuple

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+    """
+
+    widths = numpy.logspace(0, numpy.log10(len(value) / 20), 50)
+    timeScale = widths * samplingInterval * 4
+
+    # continuous wavelet transformation with ricker wavelet
+    cwtmatr = _cwt(value, _rickerWavelet, widths)
+    # clip wavelet coefficients to minimize boundary effect
+    maxTimeScale = int(widths[-1])
+    cwtmatr = cwtmatr[:, 4 * maxTimeScale:-4 * maxTimeScale]
```
vitaly-krugl: What does the comma mean in `cwtmatr[:, `?

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+    :param value: value of the time series
+    :return cwtVar: numpy array of flaots, variance of the wavelet coeff
+    :return timeScale: numpy array of floats, corresponding time scales
+    """
+
+    widths = numpy.logspace(0, numpy.log10(len(value) / 20), 50)
+    timeScale = widths * samplingInterval * 4
+
+    # continuous wavelet transformation with ricker wavelet
+    cwtmatr = _cwt(value, _rickerWavelet, widths)
```
vitaly-krugl: "matr" may be a non-standard abbreviation for matrix. It's not desirable to use non-standard abbreviations. `cwtMatrix` would be better.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+  @staticmethod
+  def calculateContinuousWaveletTransform(samplingInterval, value):
+    """
+    Calculate continuous wavelet transformation (CWT)
+    Return variance of the cwt coefficients over time
+
+    :param samplingInterval: sampling interval of the time series
+    :param value: value of the time series
+    :return cwtVar: numpy array of flaots, variance of the wavelet coeff
+    :return timeScale: numpy array of floats, corresponding time scales
```
vitaly-krugl: This is not a standard way to document a two-tuple return value. There is a single object being returned that is a two-tuple.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+    """
+    cumulativeCwtVar = numpy.cumsum(cwtVar)
+    cutoffTimeScale = timeScale[numpy.where(cumulativeCwtVar >= thresh)[0][0]]
+
+    aggregationTimeScale = cutoffTimeScale / 10.0
+    if aggregationTimeScale < samplingInterval * 4:
+      aggregationTimeScale = samplingInterval * 4
+
+    # make sure there is at least 1000 records after aggregation
+    if numDataPts < 1000:
```
vitaly-krugl: This `1000` should probably be a class-level or module-level constant.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+    (2) Above the baseline value
+  
+    The algorithm will suggest an encoder if its corresponding
+    periodicity is close to a strong maxima:
+    (1) horizontally must within the nearest local minimum
+    (2) vertically must within 50% of the peak of the strong maxima
+
+    :param cwtVar: numpy array, wavelet coefficients variance over time
+    :param timeScale: numpy array, corresponding time scales for wavelet coeffs
+    """
```
vitaly-krugl: Document return value

unicorn/py/unicorn_backend/param_finder_runner.py
```diff
@@ -0,0 +1,163 @@
...
+
+  parser.add_argument("--csv",
+                      type=str,
+                      dest="csv",
+                      required=True,
+                      help="REQUIRED: path to input CSV file")
+
+  parser.add_argument("--rowOffset",
+                      type=int,
+                      default=1,
```
vitaly-krugl: As explained in the review of `param_finder.py`, there are not really any good, intuitive defaults for rowOffset, timestampIndex, and valueIndex. It's better to make them required and not specify defaults.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,536 @@
...
+
+    :param fileName: str, path to input csv file
+    :param rowOffset: int, index of first data row in csv
+    :param timestampIndex: int, column index of the timeStamp
+    :param valueIndex: int, column index of the value
+    :return: timeStamps: numpy array of time stamps
+             values: numpy array of data values
+    """
+
+    with open(fileName, 'r') as csvFile:
```
vitaly-krugl: The opening of the file and creation of the csv reader need a few more nuances. For creation of csv reader, see https://github.com/numenta/numenta-apps/blob/64fcd471fb79b97909537179528734256fcb7333/unicorn/py/unicorn_backend/model_runner_2.py#L250-L257.

For opening of the file, you need to specify "universal end-of-line" to work with Mac OS X, Windows (and linux) CSV files. Use "rU" as the open mode (instead of 'r').

vitaly-krugl:
Hi @ywcui1990, I am done with this round of code review. I focused primarily on code structure and coding/naming conventions. I am not familiar with the algorithms in question, so I didn't review them for correctness.

The main structural feedback for `param_finder.py` is:
1. The public interface to param finder should be a function rather than class. See https://github.com/numenta/numenta-apps/pull/571#discussion_r51220049 for explanation.
2. Accepting a CSV file as input is not desirable; see https://github.com/numenta/numenta-apps/pull/571/files#r51221239.

ywcui1990:
@vitaly-krugl  Thanks very much for the detailed code review. I will take a thorough look at your comments tomorrow

ywcui1990:
@marionleborgne @vitaly-krugl 

I have incorporated Vitaly's comments and added some unit tests. 

Summary of major changes
* Make param_finder a set of functions, instead of a class
* param_finder now takes timeStamps and values as input, instead of the csv file. The readCSVFile function has been moved to param_finder_runner
* No default values are given for any of the input parameters for param_finder_runner, the user has to provide all of them
* Add "timestampFieldName" and "valueFieldName" to output of param_finder, which are now hard coded as "c0" and "c1" respectively

Example usage:
python param_finder_runner.py --csv example_data/art_daily_no_noise.csv --rowOffset 1 --timestampIndex 0 --valueIndex 1 --datetimeFormat "%Y-%m-%d %H:%M:%S"

python param_finder_runner.py --csv [PATH_TO_NUPIC]/examples/opf/clients/hotgym/anomaly/one_gym/rec-center-hourly.csv --rowOffset 3 --timestampIndex 0 --valueIndex 1 --datetimeFormat "%m/%d/%y %H:%M"

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,529 @@
...
+    output[ind, :] = _convolve(data, waveletData, mode=_CORRELATION_MODE_SAME)
+  return output
+
+
+
+def find_parameters(timeStamps, values):
+  """
+  find parameters for a given time series dataset with heuristics
+
+  :param timeStamps: array_like, each entry is a timeStamp in datetime64 type
```
vitaly-krugl: `datetime64`: I think that this is a numpy-specific data type. I don't think that we should expose our internal implementation details to users of the API. I think it would be more standard to take the builtin datetime.datetime type. The function would need to convert to whatever implementation-dependent type it needs.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,529 @@
...
+
+from nupic.frameworks.opf.common_models.cluster_params import (
+  getScalarMetricWithTimeOfDayAnomalyParams)
+
+_CORRELATION_MODE_VALID = 0
+_CORRELATION_MODE_SAME = 1
+_CORRELATION_MODE_FULL = 2
+
+_AGGREGATION_WINDOW_THRESH = 0.2
+_MAX_ROW_NUMBER = 20000
```
vitaly-krugl: If this is a suggested number of rows that users should pass, then this constant should be made public (remove leading underscore) and `find_parameters` docstring should refer to the constant in the explanation of using the API.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,529 @@
...
+  aggFunc = getAggregationFunction(medianSamplingInterval,
+                                   medianAbsoluteDevSamplingInterval)
+
+  outputInfo = {
+    "aggInfo": getAggInfo(medianSamplingInterval,
+                          suggestedSamplingInterval,
+                          aggFunc),
+    "modelInfo": getModelParams(useTimeOfDay, useDayOfWeek, values),
+    "timestampFieldName": "c0",
+    "valueFieldName": "c1"
```
lscheinkman: @ywcui1990 `timestampFieldName` and `valueFieldName` should be properties of `modelInfo`, see https://github.com/numenta/numenta-apps/blob/master/unicorn/py/unicorn_backend/model_opt_schema.json

vitaly-krugl:
@ywcui1990, I started the review last night, but have not finished it. Please don't push changes to this PR until I let you know that I am done with the review. Thx.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+  computation.
+
+  Parameters
+  ----------
+  :param vector1: (N,) array_like
+      First one-dimensional input array.
+  :param vector2: (M,) array_like
+      Second one-dimensional input array.
+  :param mode: int that indicates mode of convolution, which takes the value of
+      2:
```
vitaly-krugl: Use `_CORRELATION_MODE_FULL`, etc. instead of 2, etc. in documentation.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+
+_CORRELATION_MODE_VALID = 0
+_CORRELATION_MODE_SAME = 1
+_CORRELATION_MODE_FULL = 2
+
+_AGGREGATION_WINDOW_THRESH = 0.2
+MAX_ROW_NUMBER = 20000
+
+
+def _convolve(vector1, vector2, mode=2):
```
vitaly-krugl: Use `_CORRELATION_MODE_FULL` instead of `2` in mode default value

vitaly-krugl: Is `_CORRELATION_MODE_FULL` an intuitive default value for `_convolve`? In our own code, we are calling it with `_CORRELATION_MODE_SAME`. Since it's a private method, and we're actually calling it with an explicit `mode` that differs from default, let's get rid of the default value altogether.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+
+  It models the function:
+
+      ``A (1 - x^2/a^2) exp(-t^2/a^2)``,
+
+  where ``A = 2/sqrt(3a)pi^1/3``, a  is the width parameter of the wavelet
+
+  Parameters
+  ----------
+  :param points: int
```
vitaly-krugl: Using Sphinx param documentation, this would be `:param int points:` for types like int or datetime.datetime. If the type is more complex, such that it requires multiple words with spaces, then you would use the following two-line syntax:

```
:param dates: metric dates
:type dates: sequence of datetime.datetime objects
```

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+  Continuous wavelet transform.
+
+  Performs a continuous wavelet transform on `data`,
+  using the `wavelet` function. A CWT performs a convolution
+  with `data` using the `wavelet` function, which is characterized
+  by a width parameter and length parameter.
+
+  Parameters
+  ----------
+  :param data : (N,) ndarray
```
vitaly-krugl: In Sphinx notation, there is no space between parameter name and `:`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+      Will be centered around 0.
+  :param a: scalar Width parameter of the wavelet.
+
+  Returns
+  -------
+  :return vector: (N,) ndarray
+      Array of length `points` in shape of ricker curve.
+
+  """
+  A = 2 / (numpy.sqrt(3 * waveletWidth) * (numpy.pi ** 0.25))
```
vitaly-krugl: Per convention, variable names must start with lower-case letter and should be meaningful names. Single-letter names are not allowed.

Please run all code in this script through pylint using nupic/pylintrc configuration and see what else it finds and fix it. I am suggesting nupic/pylintrc for this script, because my understanding is that it will be migrated to nupic. If that's not the case, then use numenta-apps/pylintrc instead.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+      Wavelet function, which should take 2 arguments.
+      The first argument is the number of points that the returned vector
+      will have (len(wavelet(width,length)) == length).
+      The second is a width parameter, defining the size of the wavelet
+      (e.g. standard deviation of a gaussian). See `ricker`, which
+      satisfies these requirements.
+  widths : (M,) sequence
+      Widths to use for transform.
+
+  :return Returns
```
vitaly-krugl: `:return Returns` is neither Sphinx nor Doxygen

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+    output[ind, :] = _convolve(data, waveletData, mode=_CORRELATION_MODE_SAME)
+  return output
+
+
+
+def find_parameters(timeStamps, values):
+  """
+  find parameters for a given time series dataset with heuristics
+
+  :param timeStamps: array_like, each entry is a timeStamp in datetime type
```
vitaly-krugl: What is `array_like`? Can they pass a python list or tuple? In that case, you probably want to say:

```
:param timeStamps: metric data timestamps
:type timeStamps: sequence of datetime.datetime timestamps
``

File: unicorn/py/unicorn_backend/param_finder.py [hunk 51]
```diff
@@ -0,0 +1,537 @@
...
+  """
+  output = numpy.zeros([len(widths), len(data)])
+  for ind, width in enumerate(widths):
+    waveletData = wavelet(min(10 * width, len(data)), width)
+    output[ind, :] = _convolve(data, waveletData, mode=_CORRELATION_MODE_SAME)
+  return output
+
+
+
+def find_parameters(timeStamps, values):
```
vitaly-krugl: I think it would be cleaner and less error-prone for the API to accept a single sequence of timestamp/value two-tuples. Then there wouldn't be concerns that one might be longer than the other, etc. It would also be easier for users to use Python List Comprehension to build the parameter instead of having to use a for-loop.

vitaly-krugl: Do we need to be concerned about noisy data, where some values might be missing (e.g., None) or `NaN`?

CC @scottpurdy @marionleborgne

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+      corresponding to the metric timestamp (string)
+
+      “valueFieldName”: The name of the field in `modelConfig`
+      corresponding to the metric value (string)
+  """
+  if len(timeStamps) != len(values):
+    raise ValueError('timeStamps and Values must have the same length')
+
+  try:
+    assert(type(timeStamps[0]) is datetime.datetime)
```
vitaly-krugl: Instead of assert, you probably want to do something similar to `ValueError` above, except use the `TypeError` exception.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+    if len(timeStamps) > MAX_ROW_NUMBER:
+      timeStamps = timeStamps[:MAX_ROW_NUMBER]
+
+    timeStamps = numpy.array(timeStamps, dtype='datetime64[s]')
+  except:
```
vitaly-krugl: This is too broad of a catch. Pylint wouldn't like it either. It will masquerade other problems/bugs as `TypeError`, making it confusing to debug. You want to catch more specific exceptions classes.

Also, you have at least two statements in the block that would cause the `TypeError` exception to be raised below. This would make it difficult for the user to debug whether the problem was in the first one `assert(type(timeStamps[0]) is datetime.datetime)` or the second one `numpy.array(timeStamps, dtype='datetime64[s]')`.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+    raise TypeError('timeStamps must be an array of datetime64')
+
+  try:
+    if len(values) > MAX_ROW_NUMBER:
+      values = values[:MAX_ROW_NUMBER]
+
+    values = numpy.array(values).astype('float64')
+  except:
```
vitaly-krugl: See prior feedback about "too broad of a catch".

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+    assert(type(timeStamps[0]) is datetime.datetime)
+    if len(timeStamps) > MAX_ROW_NUMBER:
+      timeStamps = timeStamps[:MAX_ROW_NUMBER]
+
+    timeStamps = numpy.array(timeStamps, dtype='datetime64[s]')
+  except:
+    raise TypeError('timeStamps must be an array of datetime64')
+
+  try:
+    if len(values) > MAX_ROW_NUMBER:
```
vitaly-krugl: This doesn't need to be in the try-except block.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+    "aggInfo": getAggInfo(medianSamplingInterval,
+                          suggestedSamplingInterval,
+                          aggFunc),
+    "modelInfo": getModelParams(useTimeOfDay, useDayOfWeek, values),
+  }
+  return outputInfo
+
+
+
+def getAggInfo(medianSamplingInterval, suggestedSamplingInterval, aggFunc):
```
vitaly-krugl: Is this and the other functions in this script, except `find_parameters`, supposed to be public? If not, need to add leading underscore to mark them as private.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+  else:
+    aggInfo = {
+      "windowSize": suggestedSamplingInterval.astype('int'),
+      "func": aggFunc
+    }
+  return aggInfo
+
+
+
+def getModelParams(useTimeOfDay, useDayOfWeek, value):
```
vitaly-krugl: Rename `value` as `values`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+  """
+  Return a JSON object describing the model configuration
+  :param useTimeOfDay: bool, whether to use timeOfDay encoder
+  :param useDayOfWeek: bool, whether to use dayOfWeej encoder
+  :param value: numpy array of metric data, used to compute min/max values
```
vitaly-krugl: In nupic's developer's mind, metric data usually includes a timestamp and value. So the use of "metric data" in this param description is confusing. Please use a more specific/accurate description.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+      "func": aggFunc
+    }
+  return aggInfo
+
+
+
+def getModelParams(useTimeOfDay, useDayOfWeek, value):
+  """
+  Return a JSON object describing the model configuration
+  :param useTimeOfDay: bool, whether to use timeOfDay encoder
```
vitaly-krugl: `:param bool useTimeOfDay: whether to use timeOfDay encoder`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+
+
+def find_parameters(timeStamps, values):
+  """
+  find parameters for a given time series dataset with heuristics
+
+  :param timeStamps: array_like, each entry is a timeStamp in datetime type
+  :param values: array_like, each entry is a float number
+
+  :return: JSON object with the following properties:
```
vitaly-krugl: In Sphinx, this would be `:returns:`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+
+    "aggInfo" aggregation information, JSON null if no aggregation is needed
+      otherwise, a JSON object contains the folowing properties:
+        "windowSize": aggregation window size in seconds (integer)
+        "func": A string representation of the aggregation function (string)
+          the following values are supported: "sum", "mean"
+
+    "modelInfo": JSON object describing the model configuration that
+    contains the following properties:
+      “modelConfig”: OPF Model Configuration parameters (JSON object) for
```
vitaly-krugl: Want to stick with ASCII quotes to avoid compatibility issues. Search your code for other uses of these quotes.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+  Return a JSON object describing the model configuration
+  :param useTimeOfDay: bool, whether to use timeOfDay encoder
+  :param useDayOfWeek: bool, whether to use dayOfWeej encoder
+  :param value: numpy array of metric data, used to compute min/max values
+  """
+  modelParams = getScalarMetricWithTimeOfDayAnomalyParams(metricData=value)
+
+  if useTimeOfDay:
+    modelParams['modelConfig']['modelParams']['sensorParams']['encoders'] \
+      ['c0_timeOfDay'] = dict(fieldname='c0',
```
vitaly-krugl: The logic here is very tightly coupled with `getScalarMetricWithTimeOfDayAnomalyParams`, making this implementation brittle. This implementation could easily break in subtle, unpredictable ways if object returned by `getScalarMetricWithTimeOfDayAnomalyParams` changes in certain ways.

For example, if `getScalarMetricWithTimeOfDayAnomalyParams` renames `c0_timeOfDay`, the result would not be meaningful.

Need logic (possibly in tests) to detect this.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+  samplingInterval = samplingInterval.astype('float64')
+  cumulativeCwtVar = numpy.cumsum(cwtVar)
+  cutoffTimeScale = timeScale[numpy.where(cumulativeCwtVar >= thresh)[0][0]]
+
+  aggregationTimeScale = cutoffTimeScale / 10.0
+  if aggregationTimeScale < samplingInterval * 4:
+    aggregationTimeScale = samplingInterval * 4
+
+  # make sure there is at least 1000 records after aggregation
+  if numDataPts < 1000:
```
vitaly-krugl: Use a module-level named constant instead of `1000` here. This would enable unit tests to use the constant to test each branch.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+    if ((localMaxValue - nearestLocalMinValue) / localMaxValue
+          > 0.1 and localMaxValue > baseline_value):
+      strongLocalMax.append(localMax[i])
+
+      if (timeScale[leftLocalMin] < dayPeriod < timeScale[rightLocalMin]
+          and cwtVarAtDayPeriod > localMaxValue * 0.5):
+        useTimeOfDay = True
+
+      if (timeScale[leftLocalMin] < weekPeriod < timeScale[
+        rightLocalMin]
```
vitaly-krugl: The indentation here doesn't look right. Try the following

```
if (timeScale[leftLocalMin] < weekPeriod < 
    timeScale[rightLocalMin] and
    cwtVarAtWeekPeriod > localMaxValue * 0.5):
```

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+          and cwtVarAtWeekPeriod > localMaxValue * 0.5):
```
vitaly-krugl: The indentation here doesn't look right. Try the following

```
if (timeScale[leftLocalMin] < weekPeriod < 
    timeScale[rightLocalMin] and
    cwtVarAtWeekPeriod > localMaxValue * 0.5):
```

unicorn/py/unicorn_backend/param_finder_runner.py
```diff
@@ -0,0 +1,242 @@
...
+    values = []
+
+    numRow = 0
+    for row in fileReader:
+      timeStamp = datetime.datetime.strptime(row[timestampIndex],
+                                             datetimeFormat)
+
+      # use local timezone to be consistent with the default assumption
+      # in numpy datetime64 object
+      timeStamp = timeStamp.replace(tzinfo=tz.tzlocal())
```
vitaly-krugl: This may result in invalid date that would cause an exception at some point in execution. For example, if the original timestamp falls in the non-existent one-hour gap that happens during spring-time daylight savings time adjustment, then the resulting timestamp becomes invalid.

Perhaps a UTC time zone would be a better choice.

Also, add a data sample to test for this.

unicorn/py/unicorn_backend/param_finder_runner.py
```diff
@@ -0,0 +1,242 @@
...
+  :param valueIndex: int, column index of the value
+  :param datetimeFormat: str, datetime format string for python's
+                        datetime.strptime
+  :return: A two-tuple (timestamps, values), where
+           timeStamps is a list of datetime
+           values is a list of float data values
+  """
+
+  with open(fileName, 'rU') as csvFile:
+    csv.field_size_limit(2 ** 27)
```
vitaly-krugl: Also need to tell the system to use "\n" as line separators for Windows compatibility. See https://github.com/numenta/numenta-apps/blob/64fcd471fb79b97909537179528734256fcb7333/unicorn/py/unicorn_backend/model_runner_2.py#L254-L255

unicorn/py/unicorn_backend/param_finder_runner.py
```diff
@@ -0,0 +1,242 @@
...
+import sys
+import traceback
+
+from dateutil import tz
+
+from param_finder import find_parameters
+
+g_log = logging.getLogger(__name__)
+
+_MAX_ROW_NUMBER = 20000
```
vitaly-krugl: Don't define your own max row number. Use the public one from param_finder.py instead.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+
+from nupic.frameworks.opf.common_models.cluster_params import (
+  getScalarMetricWithTimeOfDayAnomalyParams)
+
+_CORRELATION_MODE_VALID = 0
+_CORRELATION_MODE_SAME = 1
+_CORRELATION_MODE_FULL = 2
+
+_AGGREGATION_WINDOW_THRESH = 0.2
+MAX_ROW_NUMBER = 20000
```
vitaly-krugl: `MAX_NUM_ROWS` would be a better name. Also, add a comment that explains what it's for

unicorn/py/unicorn_backend/param_finder_runner.py
```diff
@@ -0,0 +1,242 @@
...
+
+  class SilentArgumentParser(ArgumentParser):
+    def error(self, msg):
+      """Override `error()` to prevent unstructured output to stderr"""
+      raise _CommandLineArgError(msg)
+
+
+  parser = SilentArgumentParser(description=("Start Unicorn ParamFinderRunner"))
+
+  parser.add_argument("--csv",
```
vitaly-krugl: These options don't match the interface specification for Param Finder. See "ParamFinder Command-line Args" in https://docs.google.com/document/d/1XZYGFCIoK7NarAl-DzCqluaTHac1PYFZQaEiu0r2rhs/edit

unicorn/py/unicorn_backend/tests/unit/param_finder_runner_test.py
```diff
@@ -0,0 +1,126 @@
...
+
+      argumentPattern = ["unicorn_backend/param_finder_runner.py"
+                         ] + argumentPattern
+      with patch.object(sys, "argv", argumentPattern):
+        # pylint: disable=W0212)
+        param_finder_runner._parseArgs()
+
+
+    _assertArgumentPatternFails()
+    _assertArgumentPatternFails(["--csv="])
```
vitaly-krugl: See comment about invalid options https://github.com/numenta/numenta-apps/pull/571#discussion_r51661994

unicorn/py/unicorn_backend/tests/unit/param_finder_runner_test.py
```diff
@@ -0,0 +1,126 @@
...
+                                 '--rowOffset=1',
+                                 '--timestampIndex=1',
+                                 '--datetimeFormat="%Y-%m-%d %H:%M:%S"'])
+
+
+  def testReadCSVFile(self):
+    """
+    Verify CSV file can be correctly read via  _readCSVFile function
+    """
+    tmpfilepath = os.path.join(tempfile.gettempdir(), "tmp-testfile.csv")
```
vitaly-krugl: Use mkdtemp instead (safe) with immediate addCleanup like this https://github.com/numenta/numenta-apps/blob/72ea1f688fd5309dc00445fa2bcaf06f161571b4/htm.it/tests/py/integration/app/runtime/metric_collector_int_test.py#L121-L123

and get rid of os.remove at the bottom

unicorn/py/unicorn_backend/tests/unit/param_finder_runner_test.py
```diff
@@ -0,0 +1,126 @@
...
+      timestampIndex=0,
+      valueIndex=1,
+      datetimeFormat="%Y-%m-%d %H:%M:%S"
+    )
+    self.assertAlmostEqual(values[0], 20.0)
+    self.assertAlmostEqual(timeStamps[0],
+                           numpy.datetime64('2014-04-01 00:00:00'))
+    os.remove(tmpfilepath)
+
+
```
vitaly-krugl: Need end-to-end test that calls param finder runner's `main` with the appropriate `sys.argv` and then validates the value written to stdout.

For this, 

1. Create a temp csv file
2. patch `sys.argv` with appropriate args
3. patch sys.stdout with `StringIO.StringIO` object
4. Call `main`
5. After `main` returns, verify that the `StringIO.StringIO` object has the expected values
6. Verify that the `StringIO.StringIO` object's contents match the schema for the return value (you will need to define a JSON schema for param finder's result similar to the ones that I defined for ModelRunner's command-line option values)

unicorn/py/unicorn_backend/tests/unit/param_finder_test.py
```diff
@@ -0,0 +1,175 @@
...
+      timeScale=timeScale,
+      cwtVar=cwtVar,
+      thresh=0.2,
+      samplingInterval=samplingInterval,
+      numDataPts=40000
+    )
+    self.assertLessEqual(aggregationTimeScale, maxSamplingInterval)
+    self.assertGreater(aggregationTimeScale, samplingInterval.astype('float64'))
+
+
```
vitaly-krugl: Missing test for `find_parameters`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+
+  (_cwtVar, _timeScale) = calculateContinuousWaveletTransform(
+    medianSamplingInterval, values)
+
+  suggestedSamplingInterval = determineAggregationWindow(
+    timeScale=_timeScale,
+    cwtVar=_cwtVar,
+    thresh=_AGGREGATION_WINDOW_THRESH,
+    samplingInterval=medianSamplingInterval,
+    numDataPts=numDataPts)
```
vitaly-krugl: Need to test both code paths: below numDataPts threshold and above. See https://github.com/numenta/numenta-apps/pull/571#discussion_r51649813

vitaly-krugl:
@ywcui1990, I am done with this round of code review. Please seem my comments for issues that need to be addressed.

I usually did not repeat comments for an issue that occurs in multiple places for things like coding/naming conventions. For example, if you see feedback about format of parameter or return value documentation, self-review and correct all parameter and/or return value documentation in all scripts in this PR that are not consistent with our convention.

Also, run pylint on all scripts, including tests, and fix what it finds. The pylint configuration file for nupic is is nupic/pylintrc and the one for numenta-apps is in numenta-apps/pylintrc.

CC @marionleborgne

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -0,0 +1,537 @@
...
+  for ind, width in enumerate(widths):
+    waveletData = wavelet(min(10 * width, len(data)), width)
+    output[ind, :] = _convolve(data, waveletData, mode=_CORRELATION_MODE_SAME)
+  return output
+
+
+
+def find_parameters(timeStamps, values):
+  """
+  find parameters for a given time series dataset with heuristics
```
vitaly-krugl: Describe in docstring the expectation of `MAX_ROW_NUMBER` rows.

ywcui1990:
@marionleborgne @lscheinkman 
Here is a version of the code that follows the technical requirements with the correct input/output. I am still addressing Vitaly's comments on internal implementations and coding styles. But this version should unblock you for now.

unicorn/py/unicorn_backend/input_opt_schema_param_finder.json
```diff
@@ -0,0 +1,37 @@
+{
+  "description": "--input command-line option object passed to model runner",
```
vitaly-krugl: Need to update descriptions that refer to model-runner-specifics.

unicorn/py/unicorn_backend/input_opt_schema_param_finder.json
```diff
@@ -0,0 +1,37 @@
...
+  "type": "object",
+  "additionalProperties": false,
+
+  "properties": {
+    "csv": {
+      "description": "Absolute path to input CSV file; if omitted, will read CSV from stdin.",
```
vitaly-krugl: "if omitted, will read CSV from stdin" doesn't belong here, since "csv" path in not optional in param finder runner

marionleborgne:
FYI, currently testing the script. Reported errors to @ywcui1990

marionleborgne:
:+1:  it works. Thanks @ywcui1990. You can merge this one and open another PR to address @vitaly-krugl 's comments.

ywcui1990:
@marionleborgne I don't have write access to this repository. Feel free to merge it on my behalf. I will address the rest of @vitaly-krugl 's comments in another PR.

marionleborgne:
:+1: great. Thanks @ywcui1990

PR #581, opened 2016-01-27
UNI-300 Implemented model_runner per phase 2 ModelRunner interface specification

vitaly-krugl:
CC @scottpurdy @marionleborgne @ywcui1990 

The new Model Runner module is initially named `model_runner_2.py`, while we still have the original around. We can get rid of the old and rename the new module after FrontEnd integrates with the new Model Runner.

### DIFF
--- a/unicorn/js/config/momentjs_to_datetime_strptime.json
+++ b/unicorn/js/config/momentjs_to_datetime_strptime.json
@@ -1,0 +1,42 @@
+{
+  "ISO 8601": {
+    "YYYY-MM-DDTHH:mm:ss.SZ": "%Y-%m-%dT%H:%M:%S.%fZ",
+    "YYYY-MM-DDTHH:mm:ss.S": "%Y-%m-%dT%H:%M:%S.%f",
+    "YYYY-MM-DDTHH:mm:ssZ": "%Y-%m-%dT%H:%M:%SZ",
+    "YYYY-MM-DDTHH:mm:ss": "%Y-%m-%dT%H:%M:%S",
+    "YYYY-MM-DDTHH:mmZ": "%Y-%m-%dT%H:%MZ",
+    "YYYY-MM-DDTHH:mm": "%Y-%m-%dT%H:%M"
+  },
+
+  "ISO 8601 no T": {
+    "YYYY-MM-DD HH:mm:ss.SZ": "%Y-%m-%d %H:%M:%S.%fZ",
+    "YYYY-MM-DD HH:mm:ss.S": "%Y-%m-%d %H:%M:%S.%f",
+    "YYYY-MM-DD HH:mm:ssZ": "%Y-%m-%d %H:%M:%SZ",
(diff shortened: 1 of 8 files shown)


unicorn/js/config/momentjs_to_datetime_strptime.json
```diff
@@ -0,0 +1,42 @@
+{
+  "ISO 8601": {
```
vitaly-krugl: @marionleborgne, I packaged some work-in-progress datetime format mappings (from moment.js to python datetime.strptime format) in this JSON file for use as a starting point for the format mapping logic. It's not referenced by the backend code, so FrontEnd may modified it as needed.

CC @lscheinkman @brev

marionleborgne:
@vitaly-krugl The PR looks good to me except one thing: I noticed that when running the model runner, it doesn't complain about me not passing it a value for the flag `--model`. If I don't pass `--model`  the error returned ends up being:
```
{"errorText": "expected string or buffer", "diagnosticInfo": "Traceback (most recent call last):\n  File \"model_runner_2.py\", line 389, in main\n    options = _parseArgs()\n  File \"model_runner_2.py\", line 165, in _parseArgs\n    modelSpec = json.loads(options.modelSpec)\n  File \"/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py\", line 338, in loads\n    return _default_decoder.decode(s)\n  File \"/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py\", line 366, in decode\n    obj, end = self.raw_decode(s, idx=_w(s, 0).end())\nTypeError: expected string or buffer\n"}
```

Instead of being something like:
```
{"errorText": "argument --input is required", "diagnosticInfo": "Traceback (most recent call last):\n  File \"model_runner_2.py\", line 389, in main\n    options = _parseArgs()\n  File \"model_runner_2.py\", line 124, in _parseArgs\n    options = parser.parse_args()\n  File \"/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/argparse.py\", line 1701, in parse_args\n    args, argv = self.parse_known_args(args, namespace)\n  File \"/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/argparse.py\", line 1733, in parse_known_args\n    namespace, args = self._parse_known_args(args, namespace)\n  File \"/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/argparse.py\", line 1957, in _parse_known_args\n    self.error(_('argument %s is required') % name)\n  File \"model_runner_2.py\", line 92, in error\n    raise _CommandLineArgError(msg)\n_CommandLineArgError: argument --input is required\n"}
```

Since the param `model` is mandatory we should consider returning a more explicit error.

vitaly-krugl:
Thanks, @marionleborgne, I will take a look now. I expected the `required=True` in the argparse option to do the "right thing".

PR #589, opened 2016-01-29
IMBU-94: Add `IMBU_DOCKER_OPTIONS` environment variable

lscheinkman:
@oxtopus Please review


oxtopus:
Please reconcile with `${IMBU_DOCKER_EXTRAS}`, which serves the same purpose.

oxtopus:
Actually, now that I look at it, they aren't exactly the same, but the two are potentially confusing.

lscheinkman:
@oxtopus `IMBU_DOCKER_EXTRAS` refers to options for the `docker run` command only. `IMBU_DOCKER_OPTIONS` refers to option for all `docker` client commands.

PR #590, opened 2016-01-29
TAUR-1761 Cleanup duplicated config, including removal of unused docker configurations

oxtopus:
This primarily removes the duplicated nginx config, requiring a new env var (`NGINX_GLOBAL_PARAMS`) to specify global nginx parameters such as user, and any other possible global configuration directives (see http://nginx.org/en/docs/ngx_core_module.html)


vitaly-krugl:
`unicorn/py/unicorn_backend/tests/unit/__init__.py` should not be added by this PR.

vitaly-krugl:
LGTM once `unicorn/py/unicorn_backend/tests/unit/__init__.py` is removed from the PR

PR #598, opened 2016-02-05
UNI-305 Package update block

brev:
- Update packages
- Fix couple of warnings
- Load Material-UI components individually (see http://www.material-ui.com/#/get-started/usage)
- Remove icon from chart titles

Please review @lscheinkman or @marionleborgne


brev:
https://jira.numenta.com/browse/UNI-305

marionleborgne:
@brev LGTM. We keep getting a warning [1]. Can we fix this as well as part of this PR?

[1] ```Warning: img is a void element tag and must not have `children` or use `props.dangerouslySetInnerHTML`. Check the render method of Avatar.```

brev:
thanks @marionleborgne 
I think that error is from inside Material-UI, so it might take a little longer to fix, I'll take a look on my current task, thanks.

PR #599, opened 2016-02-05
Updated windows build script

marionleborgne:
@oxtopus @lscheinkman

### DIFF
--- a/unicorn/scripts/Windows64/simple_build_portable_python_with_nupic.ps1
+++ b/unicorn/scripts/Windows64/simple_build_portable_python_with_nupic.ps1
@@ -1,4 +1,10 @@
-﻿# Make the script fail on cmlets errors
+﻿# See README.md for instructions about prerequisites and how to run this script.
+
+param (
+    [string]$nupic_version = "0.3.6"
+)
+
+# Make the script fail on cmlets errors
 $ErrorActionPreference = "Stop"
 
 $script_path = split-path -parent $MyInvocation.MyCommand.Definition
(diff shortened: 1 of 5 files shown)


marionleborgne:
The script `simple_build_portable_python_with_nupic.ps1` will install a portable python distribution with NuPIC (with `pip install nupic`). We can use this script to build the Windows portable python distribution with Bamboo until Windows64 artifacts are being built by Bamboo - at that point I'll use these .whl to install nupic with the portable python distrib instead of pip installing it like what `simple_build_portable_python_with_nupic.ps1` is doing.

unicorn/scripts/Windows64/simple_build_portable_python_with_nupic.ps1
```diff
@@ -0,0 +1,82 @@
+﻿# See README.md for instructions about prerequisites and how to run this script.
+
+param (
+    [string]$nupic_version = "0.3.6"
```
lscheinkman: We need a better way to specify `nupic` version. Currently it is hardcoded in every script and requirement file.
 1. `0.3.4`: https://github.com/numenta/numenta-apps/blob/master/unicorn/py/requirements.txt#L6
 1. `master`: https://github.com/numenta/numenta-apps/blob/master/unicorn/scripts/OSX/build-python.sh#L83
 1. `master`: https://github.com/numenta/numenta-apps/blob/master/unicorn/scripts/OSX/build-python.sh#L112
 1.  `0.3.6`:  https://github.com/marionleborgne/numenta-apps/blob/updated-windows-build/unicorn/scripts/Windows64/simple_build_portable_python_with_nupic.ps1#L4

lscheinkman:
@marionleborgne  The scripts worked for me. We still need a better way to specify nupic version but besides that it looks good to me.

marionleborgne:
@lscheinkman should we deal with this nupic version issue now? The scripts will have to be updated again anyway, once the nupic / nupic.bindings artifacts are available on bamboo.

lscheinkman:
@marionleborgne We can postpone the generic solution for version issue to a later time but we should at least use the same version for all scripts. Let's say `0.3.6` everywhere.

marionleborgne:
OK. Merging this now and creating another PR for the nupic versions

PR #602, opened 2016-02-07
[UNI-299] Implementation of paramFinder and paramFinder runner 

ywcui1990:
@marionleborgne @vitaly-krugl 

Address the rest of Vitaly's comment for https://github.com/numenta/numenta-apps/pull/571

Summary
* Use Doxygen convention for documentation in param_finder
* Add end-to-end test for param_finder runner
* Add tests for find_parameters, _getModelParams
* Fix many code styling issues

### DIFF
--- a/unicorn/py/unicorn_backend/param_finder.py
+++ b/unicorn/py/unicorn_backend/param_finder.py
@@ -29,7 +29,7 @@ Implements param finder, which automatically
 """
 
 import datetime
-
+import sys
 import numpy
 
 from nupic.frameworks.opf.common_models.cluster_params import (
@@ -42,7 +42,11 @@ _CORRELATION_MODE_FULL = 2
 _AGGREGATION_WINDOW_THRESH = 0.2
 
 # Maximum number of rows param_finder will process
+# If the input data exceeds MAX_NUM_ROWS, the first MAX_NUM_ROWS will be used
 MAX_NUM_ROWS = 20000
+# Minimum number of rows. If the number of record in the input data is
+# less than MIN_NUM_ROWS, the param_finder will return the default parameters
+MIN_NUM_ROWS = 100
 MIN_ROW_AFTER_AGGREGATION = 1000
 
 
@@ -193,15 +197,21 @@ def find_parameters(data):
       corresponding to the metric value (string)
   """
   (timeStamps, values) = data
+  numRecords = len(data)
 
   if type(timeStamps[0]) is not datetime.datetime:
     raise TypeError('timeStamps must be datetime type')
 
-  if len(timeStamps) > MAX_NUM_ROWS:
+  if numRecords > MAX_NUM_ROWS:
     timeStamps = timeStamps[:MAX_NUM_ROWS]
-
-  if len(values) > MAX_NUM_ROWS:
     values = values[:MAX_NUM_ROWS]
+
+  if numRecords < MIN_NUM_ROWS:
+    outputInfo = {
+      "aggInfo": None,
+      "modelInfo": _getModelParams(True, False, values),
+    }
+    return outputInfo
 
   timeStamps = numpy.array(timeStamps, dtype='datetime64[s]')
 
--- a/unicorn/py/unicorn_backend/param_finder.py
+++ b/unicorn/py/unicorn_backend/param_finder.py
@@ -215,11 +215,7 @@ def findParameters(samples):
     return outputInfo
 
   timestamps = numpy.array(timestamps, dtype="datetime64[s]")
-
-  try:
-    values = numpy.array(values).astype("float64")
-  except TypeError:
-    print "values must be an array with float numbers"
+  values = numpy.array(values).astype("float64")
 
   numDataPts = len(values)
 
--- a/unicorn/py/unicorn_backend/param_finder.py
+++ b/unicorn/py/unicorn_backend/param_finder.py
@@ -197,7 +197,7 @@ def find_parameters(data):
       corresponding to the metric value (string)
   """
   (timeStamps, values) = data
-  numRecords = len(data)
+  numRecords = len(data[0])
 
   if type(timeStamps[0]) is not datetime.datetime:
     raise TypeError('timeStamps must be datetime type')
@@ -366,7 +366,7 @@ def _calculateContinuousWaveletTransform(samplingInterval, values):
     'timeScale' (numpy array) Stores the corresponding time scales
   """
 
-  widths = numpy.logspace(0, numpy.log10(len(values) / 20), 50)
+  widths = numpy.logspace(0, numpy.log10(len(values) / 10), 50)
   timeScale = widths * samplingInterval * 4
 
   # continuous wavelet transformation with ricker wavelet
--- a/unicorn/py/unicorn_backend/param_finder.py
+++ b/unicorn/py/unicorn_backend/param_finder.py
@@ -46,6 +46,9 @@ MAX_NUM_ROWS = 20000
 # Minimum number of rows. If the number of record in the input data is
 # less than MIN_NUM_ROWS, the param_finder will return the default parameters
 MIN_NUM_ROWS = 100
+# Minimum number of rows after the data aggregation. The suggested aggInfo will
+# make sure more than MIN_ROW_AFTER_AGGREGATION number of records after
+# aggregation
 MIN_ROW_AFTER_AGGREGATION = 1000
 
 
@@ -86,16 +89,16 @@ def _convolve(vector1, vector2, mode):
         for points where the signals overlap completely.  Values outside
         the signal boundary have no effect.
 
-  @returns (ndarray) Discrete, linear convolution of `vector1` and `vector2`
+  @return (ndarray) Discrete, linear convolution of `vector1` and `vector2`
 
   """
   vector1 = numpy.array(vector1, ndmin=1)
   vector2 = numpy.array(vector2, ndmin=1)
 
   if len(vector1) == 0:
-    raise ValueError('vector1 cannot be empty')
+    raise ValueError("vector1 cannot be empty")
   if len(vector2) == 0:
-    raise ValueError('vector2 cannot be empty')
+    raise ValueError("vector2 cannot be empty")
 
   if len(vector2) > len(vector1):
     vector1, vector2 = vector2, vector1
@@ -118,7 +121,7 @@ def _rickerWavelet(numPoints, waveletWidth):
 
   @param waveletWidth (scalar) Width parameter of the wavelet
 
-  @returns (ndarray) Array of length `points` in shape of ricker curve
+  @return (ndarray) Array of length `points` in shape of ricker curve
 
   """
   normalizeFactor = 2 / (numpy.sqrt(3 * waveletWidth) * (numpy.pi ** 0.25))
@@ -152,7 +155,7 @@ def _cwt(data, wavelet, widths):
 
   @param widths (sequence) Widths to use for transform
 
-  @returns (ndarray) Will have shape of (len(data), len(widths))
+  @return (ndarray) Will have shape of (len(data), len(widths))
 
   """
   output = numpy.zeros([len(widths), len(data)])
@@ -163,15 +166,14 @@ def _cwt(data, wavelet, widths):
 
 
 
-def find_parameters(data):
+def findParameters(samples):
   """
   Find parameters for a given time series dataset with heuristics.
 
-  @param data (tuple) Contains:
-    'timeStamps' (sequence) Each entry is an datetime.datetime instance
-    'values' (sequence) Each entry is a float number
+  @param samples Sequence of two tuples (timestamp, value), where
+    timestamp of type datetime.datetime and value is a number (int of float)
 
-  @returns: JSON object with the following properties:
+  @return: JSON object with the following properties:
 
     "aggInfo" aggregation information, JSON null if no aggregation is needed
       otherwise, a JSON object contains the folowing properties:
@@ -182,8 +184,8 @@ def find_parameters(data):
     "modelInfo": JSON object describing the model configuration that
     contains the following properties:
       "modelConfig": OPF Model Configuration parameters (JSON object) for
-      passing to the OPF 'ModelFactory.create()' method as the
-      'modelConfig' parameter.
+      passing to the OPF "ModelFactory.create()" method as the
+      "modelConfig" parameter.
 
       "inferenceArgs": OPF Model Inference parameters (JSON object) for
       passing to the resulting model's `enableInference()` method as the
@@ -195,14 +197,14 @@ def find_parameters(data):
       "valueFieldName": The name of the field in 'modelConfig'
       corresponding to the metric value (string)
   """
-  (timeStamps, values) = data
-  numRecords = len(data[0])
+  (timestamps, values) = zip(*samples)
+  numRecords = len(timestamps)
 
-  if type(timeStamps[0]) is not datetime.datetime:
-    raise TypeError('timeStamps must be datetime type')
+  if not isinstance(timestamps[0], datetime.datetime):
+    raise TypeError("timestamps must be datetime type")
 
   if numRecords > MAX_NUM_ROWS:
-    timeStamps = timeStamps[:MAX_NUM_ROWS]
+    timestamps = timestamps[:MAX_NUM_ROWS]
     values = values[:MAX_NUM_ROWS]
 
   if numRecords < MIN_NUM_ROWS:
@@ -212,46 +214,46 @@ def find_parameters(data):
     }
     return outputInfo
 
-  timeStamps = numpy.array(timeStamps, dtype='datetime64[s]')
+  timestamps = numpy.array(timestamps, dtype="datetime64[s]")
 
   try:
-    values = numpy.array(values).astype('float64')
-  except:
-    raise TypeError('values must be an array with float numbers')
+    values = numpy.array(values).astype("float64")
+  except TypeError:
+    print "values must be an array with float numbers"
 
   numDataPts = len(values)
 
   (medianSamplingInterval,
-   medianAbsoluteDevSamplingInterval) = _getMedianSamplingInterval(timeStamps)
+   medianAbsoluteDevSamplingInterval) = _getMedianSamplingInterval(timestamps)
 
-  (timeStamps, values) = _resampleData(timeStamps,
+  (timestamps, values) = _resampleData(timestamps,
                                        values,
                                        medianSamplingInterval)
 
-  (_cwtVar, _timeScale) = _calculateContinuousWaveletTransform(
-      medianSamplingInterval, values)
+  (cwtVar, timeScale) = _calculateContinuousWaveletTransform(
+    medianSamplingInterval, values)
 
   suggestedSamplingInterval = _determineAggregationWindow(
-      timeScale=_timeScale,
-      cwtVar=_cwtVar,
-      thresh=_AGGREGATION_WINDOW_THRESH,
-      samplingInterval=medianSamplingInterval,
-      numDataPts=numDataPts)
+    timeScale=timeScale,
+    cwtVar=cwtVar,
+    thresh=_AGGREGATION_WINDOW_THRESH,
+    samplingInterval=medianSamplingInterval,
+    numDataPts=numDataPts)
 
   # decide whether to use TimeOfDay and DayOfWeek encoders
-  (useTimeOfDay, useDayOfWeek) = _determineEncoderTypes(_cwtVar, _timeScale)
+  (useTimeOfDay, useDayOfWeek) = _determineEncoderTypes(cwtVar, timeScale)
 
   # decide the aggregation function ("mean" or "sum")
   aggFunc = _getAggregationFunction(medianSamplingInterval,
-                                    medianAbsoluteDevSamplingInterval)
+                                    medianAbsoluteDevSamplingInterval,
+                                    _AGGREGATION_WINDOW_THRESH)
 
-  outputInfo = {
+  return {
     "aggInfo": _getAggInfo(medianSamplingInterval,
                            suggestedSamplingInterval,
                            aggFunc),
     "modelInfo": _getModelParams(useTimeOfDay, useDayOfWeek, values),
   }
-  return outputInfo
 
 
 
@@ -272,7 +274,7 @@ def _getAggInfo(medianSamplingInterval, suggestedSamplingInterval, aggFunc):
     aggInfo = None
   else:
     aggInfo = {
-      "windowSize": suggestedSamplingInterval.astype('int'),
+      "windowSize": suggestedSamplingInterval.astype("int"),
       "func": aggFunc
     }
   return aggInfo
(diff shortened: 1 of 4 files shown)


unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -40,7 +39,14 @@
...
 _AGGREGATION_WINDOW_THRESH = 0.2
-MAX_ROW_NUMBER = 20000
+
+# Maximum number of rows param_finder will process
+# If the input data exceeds MAX_NUM_ROWS, the first MAX_NUM_ROWS will be used
+MAX_NUM_ROWS = 20000
+# Minimum number of rows. If the number of record in the input data is
+# less than MIN_NUM_ROWS, the param_finder will return the default parameters
+MIN_NUM_ROWS = 100
+MIN_ROW_AFTER_AGGREGATION = 1000
```
vitaly-krugl: Document `MIN_ROW_AFTER_AGGREGATION`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -206,17 +195,22 @@ def find_parameters(timeStamps, values):
       "valueFieldName": The name of the field in 'modelConfig'
       corresponding to the metric value (string)
   """
-  if len(timeStamps) != len(values):
-    raise ValueError('timeStamps and Values must have the same length')
+  (timeStamps, values) = data
+  numRecords = len(data[0])
 
   if type(timeStamps[0]) is not datetime.datetime:
```
vitaly-krugl: Pythonic: `if not isinstance(timeStamps[0], datetime.datetime)`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -206,17 +195,22 @@ def find_parameters(timeStamps, values):
       "valueFieldName": The name of the field in 'modelConfig'
       corresponding to the metric value (string)
   """
-  if len(timeStamps) != len(values):
-    raise ValueError('timeStamps and Values must have the same length')
+  (timeStamps, values) = data
+  numRecords = len(data[0])
```
vitaly-krugl: `numRecords = len(timeStamps)`, since you already have `timeStamps`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -175,14 +163,15 @@ def _cwt(data, wavelet, widths):
...
+def find_parameters(data):
   """
-  find parameters for a given time series dataset with heuristics
+  Find parameters for a given time series dataset with heuristics.
 
-  :param timeStamps: array_like, each entry is a timeStamp in datetime type
-  :param values: array_like, each entry is a float number
+  @param data (tuple) Contains:
+    'timeStamps' (sequence) Each entry is an datetime.datetime instance
+    'values' (sequence) Each entry is a float number
```
vitaly-krugl: The `data` parameter change here is different from what was suggested in feedback of previous PR and still exhibits the same issues. Here is clarification of previous feedback:

```
def find_parameters(data):
  """
  @param data Sequence of two-tuples (timestamp, scalar), where timestamp is of
      type `datetime.datetime` and scalar is a number (integer or float)
  """
```

For example:
```
result = find_parameters([(datetime(2016, 1, 1)), (2016, 1, 2), ...])
```

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -206,17 +195,22 @@ def find_parameters(timeStamps, values):
...
-  if len(values) > MAX_ROW_NUMBER:
-    values = values[:MAX_ROW_NUMBER]
+  if numRecords < MIN_NUM_ROWS:
+    outputInfo = {
+      "aggInfo": None,
+      "modelInfo": _getModelParams(True, False, values),
+    }
+    return outputInfo
 
   timeStamps = numpy.array(timeStamps, dtype='datetime64[s]')
```
vitaly-krugl: In the code block below:

```
  try:
    values = numpy.array(values).astype('float64')
  except:
    raise TypeError('values must be an array with float numbers')
```

Use a more specific exception instead of just `except:`.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -206,17 +195,22 @@ def find_parameters(timeStamps, values):
       "valueFieldName": The name of the field in 'modelConfig'
       corresponding to the metric value (string)
   """
-  if len(timeStamps) != len(values):
-    raise ValueError('timeStamps and Values must have the same length')
+  (timeStamps, values) = data
```
vitaly-krugl: nit - `timeStamps` doesn't need to be camel-case. `timestamp` is a legitimate word.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -231,30 +225,30 @@ def find_parameters(timeStamps, values):
    medianAbsoluteDevSamplingInterval) = _getMedianSamplingInterval(timeStamps)
 
   (timeStamps, values) = _resampleData(timeStamps,
-                                      values,
-                                      medianSamplingInterval)
+                                       values,
+                                       medianSamplingInterval)
 
   (_cwtVar, _timeScale) = _calculateContinuousWaveletTransform(
```
vitaly-krugl: Leading underscores are reserved for private members and private module-level constants. Local variables in a function should not have them. These should be `cwtVar, timeScale`

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -231,30 +225,30 @@ def find_parameters(timeStamps, values):
...
-    medianSamplingInterval, values)
+      medianSamplingInterval, values)
 
   suggestedSamplingInterval = _determineAggregationWindow(
-    timeScale=_timeScale,
-    cwtVar=_cwtVar,
-    thresh=_AGGREGATION_WINDOW_THRESH,
-    samplingInterval=medianSamplingInterval,
-    numDataPts=numDataPts)
+      timeScale=_timeScale,
```
vitaly-krugl: Per numenta's coding standard, indentation should be two spaces in python code.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -231,30 +225,30 @@ def find_parameters(timeStamps, values):
...
 
   outputInfo = {
     "aggInfo": _getAggInfo(medianSamplingInterval,
-                          suggestedSamplingInterval,
-                          aggFunc),
+                           suggestedSamplingInterval,
+                           aggFunc),
     "modelInfo": _getModelParams(useTimeOfDay, useDayOfWeek, values),
   }
   return outputInfo
```
vitaly-krugl: nit - this might be more concise without loss of readability:

```
 return {
    "aggInfo": _getAggInfo(medianSamplingInterval,
                           suggestedSamplingInterval,
                           aggFunc),
    "modelInfo": _getModelParams(useTimeOfDay, useDayOfWeek, values),
  }
```

vitaly-krugl:
Some of the code in `param_finder.py` is using `@returns` and some is using `@return`. I think that only one of the two is correct.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -500,8 +516,7 @@ def _determineEncoderTypes(cwtVar, timeScale):
           and cwtVarAtDayPeriod > localMaxValue * 0.5):
         useTimeOfDay = True
 
-      if (timeScale[leftLocalMin] < weekPeriod < timeScale[
-        rightLocalMin]
+      if (timeScale[leftLocalMin] < weekPeriod < timeScale[rightLocalMin]
           and cwtVarAtWeekPeriod > localMaxValue * 0.5):
         useDayOfWeek = True
```
vitaly-krugl: In the code snippet

```
 def _getAggregationFunction(medianSamplingInterval,
                             medianAbsoluteDevSamplingInterval,
                             aggregationFuncThresh=0.2):
```

Is the `0.2` default value related to `_AGGREGATION_WINDOW_THRESH` or is it just a coincidence that they have the same value? If they are semantically the same, then this function should use the same named constant instead of the "naked" constant.

Also, does it make sense to have a default value here? It's a private function that is only used once and without passing the `aggregationFuncThresh` arg. Either this parameter is not needed at all (at this time) and `_getAggregationFunction` should simply use a (named) constant internally, or the default value should be removed from the parameter and the caller should pass an explicit value.

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -519,13 +534,16 @@ def _getAggregationFunction(medianSamplingInterval,
```
vitaly-krugl: In the code snippet

```
 def _getAggregationFunction(medianSamplingInterval,
                             medianAbsoluteDevSamplingInterval,
                             aggregationFuncThresh=0.2):
```

Is the `0.2` default value related to `_AGGREGATION_WINDOW_THRESH` or is it just a coincidence that they have the same value? If they are semantically the same, then this function should use the same named constant instead of the "naked" constant.

Also, does it make sense to have a default value here? It's a private function that is only used once and without passing the `aggregationFuncThresh` arg. Either this parameter is not needed at all (at this time) and `_getAggregationFunction` should simply use a (named) constant internally, or the default value should be removed from the parameter and the caller should pass an explicit value.

unicorn/py/unicorn_backend/param_finder_runner.py
```diff
@@ -176,7 +187,7 @@ def _readCSVFile(fileName,
       values.append(float(row[valueIndex]))
 
       numRow += 1
```
vitaly-krugl: numRow => numRows

vitaly-krugl:
Need to fix up Sphinx param/return documentation in `param_finder_runner.py`. For example:

```
   :param fileName: str, path to input csv file
   :param rowOffset: int, index of first data row in csv
   :param timestampIndex: int, column index of the timeStamp
   :param valueIndex: int, column index of the value
   :param datetimeFormat: str, datetime format string for python's
                         datetime.strptime
   :return: A two-tuple (timestamps, values), where
            timeStamps is a list of datetime
            values is a list of float data values
```

Should be 
Note: use of param types between keyword param and param's name; also note `:return:` => `:returns:`
```
  :param str fileName: path to input csv file
  :param int rowOffset: index of first data row in csv
  :param int timestampIndex: column index of the timeStamp
  :param int valueIndex: column index of the value
  :param str datetimeFormat: datetime format string for python's
    datetime.strptime
  :returns: A two-tuple (timestamps, values), where
    timeStamps is a list of datetime 
    values is a list of float data values
```

unicorn/py/unicorn_backend/param_finder_runner.py
```diff
@@ -176,7 +187,7 @@ def _readCSVFile(fileName,
       values.append(float(row[valueIndex]))
```
vitaly-krugl: In the code snippet:

```
       timeStamp = datetime.datetime.strptime(row[timestampIndex],
                                              datetimeFormat)
 
       # use local timezone to be consistent with the default assumption
       # in numpy datetime64 object
       timeStamp = timeStamp.replace(tzinfo=tz.tzutc())
```
What if `datetime.datetime.strptime` returned a `datetime` with `tzinfo` already in it? Could the unconditional `replace` have an adverse impact on the resulting parameters in such a scenario?

Also, the comment "use local timezone to be consistent with the default assumption numpy datetime64 object" is not consistent with the code, which now uses UTC time zone (not local).

unicorn/tests/py/unit/unicorn_backend/param_finder_runner_test.py
```diff
@@ -77,37 +76,76 @@ def _assertArgumentPatternPasses(argumentPattern=None):
...
     (timeStamps, values) = param_finder_runner._readCSVFile(
-      fileName=tmpfilepath,
+      fileName=csvPath,
       rowOffset=1,
       timestampIndex=0,
       valueIndex=1,
       datetimeFormat="%Y-%m-%d %H:%M:%S"
     )
+
     self.assertAlmostEqual(values[0], 20.0)
```
vitaly-krugl: Why is "almost equal" needed? Is parsing of `20.0` losing precision?

unicorn/tests/py/unit/unicorn_backend/param_finder_runner_test.py
```diff
@@ -77,37 +76,76 @@ def _assertArgumentPatternPasses(argumentPattern=None):
...
     self.assertAlmostEqual(str(timeStamps[0]),
```
vitaly-krugl: Why not simply `self.assertEqual` for the strings? The "almost equal" has some confusing implications for the maintainer/reader/reviewer of the code.

unicorn/tests/py/unit/unicorn_backend/param_finder_runner_test.py
```diff
@@ -77,37 +76,76 @@ def _assertArgumentPatternPasses(argumentPattern=None):
...
     """
-    tmpfilepath = os.path.join(tempfile.gettempdir(), "tmp-testfile.csv")
-    with open(tmpfilepath, "wb") as file:
-      fileWriter = csv.writer(file)
-      fileWriter.writerow(['timeStamps', 'values'])
-      fileWriter.writerow(['2014-04-01 00:00:00', 20.0])
+    csvFd, csvPath = tempfile.mkstemp()
+    self.addCleanup(os.unlink, csvPath)
+
+    with open(csvPath, "wb") as csvFile:
```
vitaly-krugl: Use `os.fdopen` instead so that file descriptor `csvFd` would be closed automatically when scope leaves the context manager; see https://github.com/numenta/numenta-apps/blob/e09e5af00143acf48acd311da160592bf5f099d9/unicorn/tests/py/integration/unicorn_backend/model_runner_2_test.py#L643

unicorn/tests/py/unit/unicorn_backend/param_finder_runner_test.py
```diff
@@ -77,37 +76,76 @@ def _assertArgumentPatternPasses(argumentPattern=None):
...
+                  "rowOffset": 1,
+                  "csv": csvPath}
+    argumentPattern = ["unicorn_backend/param_finder_runner.py"] + \
+                      ["--input=" + json.dumps(INPUT_INFO)]
+
+    with patch.object(sys, "argv", argumentPattern):
+      with patch('sys.stdout', new=StringIO.StringIO()) as fake_out:
+        param_finder_runner.main()
+        outputInfo = fake_out.getvalue()
+        outputInfo = json.loads(outputInfo)
```
vitaly-krugl: nit - a bit more compact without losing readability: `outputInfo = json.loads(fake_out.getvalue())`

unicorn/tests/py/unit/unicorn_backend/param_finder_runner_test.py
```diff
@@ -77,37 +76,76 @@ def _assertArgumentPatternPasses(argumentPattern=None):
...
+
+        # since we used a flat line as test data, there should be no aggregation
+        self.assertTrue(outputInfo["aggInfo"] == None)
```
vitaly-krugl: Use `self.assertIsNone(outputInfo["aggInfo"])` instead. Using the special-purpose assertion methods allows unittest to print more useful diagnostic messages when the assertion fails.

unicorn/tests/py/unit/unicorn_backend/param_finder_test.py
```diff
@@ -172,4 +171,98 @@ def testDetermineAggregationWindow(self):
...
+      timeScale=timeScale,
+      cwtVar=cwtVar,
+      thresh=0.2,
+      samplingInterval=samplingInterval,
+      numDataPts=param_finder.MIN_ROW_AFTER_AGGREGATION-1
+    )
+    self.assertEqual(aggregationTimeScale, samplingInterval.astype('float64'))
+
+  def testFindParameters(self):
+    def createTestData(type, timeStepSeconds=300):
```
vitaly-krugl: Don't use names that override builtin objects. Here, `type` parameter overrides the builtin `type` function.

unicorn/tests/py/unit/unicorn_backend/param_finder_test.py
```diff
@@ -172,4 +171,98 @@ def testDetermineAggregationWindow(self):
...
+        if type == "daily":
+          values.append(numpy.sin(2 * numpy.pi *
+                                  (timeStep.seconds * i) / dayPeriod))
+        if type == "weekly":
+          values.append(numpy.sin(2 * numpy.pi *
+                                  (timeStep.seconds * i) / weekPeriod))
+      return (timeStamps, values)
+
+
+    outputInfo = param_finder.find_parameters(createTestData('flat', 300))
```
vitaly-krugl: In the assertions below, use the special-purpose assertions such as `self.assertIsNone`, `self.assertIsNotNone`, `self.assertGreater`, `self.assertEqual`, etc., to enable unittest to display more useful diagnostic messages when an assertion fails.

For example:

```
    self.assertGreater(outputInfo["aggInfo"]['windowSize'], 300)
    self.assertEqual(outputInfo["aggInfo"]['func'], 'mean')
    # Exclude timeOfDay and dayOfWeek encoder for flat line
    self.assertEqual(outputInfo["modelInfo"]['modelConfig']['modelParams'] \
                      ['sensorParams']['encoders']['c0_timeOfDay'], None)
    self.assertEqual(outputInfo["modelInfo"]['modelConfig']['modelParams'] \
                      ['sensorParams']['encoders']['c0_dayOfWeek'], None)


```

unicorn/tests/py/unit/unicorn_backend/param_finder_test.py
```diff
@@ -172,4 +171,98 @@ def testDetermineAggregationWindow(self):
...
+    self.assertTrue(outputInfo["aggInfo"]['windowSize'] > 300)
```
vitaly-krugl: In numenta-apps, the convention is to use double-quotes

unicorn/tests/py/unit/unicorn_backend/param_finder_test.py
```diff
@@ -172,4 +171,98 @@ def testDetermineAggregationWindow(self):
...
+    self.assertTrue(outputInfo["aggInfo"]['func'] == 'mean')
+    # Use timeOfDay but not dayOfWeek encoder for dataSet with daily period
+    self.assertFalse(outputInfo["modelInfo"]['modelConfig']['modelParams'] \
+                       ['sensorParams']['encoders']['c0_timeOfDay'] == None)
+    self.assertTrue(outputInfo["modelInfo"]['modelConfig']['modelParams'] \
+                      ['sensorParams']['encoders']['c0_dayOfWeek'] == None)
+
+    outputInfo = param_finder.find_parameters(createTestData('weekly', 7200))
+    self.assertTrue(outputInfo["aggInfo"]['windowSize'] > 7200)
+    self.assertTrue(outputInfo["aggInfo"]['func'] == 'mean')
```
vitaly-krugl: All of these test cases test for "mean". Should there also be a test for "sum"?

unicorn/tests/py/unit/unicorn_backend/param_finder_test.py
```diff
@@ -172,4 +171,98 @@ def testDetermineAggregationWindow(self):
...
+    # Use dayOfWeek but not timeOfDay encoder for dataSet with daily period
+    self.assertTrue(outputInfo["modelInfo"]['modelConfig']['modelParams'] \
```
vitaly-krugl: No need for line-continuation backslash. When inside parentheses, python interpreter does the right thing. Our coding convention favors placing long lines in parentheses instead of using the backslash.

unicorn/tests/py/unit/unicorn_backend/param_finder_test.py
```diff
@@ -172,4 +171,98 @@ def testDetermineAggregationWindow(self):
...
+                      ['sensorParams']['encoders']['c0_timeOfDay'] == None)
+    self.assertFalse(outputInfo["modelInfo"]['modelConfig']['modelParams'] \
+                       ['sensorParams']['encoders']['c0_dayOfWeek'] == None)
+
+
+  def testGetModelParams(self):
+    values = numpy.linspace(0, 10, 10)
+    modelParams = param_finder._getModelParams(
+      useTimeOfDay=False, useDayOfWeek=False, values=values)
+
```
vitaly-krugl: use `self.assertIsNone`, `self.assertIsNotNone`, etc.

vitaly-krugl:
Run pylint, and fix findings. For example, in `param_finder.py`, pylint complains:
```pylint
Line 232	Column 0	C0330: Wrong hanging indentation.
Line 235	Column 0	C0330: Wrong hanging indentation.
Line 236	Column 0	C0330: Wrong hanging indentation.
Line 237	Column 0	C0330: Wrong hanging indentation.
Line 238	Column 0	C0330: Wrong hanging indentation.
Line 239	Column 0	C0330: Wrong hanging indentation.
Line 347	Column 0	C0330: Wrong continued indentation.
Line 348	Column 0	C0330: Wrong continued indentation.
Line 512	Column 0	C0330: Wrong continued indentation.
Line 166	Column 0	C0103: Invalid function name "find_parameters"
Line 478	Column 2	C0103: Invalid variable name "baseline_value"
```

vitaly-krugl:
In tests that intentionally access private/protected members, add the following pragma just before the imports to disable such warnings:

```
# Disable pylint warnings concerning access to protected members
# pylint: disable=W0212
```

vitaly-krugl:
@ywcui1990, I am done with the review. Please address the findings. Use your judgement regarding feedback tagged as "nit" (for nitpick)

unicorn/py/unicorn_backend/param_finder.py
ywcui1990: It is intended to be public. The user may want to use further aggregation that results in a data file with less than 1000 rows.

unicorn/py/unicorn_backend/param_finder.py
ywcui1990: I changed it to: 
  try:
    values = numpy.array(values).astype("float64")
  except TypeError:
    print "values must be an array with float numbers"

Is it the right way to catch specific exceptions?

ywcui1990:
@vitaly-krugl Thanks for the detailed feedback. I have addressed your comments and added a new test for transaction data (aggFunc="sum"). I also ran pylint on my code and addressed most of the issues. There are still some remaining pylint issues that seems to be false alert, which are related to (1) order of imports (2) missing members of numpy and (3) unsubscriptable-object issues

unicorn/py/unicorn_backend/param_finder.py
```diff
@@ -206,75 +197,84 @@ def find_parameters(timeStamps, values):
...
-  timeStamps = numpy.array(timeStamps, dtype='datetime64[s]')
+  timestamps = numpy.array(timestamps, dtype="datetime64[s]")
 
   try:
-    values = numpy.array(values).astype('float64')
-  except:
-    raise TypeError('values must be an array with float numbers')
+    values = numpy.array(values).astype("float64")
+  except TypeError:
+    print "values must be an array with float numbers"
```
vitaly-krugl: @ywcui1990, there are a couple of problems with this:

1. The code now suppresses the TypeError exception and subsequent code will be in trouble when it attempts to access the `value` variable.
2. Printing from library code is not a good idea. Such messages go to stdout, and possibly corrupt the application's stdout data stream. So, we definitely don't want to print from here.
3. For now, we could remove this try/except and simply rely on numpy's own message in its exception

unicorn/py/unicorn_backend/param_finder.py
vitaly-krugl: No. I replied to this in another comment.

vitaly-krugl:
Thanks for addressing most of the feedback, @ywcui1990. We're almost there. Please address the new issue with except/print https://github.com/numenta/numenta-apps/pull/602#discussion-diff-52414448

unicorn/tests/py/unit/unicorn_backend/param_finder_test.py
```diff
@@ -169,7 +179,121 @@ def testDetermineAggregationWindow(self):
...
+    modelParams = param_finder._getModelParams(
+      useTimeOfDay=True, useDayOfWeek=True, values=values)
+    self.assertIsNotNone(modelParams["modelConfig"]["modelParams"]
+                         ["sensorParams"]["encoders"]["c0_timeOfDay"])
+    self.assertIsNotNone(modelParams["modelConfig"]["modelParams"]
+                         ["sensorParams"]["encoders"]["c0_dayOfWeek"])
+
+if __name__ == "__main__":
+  unittest.main()
```
vitaly-krugl: add one empty line at end of script per python convention.

ywcui1990:
Thanks @vitaly-krugl , I have addressed the except/print issue. Is there any other issue you noticed?

vitaly-krugl:
LGTM

ywcui1990:
Great. Please merge then @vitaly-krugl  I don't have write access to this repository.

vitaly-krugl:
Thanks, @ywcui1990

PR #607, opened 2016-02-08
More Post-upgrade tweaking

brev:
- Remove babel polyfill+runtime since Node 5 has internal es6 support.
  (move from `es2015` preset to `es2015-node`, etc)
- Fix API calls to new method names after pkg upgrade
- Upgrade to Babel 6.5

This should be it. We're all up to date now, and everything seems to be working great.

Please review @marionleborgne @lscheinkman


brev:
never mind, ran into new problem, will try again later

PR #609, opened 2016-02-08
Figured out how to use new API for Electron `ipc` after upgrade.

brev:
Figured out how to use new API for Electron `ipc` after upgrade.


brev:
FYI @lscheinkman @marionleborgne

PR #612, opened 2016-02-09
updated nupic & nupic.core version

marionleborgne:
> Note: Don't merge yet. 

WIP but has to be put on the back burner for now. Fixes UNI-317


marionleborgne:
Closing this PR which is obsolete now that we know that this script will be split up between nupic / nupic.core and unicorn build pipelines. We will have a mapping of the nupic version VS build number of the portable python distribution with the right version of nupic installed.

cc @lscheinkman @scottpurdy @oxtopus

PR #614, opened 2016-02-09
UNI-306 A. Remove extraneous button "secondary" property

brev:
UNI-306 A. Remove extraneous button "secondary" property (fixes mis-coloring)


brev:

PR #615, opened 2016-02-09
UNI-306 B. Automatically open freshly built docs after running

brev:
UNI-306 #comment B. Automatically open freshly built docs after running `npm run docs`. Upgrade dev pkg.


brev:

PR #619, opened 2016-02-10
UNI-319: Add `portable_python` as a npm dependency

lscheinkman:
@marionleborgne Please review.
CC: @brev


unicorn/js/main/ModelService.js
```diff
@@ -19,15 +19,24 @@ import childProcess from 'child_process';
 import EventEmitter from 'events';
 import path from 'path';
 import system from 'os';
+import process from 'process';
 import UserError from './UserError';
 
-// @todo https://bitbucket.org/anthony_tuininga/cx_freeze/issues/161
+// Get platform specific python.
+// See 'optionalDependencies' in 'package.json' for available platforms
+let PYTHON_EXECUTABLE;
```
marionleborgne: shouldn't we initialize this to `null` or something?

lscheinkman: It is `undefined` by default and if the module does not exist for the current platform it will throw an exception lines [28-32](https://github.com/lscheinkman/numenta-apps/blob/UNI-319/unicorn/js/main/ModelService.js#L28-L32)

PR #625, opened 2016-02-12
UNI-306 #comment Remove `npm prune` usage, it doesn't fit in automatically...

brev:
UNI-306 #comment Remove `npm prune` usage, it doesn't fit in automatically during our production build (https://docs.npmjs.com/cli/prune).

@lscheinkman @marionleborgne


lscheinkman:
:+1: LGTM 
@marionleborgne We should revisit this parameter once we are ready to optimize the application for size

PR #636, opened 2016-02-15
Fix send data

marionleborgne:
@brev fixed 'sendData'. The frontend receives data correctly form model runner now.


brev:
thanks @marionleborgne !

PR #648, opened 2016-02-19
All the pieces are connected

marionleborgne:
It works! @lscheinkman @brev  (except that the charts have a little issue and don't change amplitude with the anomaly likelihood values)


marionleborgne:
@lscheinkman the file upload is broken too actually. This doesn't show up in the left menu after upload. that might be due to the store merge.

brev:
@marionleborgne I'm looking into the bugs, in `unicorn-oobe`

:+1: :100:

PR #657, opened 2016-02-24
UNI-306 OOBE 2. Range selector anomaly bars.

brev:
UNI-306 #comment OOBE 2. Range selector anomaly bars.

cc @marionleborgne @lscheinkman @cbaranski


brev:
oops, forgot one thing, will re-file

brev:
@marionleborgne @lscheinkman this is ready for review now, thanks.

marionleborgne:
Nice work @brev. A couple of questions/comments:
1. If we have a file with a lot of records (say > 20,000) will we still see the anomaly bars indicators? i.e. is there a minimum width for the bars in the x-axis navigator?
* If not, we should add min bar width as constraint so that the anomaly indicators are readable even if there is a lot of data to display in the x-axis navigator. 
* We might want to consider to make the bar 1px bigger (like in the mockups). What are your thoughts on that?

2. I wasn't able to retrieve the same number of red markers in the highlighted area of the x-axis navigator and in the chart area (red bar charts). I am attaching an example where I have 2 red bars in the bar chart, but only one red marker. Let me know what you think of this.

Otherwise, it looks great :+1: 

screen shot 2016-02-25 at 11 50 44 am

brev:
Thanks @marionleborgne. All your questions are kind of about the same thing, let me detail my algorithm for you below. This should scale up cleanly on the x-axis navigator. 

With 20,000 datapoints, on our current ~900px wide charts, gives us ~22 datapoints per pixel. Of the 22 datapoints, the one with the highest anomaly score wins the 1 single pixel. Just a simple aggregation.

Our default sample file has ~5000 datapoints, which at 900px wide, is ~6 datapoints per pixel. A few color aggregation examples for this:
````
GGYRGR => R
GYGGGG => Y
GGGGGG => G
```

If your file has 900 datapoints to match the 900 pixels, you would see a 1:1 ratio above and below.

**About making the bar 1px bigger:**
I agree this would look better visually (like the mockups). BUT, we must consider the trade-off: if we do so, we cut the resolution described above in half, so that:  
```
20k datapoints => 44 datapoints per pixel
5k datapoints => 12 datapoints per pixel
450 datapoints => 1:1 ration above and below (i.e. blurry)
```

Sounds like you're pondering a more customized aggregation method, skewing the results a bit? This should probably be a new task, which needs more definition.

Does this answer your questions? Glad to discuss more. cc @lscheinkman

brev:
@marionleborgne I thought I saw a small off-by-one error, it may explain if you saw a missing red bar, I'm keeping an eye out, thanks

PR #658, opened 2016-02-25
Fix file upload

marionleborgne:
cc @lscheinkman @brev


marionleborgne:
this is based off the oobe-branch. now that file upload is fixed, it can be merged into master.

PR #651, opened 2016-02-19
UNI-326 Use unicorn_backend.date_time_utils.parseDatetime instead of datetime.strptime in param_finder_runner.py

vitaly-krugl:
CC @marionleborgne @ywcui1990 

- UNI-326 Use unicorn_backend.date_time_utils.parseDatetime instead of datetime.strptime in param_finder_runner.py
- Also fix several pylint warnings concerning unused imports and relative imports.

PR #662, opened 2016-02-26
python imports

marionleborgne:
@vitaly-krugl  @lscheinkman here is a proposed solution (for https://github.com/numenta/numenta-apps/pull/651) to avoid having relative imports while not introducing the need to alter the PYTHONPATH on the node side when we spawn the python process. 

I added a few lines to udpate the pythonpath dynamically at runtime. Let me know your thoughts on that.

cc @brev @oxtopus

### DIFF
--- a/unicorn/README.md
+++ b/unicorn/README.md
@@ -187,7 +187,6 @@ git clone https://github.com/numenta/numenta-apps
 cd numenta-apps/unicorn
 pip install -r py/requirements.txt
 npm install
-unset PYTHONPATH  # force usage of our custom internal Python
 ```
 
 
--- a/unicorn/py/unicorn_backend/model_runner_2.py
+++ b/unicorn/py/unicorn_backend/model_runner_2.py
@@ -41,7 +41,6 @@ import os
 import pkg_resources
 import sys
 import traceback
-
 import validictory
 
 from nupic.algorithms.anomaly_likelihood import AnomalyLikelihood
@@ -49,6 +48,10 @@ from nupic.data import aggregator
 from nupic.data import fieldmeta
 from nupic.data import record_stream
 from nupic.frameworks.opf.modelfactory import ModelFactory
+
+# Add numenta-apps/unicorn/py to the path
+py_dir = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+sys.path.append(py_dir)
 
 from unicorn_backend import date_time_utils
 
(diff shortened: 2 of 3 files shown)


unicorn/README.md
```diff
@@ -187,7 +187,6 @@ git clone https://github.com/numenta/numenta-apps
 cd numenta-apps/unicorn
 pip install -r py/requirements.txt
 npm install
-unset PYTHONPATH  # force usage of our custom internal Python
```
lscheinkman: Why are you removing the `unset PYTHONPATH` instructions?

vitaly-krugl:
@marionleborgne, this might work, but it's a hack that also alters how this code runs in other environments (such as test, etc.)

unicorn/README.md
marionleborgne: The documentation says:

```
sys.path

A list of strings that specifies the search path for modules. 
Initialized from the environment variable PYTHONPATH, plus an 
installation-dependent default.
```

So my thinking is the following: by inserting the path to `numenta-apps/unicorn/py` in first position in sys.path, python will look to import modules from `numenta-apps/unicorn/py` first (before the rest of the search paths in the sys.path list)

Because of that, I don't thing we have to unset the PYTHONPATH anymore, but I might be wrong. Please explain if my thinking is incorrect. @oxtopus you might know better.

vitaly-krugl: @marionleborgne, it becomes a hack on top of a hack :)

marionleborgne: @vitaly-krugl could you elaborate? what would be a better way to do all this? We want to avoid altering environment variables. The node process should be able to spawn the scripts in unicorn/py like this `<path_to_portable_python>/bin/python param_finder_runner.py`

vitaly-krugl: @marionleborgne: I am investigating.

vitaly-krugl:
@marionleborgne, see PR #663. I personally think that there is something wrong with the unicorn build or that we're missing a build step. `unicorn_backend` should be installed as a package, just like all other packages, and then nothing special should be necessary on our end at all. The fact that we're going to this trouble, suggests that unicorn_backend is not being installed (properly).

oxtopus:
This is something of a special case because if it were to be installed, it's not to be installed in any of the traditional locations, and the user will be shielded from the process itself.  My understanding is that the the node.js runtime will set the PYTHONPATH to something relative to the application bundle itself when it invokes the python subprocess.  The build and test phase should keep that in mind, and not necessarily the traditional installation technique.

vitaly-krugl:
@oxtopus: when I print `process.env.PYTHONPATH` from the two javascript modules that start the two python backend apps, the value is 'undefined'. In PR #663, I demonstrate that setting PYTHONPATH to a location where unicorn_backend may be found enables the existing unicorn_backend modules to load as-is.

lscheinkman:
Replaced by #667

PR #663, opened 2016-02-26
Specify unicorn_backend location via PYTHONPATH env var and launch python modules via -m option from javascript

vitaly-krugl:
@marionleborgne, would this work? Please try it out.

CC @oxtopus @brev

### DIFF
--- a/unicorn/js/main/ModelService.js
+++ b/unicorn/js/main/ModelService.js
@@ -21,14 +21,12 @@
 import getPortablePython from './PortablePython';
 import childProcess from 'child_process';
 import EventEmitter from 'events';
-import path from 'path';
 import system from 'os';
+
 import UserError from './UserError';
+import Utils from './Utils';
 
 const PYTHON_EXECUTABLE = getPortablePython();
-const MODEL_RUNNER_PATH = path.join(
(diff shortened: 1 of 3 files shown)


vitaly-krugl:
@oxtopus, this change in the environment variables with which the backend apps are launched appears to make absolute imports succeed in those python backend apps. However, the fact that this was necessary to get it to work might expose a flaw in the Unicorn build process: isn't the build supposed to get the unicorn_backend  in the same place as the other 3rd-party packages, such that its imports would work correctly without any additional magic???

oxtopus:
I believe this new change is the right approach and not indicative of a flaw in the build process, but specific to the way that the unicorn app will be deployed in practice.

vitaly-krugl:
I prototyped this solution, using duplicate code for setting up PYTHONPATH in each of the js modules. I would prefer to have this logic in a shared function or perhaps in another obvious place where it would only run once. Ideally, instead of modifying `process.env`, we would instead make a copy of `process.env` and make our PYTHONPATH changes in the copy that is used for spawining the python apps. If someone can give me some pointers about how and where to do this, I would be glad to give it a try.

lscheinkman:
Replaced by #667

PR #664, opened 2016-02-26
UNI-306 #comment J. Lock range selector width, always 100 points now.

brev:
UNI-306 #comment J. 
- Lock range selector width, always 100 points now.
- Remove js-csp
- Fix a small off-by-one pixel bug in range selector anomaly bars, thanks for review @marionleborgne.

ready for review @marionleborgne @lscheinkman

### DIFF
--- a/unicorn/js/browser/components/ModelData.jsx
+++ b/unicorn/js/browser/components/ModelData.jsx
@@ -60,6 +60,7 @@ export default class ModelData extends React.Component {
     this._chartOptions = {
       // dygraphs global chart options
       options: {
+        clickCallback: this._chartHandleClick,
         plugins: [RangeSelectorBarChart],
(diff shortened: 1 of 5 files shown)


unicorn/js/browser/components/ModelData.jsx
```diff
@@ -151,6 +155,10 @@ export default class ModelData extends React.Component {
     context.stroke();
   }
 
+  _chartHandleClick(...args) {
+    // console.log('!!! click', args); // eslint-disable-line
+  }
+
```
lscheinkman: No sure why this callback is necessary

lscheinkman:
:+1: LGTM

unicorn/js/browser/components/ModelData.jsx
brev: thanks @lscheinkman I'll remove

PR #665, opened 2016-02-26
Convert services into singletons created in the `main` process only

lscheinkman:
Convert `main` process services into singletons created in the `main` process and accessed via `remote` by the `renderer` process.

@marionleborgne, @brev  Please review


brev:
:+1: thanks

PR #672, opened 2016-02-29
Update electron to version 0.36.9 with remote ES6 classes support

lscheinkman:
@brev electron v0.36.9 added ES6 classes support to `remote`. Now we dont have to mix ES5/ES6 classes.
This version also improves `remote` objects performance
CC: @marionleborgne


brev:
:+1:

PR #675, opened 2016-03-01
disable pip cache usage

marionleborgne:
cc @lscheinkman @vitaly-krugl


lscheinkman:
@marionleborgne The build issue is related to the `PYTHONPATH` on the build machine. See https://ci.numenta.com/download/UN-UN-BOX/build_logs/UN-UN-BOX-49.log

marionleborgne:
For posterity: after a conversation with @lscheinkman it has been agreed to leave the `--no-cache-dir` flag since it may be helpful for developers (if their cache is not empty)

PR #676, opened 2016-03-01
UNI-329: Load model and metric data from db instead of file

lscheinkman:
@marionleborgne @brev 
Sorry for the large PR but you can review each commit separate as they build up to saving and loading metric data and model data to the database instead of the file system.


brev:
:+1: @lscheinkman When you merge in `master`, there will be 4 files with conflicts. You can see resolved versions for help here:

1. https://github.com/brev/numenta-apps/blob/load_from_db_brev_revs/unicorn/js/browser/stores/ModelStore.js
1. https://github.com/brev/numenta-apps/blob/load_from_db_brev_revs/unicorn/js/database/schema/MetricData.json
1. https://github.com/brev/numenta-apps/blob/load_from_db_brev_revs/unicorn/js/database/schema/ModelData.json
1. https://github.com/brev/numenta-apps/blob/load_from_db_brev_revs/unicorn/tests/js/unit/services/DatabaseService_test.js

lscheinkman:
@brev Thanks

marionleborgne:
works great for me too :+1: 

One thing to note (but likely unrelated to this specific PR) is that the UI is still struggling to show multiple running models at the same time (UI freezes when more thant one model is running). @lscheinkman you mentioned additional performance improvements earlier, do you think this enhance the performances of having multiple models running concurrently?

unicorn/js/browser/actions/DeleteFile.js
```diff
@@ -33,14 +32,13 @@ export default function (actionContext, filename) {
   return new Promise((resolve, reject) => {
     let database = actionContext.getDatabaseClient();
     // Delete file and its data
-    let fileId = Utils.generateFileId(filename);
-    database.deleteFile(fileId, (error) => {
+    database.deleteFile(filename, (error) => {
```
marionleborgne: Are we using filenames as IDs now? If so, why do we have the `Utils.generateFileId(filename)` methods?

lscheinkman: No, `DeleteFile` will call `Utils.generateFileId` from the `main` process saving the extra `remote` call to `crypto` API

lscheinkman:
@marionleborgne I dont think this affects the UI freezing issues you are having while multiple models are running. This PR only relates to database access and store. The active models are sending data to the UI via the same mechanism as before. The previously saved models will notice the performance improvement because now the results come from the database

PR #677, opened 2016-03-01
UNI-306 FLOW A. Pass correct aggregation params to ModelRunner

brev:
UNI-306 FLOW A. Pass correct aggregation params to ModelRunner
cc @lscheinkman @marionleborgne


brev:
Wasn't able to get the visual part of this done and into the flow-testing branch before todays tests. So, now into master. I'll put the visual part into master when done.

PR #678, opened 2016-03-02
Fix installation using new ci build

lscheinkman:
@marionleborgne, @brev 

This should fix the installation issues we had with `npm install` for `portable_python`.

CC: @oxtopus


brev:
:+1:

PR #684, opened 2016-03-08
Restored and refactored docker configuration for taurus and taurus.metric_collectors

oxtopus:


oxtopus:
cc @scottpurdy @vitaly-krugl @lscheinkman

taurus.metric_collectors/docker/configure.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have purchased from
```
scottpurdy: nit: year

taurus.metric_collectors/docker/run.sh
```diff
@@ -0,0 +1,31 @@
...
+#
+
+# Be extra verbose and sensitive to failures
+set -o errexit
+set -o pipefail
+set -o verbose
+set -o xtrace
+set -o nounset
+
+supervisord -c conf/supervisord.conf --nodaemon
```
scottpurdy: Why put this in a shell script instead of directly running `supervisord` in the Dockerfile? The shell scripts seem to unnecessarily complicate the setup.

taurus/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+# Environment
+ENV APPLICATION_CONFIG_PATH=/opt/numenta/taurus/conf \
+    MYSQL_HOST=mysql \
+    MYSQL_USER=root \
+    MYSQL_PASSWD="" \
+    RABBITMQ_HOST=rabbit \
+    RABBITMQ_USER=guest \
+    RABBITMQ_PASSWD=guest \
+    TAURUS_RMQ_METRIC_DEST=rabbit \
+    TAURUS_RMQ_METRIC_PREFIX=CHANGEME \
```
scottpurdy: At what point do these get set correctly?

taurus.metric_collectors/docker/run.sh
oxtopus: For consistency with `taurus/docker/run.sh` in which there is more than the one command.

scottpurdy:
I put a few initial questions.

taurus/Dockerfile
oxtopus: `docker run` with `-e` args.

taurus.metric_collectors/docker/run.sh
scottpurdy: Is that an issue because Docker can only run a single thing? So you have it run a shell script that launches multiple?

oxtopus: Something else to consider: this is used in conjunction with `entrypoint.sh`.  `Dockerfile` specifies both `ENTRYPOINT` and `CMD`.

oxtopus: > Is that an issue because Docker can only run a single thing? So you have it run a shell script that launches multiple?

Yes.  I think it's cleaner this way than polluting the `Dockerfile`.

oxtopus:
I've resolved the copyright dates.  How important is it that `supervisord -c conf/supervisord.conf --nodaemon` be added back to `Dockerfile` instead of in `run.sh`?  That's how I originally had it, but ended up putting it into `run.sh` when I realized I was going to have to for `taurus`.

Personally, I prefer the symmetry, but am happy to switch back to not using `run.sh` to get this merged.

scottpurdy:
This is fine :+1: 

It seems pretty complex to navigate but I'm not sure if there is any way to improve that. It would also be nice to define the public interface of the numenta/nupic Docker image (what can you expect to be installed, etc) but that is outside the scope of this change.

PR #687, opened 2016-03-10
UNI-348: Show param finder progress for ay least 3 seconds

lscheinkman:
@marionleborgne let me know if 3 secs is enough time


marionleborgne:
@lscheinkman :+1: 3s works well. It's just enough to be able to read the message.

PR #689, opened 2016-03-11
Refactor utils separating `browser` process from `main` process functions

lscheinkman:
CC: @marionleborgne @brev


brev:
:+1: very nice

PR #690, opened 2016-03-11
Only updates ModelData component if the model data has changed

lscheinkman:
@brev @marionleborgne This should limit some of the UI updates when showing multiple charts


brev:
:+1:

PR #691, opened 2016-03-11
UNI-352: Persist model to database

lscheinkman:
@marionleborgne @brev This PR should fix the model action buttons and left menu icons


marionleborgne:
@lscheinkman works for me! :+1:

PR #695, opened 2016-03-15
Reduce taurus nginx and uwsgi listen queue sizes to 128 to match the default net.core.somaxconn on linux.

vitaly-krugl:
CC @oxtopus 

This PR is in response to nginx failure issue reported on nupic@lists.numenta.org by vannroath


oxtopus:
:+1:

vitaly-krugl:
@oxtopus, I updated the UWSGI listen queue size to 128 as well.

PR #700, opened 2016-03-17
WIP! UNI-328 #comment Modify charts to hold position state between visibility

brev:
UNI-328 #comment Modify charts to hold position state between visibility
and show-non-agg toggles. 

~~Starting with this change, Models are no longer rendered only on `model.visible`. Now, Models are _always_ rendered, and the CSS `display` property is used to show/hide the Model based on `model.visible`.~~

WIP! Under Construction.


brev:
@lscheinkman @marionleborgne please review anytime, thanks.

brev:
@lscheinkman I'll switch back to rendering only on visible, thanks for the feedback - shouldn't be too hard. I'll update soon.  cc @marionleborgne

unicorn/js/database/schema/Metric.json
```diff
@@ -34,6 +34,11 @@
       "type:": "integer",
       "description": "Zero based Column index"
     },
+    "viewpoint": {
```
lscheinkman: @brev This property is never read, why do we need to keep it in the store?

lscheinkman:
@brev It works great. My only comment is about `metric.viewpoint` property and all the code to update it (actions, store, schema, ...). 
This property is never read. I'm not sure why do we need it.

brev:
Sorry, this is getting confusing, I'll re-open a fresh PR soon. :) thanks

unicorn/js/database/schema/Metric.json
brev: sorry, it is now, check the latest commit :)  New PR will make more sense, coming soon. thanks.

PR #711, opened 2016-03-21
UNI-377: Bypass aggregation dialog if param_finder does not find any valid aggregation

lscheinkman:
@marionleborgne 
This PR will bypass second dialog if param_finder does not find any valid aggregation and start the model with the `raw` data by default.
You can test with small files.


marionleborgne:
LGTM :+1: thanks @lscheinkman

PR #715, opened 2016-03-22
UNI-378: Make delete file dialog buttons consistent with other dialogs

lscheinkman:
@marionleborgne I think this is the last dialog with the wrong button.


brev:
:+1:

marionleborgne:
Thanks @lscheinkman. All the dialogs look good now.

PR #719, opened 2016-03-23
UNI-351: Update model status icon and tooltip

lscheinkman:
@marionleborgne , @brev 
Add both spinner (UNI-351) and tooltip (UNI-359) to model status indicator


brev:
:+1:

PR #724, opened 2016-03-31
UNI-284: Stop and hide models when deleting files

lscheinkman:
@marionleborgne It should also fix UNI-381


marionleborgne:
LGTM :+1: It's also fixing the bug where clicking on "delete file" while the chart card is being displayed returns a null pointer exception. Looks like it's gone now :100:

PR #726, opened 2016-04-12
TAUR-1209 Remove any pretense of support for namespace packages.  The…

oxtopus:
…re is now explicitly `taurus.metric_collectors` and `taurus.engine`.

Required to get current pipeline refactoring in a functional state.


vitaly-krugl:
@oxtopus, could you please explain the reasoning behind this change? I would like to learn what problem existed with our attempted support for namespace packages in Taurus, and why namespace packages no longer make sense for Taurus. Thanks!

vitaly-krugl:
LGTM

PR #730, opened 2016-04-19
UNI-399: fix python tests

marionleborgne:
@lscheinkman `npm run test:backend` should now run all python tests (integration and unit) with the portable python distribution


marionleborgne:
FYI, I'm now switching to UNI-401 to remove the dependency on `nta.utils` and MySQL. I'll create a separate PR for that.

unicorn/package.json
```diff
@@ -52,8 +52,11 @@
     "test:coverage": "isparta cover --include '**/*.jsx' --include '**/*.js' --include-all-sources  --exclude **/bundle.js _mocha -- --opts tests/js/mocha.opts tests/js/{unit,integration}",
     "test:pipeline:coverage": "isparta cover --report clover --report html --include '**/*.jsx' --include '**/*.js' --include-all-sources -x **/bundle.* _mocha -- --opts tests/js/mocha.opts tests/js/{unit,integration}",
     "posttest:coverage": "open-url file://${PWD}/coverage/lcov-report/index.html",
-    "pretest:integration:backend": "mkdir -p test-reports && python=$PWD/$npm_package_python && cd ../nta.utils/ && $python -m pip install -r requirements.txt && $python setup.py install",
-    "test:integration:backend": "$npm_package_python -m pytest tests/py/integration/compatibility_test/model_runner_2_compatibility_test.py --junitxml test-reports/model_runner_2_compatibility_test.xml",
+    "test:backend": "npm run test:backend:unit && npm run test:backend:integration",
```
lscheinkman: @marionleborgne `pretest:backend:integration` and `pretest:backend:unit` should be replaced by `pretest:backend` otherwise it will execute twice when you call `test:backend`

lscheinkman:
👍  It worked for me. Just one small comment we can address later

unicorn/package.json
marionleborgne: @lscheinkman yes :) it will run twice. but this way you can run only the unit tests or only the integration tests.

PR #731, opened 2016-04-19
UNI-401: remove nta.utils and MySQL dependency

marionleborgne:
@lscheinkman to test this run:

* cleanup
```
npm run clean && npm install
```

* run tests

```
npm run test:backend
```


vitaly-krugl:
@marionleborgne, why is this PR creating copies of code. This adds to our maintenance burden. We should be sharing, not copying.

marionleborgne:
@vitaly-krugl I would keep the dependency on `nta.utils` and import the code from there... however to install `nta.utils` we need to install mysql, and Unicorn does not need it. It seems superfluous. What are you thoughts on that?

marionleborgne:
@lscheinkman @vitaly-krugl as discussed yesterday, I removed the heavy logging (with templating and other unnecessary features for the unicorn tests) and replaced it by the vanilla python logging. Log messages are still written to stderr, like what we had before when we were using:

```
from nta.utils.logging_support_raw import LoggingSupport

def setUpModule():
      LoggingSupport.initTestApp()
```

Let me know if you have any additional feedback. Thanks.

lscheinkman:
👍 LGTM

marionleborgne:
@vitaly-krugl Please have a look when you have time. Thanks

vitaly-krugl:
@marionleborgne, LGTM, thanks!

PR #734, opened 2016-04-22
updated text and buttons in snackbar

marionleborgne:
Incorporating Jeff's feedback regarding the snackbar.


marionleborgne:
@lscheinkman please review when you have time. Thanks

unicorn/app/config/default.json
```diff
@@ -52,8 +52,7 @@
   },
   "snackbar": {
     "completed": {
-      "message": "HTM Model completed for %s",
-      "action" : "Results"
```
lscheinkman: You should also remove the action handler code from the snackbar.
https://github.com/marionleborgne/numenta-apps/blob/snackbar-with-jeff-feedback/unicorn/app/browser/components/Model.jsx#L476-L477

marionleborgne:
Thanks @lscheinkman. I made the update.

lscheinkman:
👍 LGTM

PR #736, opened 2016-04-25
UNI-389: Update anomaly text and all related issues

lscheinkman:
@marionleborgne This PR updates the anomaly text to 'HIGH/MEDIUM/LOW' (UNI-389) and all other related issues:
 - UNI-388: CSV Export
 - UNI-418: Chart Legend Style

Please review

### DIFF
--- a/unicorn/app/browser/components/Model.jsx
+++ b/unicorn/app/browser/components/Model.jsx
@@ -102,15 +102,15 @@ export default class Model extends React.Component {
         width: '100%'
       },
       cardHeader: {
-        paddingBottom: 0
+        paddingBottom: 0,
+        height: '3rem'
       },
       cardText: {
         paddingTop: 0
       },
       cardHeaderText: {
         display: 'inline-flex',
-        verticalAlign: 'middle',
-        height: '1.5rem'
+        verticalAlign: 'middle'
       },
       title: {
         fontSize: 14,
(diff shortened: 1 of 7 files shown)


marionleborgne:
Couple of comments:
* The dot indicating the position on the chart is gone when you have model results
screen shot 2016-04-25 at 5 56 15 pm

* The legend text for metric value / anomaly / non-aggregated is in bold, except for the word "Anomaly". I think it's best to make them consistent (either everything in bold or nothing). How about we try it with _nothing_ in bold, since the legend is a bit crowded already? i.e. "Value: 0.3", "Anomaly: LOW", "Non-aggregated: 9.0" all in regular font style (not bold)

legend

lscheinkman:
@marionleborgne
1) The dot is not missing, can you double check or let me know which dataset did you use?

2) The "Value" will always be bold unless we update dygraph. See https://github.com/danvk/dygraphs/blob/v1.1.1/plugins/legend.js#L285-L286
For now I will make all labels bold and defer the plain version to a later stage when we update dygraph.

marionleborgne:
@lscheinkman Ok great, let's keep everything in bold. As long as it is consistent. 

Regarding the dot, I am actually unable to reproduce the bug that I had yesterday (where the dot was missing). I'll keep an eye on it to see if I  come across it again.

LGTM 👍

PR #737, opened 2016-04-25
Skip prompt utils tests in non-tty environments such as docker-based …

oxtopus:
…bamboo ci environment

### DIFF
--- a/nta.utils/tests/unit/prompt_utils_test.py
+++ b/nta.utils/tests/unit/prompt_utils_test.py
@@ -23,7 +23,7 @@
 Unit tests for nta.utils.prompt_utils
 """
 
-
+import sys
 import time
 import unittest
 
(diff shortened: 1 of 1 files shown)


oxtopus:
fyi @vitaly-krugl

nta.utils/tests/unit/prompt_utils_test.py
```diff
@@ -36,6 +36,8 @@
 class PromptUtilsTestCase(unittest.TestCase):
 
 
+  @unittest.skipIf(sys.stdout.isatty(),
+                   "PromptUtilsTestCase not tested within tty.")
```
vitaly-krugl: Recommend changing the messages to "PromptUtilsTestCase requires tty". Also, you should be able to get the desired effect by placing a single `@unittest.skipIf` decorator on top of `PromptUtilsTestCase` class, instead of a separate one on each of the test methods.

PR #740, opened 2016-04-25
Fix models with no aggregation

lscheinkman:
@marionleborgne Fix problem when the model fails to complete when not using aggregations.
Test using 'home_data.csv' sample file


marionleborgne:
👍 it works @lscheinkman. Thanks!

PR #746, opened 2016-04-28
Fix failure of XigniteStockAgentTestCase.testForwardAfterHours unit test when executed while metric collectors are in hot_standby mode

vitaly-krugl:
CC @scottpurdy @oxtopus 

This happened after the graceful-update-testing-taurus-servers pipeline ran and failed, leaving the metric collectors in hot standby mode.

### DIFF
--- a/taurus.metric_collectors/tests/unit/xignite_stock_agent_test.py
+++ b/taurus.metric_collectors/tests/unit/xignite_stock_agent_test.py
@@ -30,6 +30,7 @@ import urlparse
 
 import pytz
 
+from taurus import metric_collectors
 from taurus.metric_collectors import logging_support
 from taurus.metric_collectors.xignite import xignite_stock_agent
 
@@ -524,7 +525,9 @@ class XigniteStockAgentTestCase(unittest.TestCase):
       stockExchange="NASDAQ",
       sampleKey="Volume")
 
-    xignite_stock_agent.forward((msft,), data, security)
+    with patch.object(xignite_stock_agent, "g_opMode",
+                      new=metric_collectors.ApplicationConfig.OP_MODE_ACTIVE):
+      xignite_stock_agent.forward((msft,), data, security)
 
     self.assertEqual(
       collectorsdb.engineFactory.return_value.execute.call_count,
@@ -654,3 +657,8 @@ class XigniteStockAgentTestCase(unittest.TestCase):
        call(metricName="XIGNITE.MSFT.VOLUME",
             value=492621,
             epochTimestamp=1421332500.0)])
+
+
+
+if __name__ == "__main__":
+  unittest.main()


taurus.metric_collectors/tests/unit/xignite_stock_agent_test.py
```diff
@@ -30,6 +30,7 @@
 
 import pytz
 
+from taurus import metric_collectors
```
oxtopus: This _may_ be problematic.  We decided to drop the namespace packages with the expectation that `taurus.metric_collectors` and `taurus.engine` are distinct.  To avoid further confusion, this should be `import taurus.metric_collectors`.

vitaly-krugl: @oxtopus, I think that a good time to make this change will be when taurus/metric_collectors is renamed to taurus_metric_collectors. It's too error-prone as it stands right now.

CC @scottpurdy

oxtopus: I'm unaware of any such plans, and broad renaming is almost certainly more risky and involved than a one line change to `import taurus.metric_collectors` in this one case to be consistent with the status quo.

vitaly-krugl: All of these are working in the current repo:

```
In [2]: import taurus
In [3]: from taurus import metric_collectors
In [4]: import taurus.metric_collectors
```

Also:

```
$ git grep 'from taurus im'
taurus.metric_collectors/taurus/metric_collectors/common_services/metric_maintenance_agent.py:from taurus import metric_collectors
taurus.metric_collectors/taurus/metric_collectors/metric_utils.py:from taurus import metric_collectors
taurus.metric_collectors/taurus/metric_collectors/set_collectors_opmode.py:from taurus import metric_collectors
taurus.metric_collectors/tests/unit/metric_maintenance_agent_test.py:from taurus import metric_collectors
taurus.metric_collectors/tests/unit/xignite_stock_agent_test.py:from taurus import metric_collectors
```

My suspicion is that the task that eliminated the namespaces is incomplete.

scottpurdy: @oxtopus can you capture whatever work is left to be done and send me the JIRAs for it? @vitaly-krugl probably better to remove this line and add `ApplicationConfig` to the import statement on the following line.

oxtopus: I took care of this one case in https://github.com/numenta/numenta-apps/pull/749 

Up to you and @vitaly-krugl if we should retroactively fix the areas that fell through the review cracks.  As pointed out, it's only a few files and, it's a trivial change.  However, IMO, it's not a huge priority.

PR #748, opened 2016-04-28
TAUR-1895 Added diagnostic logging to capture xignite error messages from failed self-check.

vitaly-krugl:
CC @scottpurdy


scottpurdy:
👍

PR #751, opened 2016-04-30
UNI-334: Fix daylight savings. Assume UTC.

mrcslws:
When we add a file to the database, we currently parse the string as if it's in the current time zone. Then we save this parsed timestamp's Unix time. For nonexistent times like 2015-03-08 02:30:00 GMT-0800, we save the UTC Unix time for 2015-03-08 03:30:00 GMT-0800. So multiple data points are saved for this timestamp, because the actual 3:30am datum is also parsed as 3:30am.

The chart retrieves this Unix time from the database and formats it as the current time zone, then displays it. So the Unix time remains constant throughout this whole process, it just matters how we parse the initial string.

### DIFF
--- a/unicorn/app/main/DatabaseService.js
+++ b/unicorn/app/main/DatabaseService.js
@@ -81,6 +81,29 @@ function stringifyResultsCallback(callback) {
   return (error, data) => callback(error, JSON.stringify(data));
 }
 
+/**
+ * Parse the timestamp as if it doesn't contain a time zone.
+ *
+ * This is a way of having "naive" datetimes, as in Python. JavaScript condenses
+ * the day/hour with the timezone by simply storing the UTC timestamp. We want
+ * to preserve the parsed day/hour -- it's what we render -- so we preserve it
+ * by throwing out time zone info.
+ *
+ * @param {string} timestamp a timestamp string intended for moment.js
+ * @param {string} format a format intended for moment.js
+ * @return {Date} a JavaScript date corresponding to the timestamp's time in
+ *                UTC, ignoring the timestamp's specified time zone.
+ */
+function parseTimestampIgnoreZone(timestamp, format) {
+  if (format.indexOf('Z') === -1) {
+    return moment.utc(timestamp, format).valueOf();
+  }
+
+  let noZoneFormat = 'YYYY-MM-DDTHH:mm:ss.ssss';
+  let strippedZone = moment.parseZone(timestamp, format).format(noZoneFormat);
+  return moment.utc(strippedZone, noZoneFormat).valueOf();
+}
+
 
 /**
  * HTM Studio: DatabaseService - Respond to a DatabaseClient over IPC.
@@ -555,7 +578,10 @@ export class DatabaseService {
     })
     .on('data', (result) => {
       let data = {
-        timestamp: moment(result.timestamp).toDate(),
+        // Don't include time zone. The model runner doesn't send time zone
+        // info, and the time zone may not be UTC.
+        timestamp: moment.utc(result.timestamp)
+              .format('YYYY-MM-DDTHH:mm:ss.ssss'),
         metric_value: result.metric_value,
         anomaly_level: mapAnomalyText(result.anomaly_score),
         raw_anomaly_score: result.anomaly_score
@@ -902,8 +928,8 @@ export class DatabaseService {
               if (field.type === 'number') {
                 let metricData = {
                   metric_uid: field.uid,
-                  timestamp: moment.utc(data[timestampField.index],
-                                        timestampField.format).valueOf(), // eslint-disable-line
+                  timestamp: parseTimestampIgnoreZone(
+                    data[timestampField.index], timestampField.format),
                   metric_value: parseFloat(data[field.index])
                 };
                 // Save data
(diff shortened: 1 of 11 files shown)


mrcslws:
@marionleborgne 

Note that this shifts all data sets. For example, hotgym is now busy between the hours of 1am and 9am. And the NYC Taxis are busy between 12am and 4pm.

Thoughts? Solutions that come to mind:

- Never translate the time zone. Render whatever they give us. Of course that means everyone with UTC logs has to read UTC timestamps.
- Use the current time zone, but parse as "standard time", never "daylight time", assuming the timestamps have been somehow doctored to be unaffected by daylight savings. I'm not quite sure how this would work. There's ambiguity in how the parsing would work, and in what UI we'd show (would we still depict DST?)
- Allow configuration when adding a file.
- Allow configuration when displaying a chart.

We can always doctor the sample data, though it's a little weird that the sample data will be shifted based on your time zone.

mrcslws:
Another solution:

- Parse as UTC only if the timestamp ends with a 'Z', as specified in https://en.wikipedia.org/wiki/ISO_8601#UTC . But this wouldn't immediately solve the current issue. Our csv files that run into the DST issue will still hit the issue until we update the csv files to use the 'Z'.

mrcslws:
With that last solution I don't even think we'd need this code change. moment.js detects the 'Z' and jumps into UTC mode. We'd just have to change our csv files to use 'Z'. When there's no 'Z', it's a little unclear what we'd do about DST. We could require the csv to jump directly from 1:30am to 3:30am in March, and make sure we know how to handle the multiple 1:30am data points in November? Or require their csv to subtract the DST hour?

We could choose to not care about these DST stutters in this non-Z case. We'll change our sample data to use 'Z', and we'll recommend using UTC (and 'Z') for the best experience.

mrcslws:
Alright, I slept on it. I think there are two fixes that will work.

## Option 1: Recommend UTC timestamps (with a 'Z')

This involves no code change. We'll change the sample data to use "2015-03-08 02:02:53Z" style timestamps.

If there's no 'Z', we parse the timestamp in the current time zone, and it will cause occasional Daylight Savings hiccups. We'll decide that's okay -- if people want perfection, they should use UTC timestamps.

#### The downside

The sample data will appear different based on which time zone you're viewing from. The NYC Taxi data will only look right if you live in NYC's time zone.

A fix for this would be to make the chart's time zone configurable. But that's a whole feature.

## Option 2: Render whatever timestamps they give us

If the CSV file says, 2:02am, we show the data point at 2:02am, no questions asked. HTM Studio will appear oblivious to time zones. It's a simple CSV viewer and anomaly detector. The UI never depicts Daylight Savings Time -- it doesn't skip 2am on March 8, and it doesn't repeat 1am on November 2nd.

Underneath, we'll accomplish this by parsing everything as UTC (i.e. my current change in this PR), and change the charts to render in UTC. But this is just implementation details.

mrcslws:
I discussed with @marionleborgne. We both like Option 2.

I just pushed the chart change. It works. It displays the timestamps from the file, and Daylight Savings is no longer depicted in the chart. 1:30am and 2:30am exist exactly once per day. 🕜 🕝 💯 📅

marionleborgne:
Thanks for the PR @mrcslws. I tested it on the NYC taxi data (one of the sample files in the electron app) and I still see the gap event if:
* I clean the DB 
* Run the analysis with or without aggregation

With aggregation:
screen shot 2016-04-30 at 12 53 41 pm

Without aggregation:
screen shot 2016-04-30 at 4 20 10 pm


Original CSV file:
screen shot 2016-04-30 at 2 55 08 pm

Results (without aggregation)
screen shot 2016-04-30 at 2 55 24 pm


Could you double check with the NYC data on your side and see if you can reproduce these results?

marionleborgne:
PR for https://jira.numenta.com/browse/UNI-334

mrcslws:
Ah, I never knew there was also an issue with the anomaly bars. I'll investigate. Thanks for testing it.

mrcslws:
Current status: I have a fix for the anomaly bars issue, but I'm thinking more about the initial solution. Simply calling `moment.utc` everywhere doesn't work when the timestamp explicitly identifies a non-UTC timezone. Example:

~~~javascript
// No timezone
moment.utc('2013-03-10 02:30', 'YYYY-MM-DD HH:mm').format()
"2013-03-10T02:30:00+00:00"

// Explicitly UTC (via 'Z')
moment.utc('2013-03-10 02:30Z', 'YYYY-MM-DD HH:mmZ').format()
"2013-03-10T02:30:00+00:00"

// Explicitly non-UTC
moment.utc('2013-03-10 02:30+0400', 'YYYY-MM-DD HH:mmZ').format()
"2013-03-09T22:30:00+00:00"
~~~

So when there's an explicit timezone we'll show the timestamps in UTC, which isn't good. e.g. in this case, the user's "2013-03-10 02:30+0400" will be displayed as 10:30pm. That's not what we want.

I'm still pondering.

unicorn/app/main/DatabaseService.js
```diff
@@ -555,7 +578,10 @@ export class DatabaseService {
     })
     .on('data', (result) => {
       let data = {
-        timestamp: moment(result.timestamp).toDate(),
+        // Don't include time zone. The model runner doesn't send time zone
+        // info, and the time zone may not be UTC.
+        timestamp: moment.utc(result.timestamp)
+              .format('YYYY-MM-DDTHH:mm:ss.ssss'),
```
mrcslws: @marionleborgne Do you have a preferred number of `s`s for this format? Specifically, the milliseconds.

lscheinkman:
👍 LGTM

PR #752, opened 2016-05-01
UNI-386: Make sidebar scrollable. Polish the scrolling experience.

mrcslws:
The experience of scrolling the sidebar is now identical to the experience of scrolling the charts.

The changes:

- You can now scroll the sidebar. It scrolls independently of the chart area.
- The scrollbar is positioned correctly. I had to work around a Chromium bug.
- The scrollbars are the same color. Previously the scrollbar in the charts area
  was white because the operating system chose that color based on the
  background color. The background color is a gradient, which causes the OS to
  be weird and choose white (even though it chooses black if you use either
  gradient endpoint as a solid-colored background). Now that the scrollbar lives in
  a child element, this no longer happens. This is all a little bit goofy.

Additionally, the app no longer bounces up and down when you scroll to the top and bottom (like a webpage). Now it feels more like a real desktop app, similar to Atom. The bounces also had the issue of showing weird rendering artifacts sometimes.

### DIFF
--- a/unicorn/app/browser/assets/styles/index.css
+++ b/unicorn/app/browser/assets/styles/index.css
@@ -24,7 +24,10 @@
 
 html {
   box-sizing: border-box;
+  /* Avoid the scroll top/bottom bounce effect. Don't let the html tag scroll. */
+  width: 100%;
   height: 100%;
+  overflow: hidden;
 }
 
 *, *:before, *:after {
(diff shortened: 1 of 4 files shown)


marionleborgne:
LGTM 👍 Thanks @mrcslws 

One unit test does not pass, but it's independent from your changes and I'll fix it in a separate PR.

PR #757, opened 2016-05-04
UNI-434 Fix unwanted logging message on stderr in unicorn model_runner_2

vitaly-krugl:
1. Add logging.NullHandler to the root logger in model_runner_2 to make sure that python logging won't implicitly add the default StreamHandler when some code logs via logging.debug, logging.info, etc.
2. Un-skip integration test `testModelRunnerFailsWithInvalidInputRecord`.
3. Fix several pylint warnings


vitaly-krugl:
CC @marionleborgne @lscheinkman

marionleborgne:
Thanks @vitaly-krugl. 

LGTM 👍

PR #754, opened 2016-05-02
UNI-428 Add advanced settings

flysonic10:
@marionleborgne 

Refactor text to use translations config
Add advanced settings to the Create Model Dialog

PR #758, opened 2016-05-05
UNI-437 Fix bug that causes multiple CreateModelDialog windows to open

flysonic10:
Local open/close state was implemented in `CreateModelStore`, causing all dialogs to respond to show/hide actions. I moved the open/close state definition for `CreateModelDialog` to the parent component `Model`.

Additionally, this surface another issue where `_startModel` was being called in the render function. This caused an infinite loop when a state change was added to `_startModel`. Generally, components should not fire actions in the rendering process. I was forced to fix this by adding the `_startModel` action to a button. This is matches @marionleborgne's intended behavior for the dialog and will be extended further in UNI-428 (#754).

Merging this is a pre-requisite to merging #754 

One test failure `1) should export ModelData from the database` matches the test failure already on master.

### DIFF
--- a/unicorn/app/browser/components/CreateModelDialog.jsx
+++ b/unicorn/app/browser/components/CreateModelDialog.jsx
@@ -25,7 +25,6 @@ import React from 'react';
 
 import ChartUpdateViewpoint from '../actions/ChartUpdateViewpoint';
 import CreateModelStore from '../stores/CreateModelStore';
-import HideCreateModelDialogAction from '../actions/HideCreateModelDialog';
 import StartModelAction from '../actions/StartModel';
 import {trims} from '../../common/common-utils';
 
@@ -38,7 +37,6 @@ import {trims} from '../../common/common-utils';
   inputOpts: context.getStore(CreateModelStore).inputOpts,
   metricId: context.getStore(CreateModelStore).metricId,
   metricName: context.getStore(CreateModelStore).metricName,
-  open: context.getStore(CreateModelStore).open,
   paramFinderResults: context.getStore(CreateModelStore).paramFinderResults
 }))
 export default class CreateModelDialog extends React.Component {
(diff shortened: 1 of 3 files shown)


marionleborgne:
Thanks @flysonic10, good catch. FYI @lscheinkman, I documented the bug here: https://jira.numenta.com/browse/UNI-437. Reviewing now.

marionleborgne:
LGTM 👍

marionleborgne:
@flysonic10  You need to fix the linting. Run `npm run lint`

unicorn/app/browser/components/CreateModelDialog.jsx
```diff
@@ -138,8 +137,21 @@ export default class CreateModelDialog extends React.Component {
             />
         );
       } else {
-        // No aggregation required, just start the model
-        this._startModel(rawPayload);
+        body = (
```
marionleborgne: Why did you bring back this dialog in this PR? I understand the use case for your other PR with the advanced settings. But I don't think this is needed here. You can just skip and start the model.

marionleborgne:
@flysonic10 With the exception of the intermediary dialog that you added when there is no aggregation needed, this PR looks good to me 👍 Nice catch with this bug.

flysonic10:
The intermediary dialog is required because otherwise the component is breaking the fundamental role of the render function which should not include new state changes.

This can alternatively be refactored out of the render function, but it would be wasted effort given that the advanced settings PR removes this completely.

unicorn/app/browser/components/CreateModelDialog.jsx
flysonic10: See below. We can't change state in the render function, so this ultimately results in a render loop.

marionleborgne:
OK, good point. Then can you pull this branch into the advanced settings PR (#754) so this gets merged when we merge PR #754 ?

marionleborgne:
Ok you already rebased #754 to this branch. All good then. Let's wait until #754 is stable in term of features and I can get the last round of feedback from Jeff. After that we can decide to merge #754 or to just work off this PR.

marionleborgne:
Closing this. It's a stale PR and the bug fix changes are already contained in #754

PR #759, opened 2016-05-05
UNI-398 Highlight probationary period

oxtopus:


marionleborgne:
Thanks @oxtopus. 

I only have one comment, although this requirement might not have been clear in the original JIRA: can you also reflect the highlighted area in the time navigation bar at the bottom? I'll make this more clear in the JIRA and will include a screenshot.

👍 otherwise.

PR #760, opened 2016-05-05
Resolves UNI-415, giving columns in the data preview table a minimum

scottpurdy:
width and adding a horizontal scrollbar when necessary. See [UNI-415](https://jira.numenta.com/browse/UNI-415) for screenshots.

@marionleborgne please review (or have someone else)

There is one minor issue which is that when there is a horiztonal scrollbar, the vertical scrollbar is only visible when you scroll all the way to the right. I don't think there is any easy way around this unless we switch to non-fixed header but I think it works much better with the fixed header and no seeing the scrollbar is not a big deal IMO.


unicorn/app/config/default.json
```diff
@@ -79,4 +79,4 @@
   "samples": {
     "path": "../samples"
   }
-}
+}
```
marionleborgne: Missing new line at EOF

marionleborgne:
Thanks @scottpurdy. LGTM 👍 

@lscheinkman there is one failing test but it is already failing in master. I am submitting a PR that fixes it.

PR #761, opened 2016-05-05
UNI-424 Don't connect dots that are too far apart

mrcslws:
Disable interpolation when we detect a time gap in the data.

Add a simple initial heuristic for detecting gaps. Compute all the time-deltas, find the 75th percentile, and multiply it by a factor (currently 1.5). The result is the gap threshold. This heuristic will be easy to tweak now that it's in place.

We have to detect gaps in aggregated and unaggregated data independently, and then merge them. This was difficult to fit into the existing code, so I shuffled the way we compute the Dygraphs input data.

### DIFF
--- a/unicorn/app/browser/components/Chart.jsx
+++ b/unicorn/app/browser/components/Chart.jsx
@@ -138,7 +138,7 @@ export default class Chart extends React.Component {
     // init, render, and draw chart!
     options.labelsUTC = true;
     options.dateWindow = this._chartRange;  // update viewport of range selector
-    options.axes.y.valueRange = [metaData.min, metaData.max];  // lock y-axis
+    options.axes.y.valueRange = [metaData.min, metaData.max];
     this._previousDataSize = data.length;
     this._dygraph = new Dygraph(element, data, options);
 
@@ -149,7 +149,7 @@ export default class Chart extends React.Component {
   }
 
   /**
-   * DyGrpahs Chart Update Logic and Re-Render
+   * DyGraphs Chart Update Logic and Re-Render
    */
   _chartUpdate() {
     let {data, metaData, options} = this.props;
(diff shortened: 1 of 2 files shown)


mrcslws:
cc @marionleborgne

mrcslws:
Note that I haven't enabled drawing edge points around gaps due to issue https://github.com/danvk/dygraphs/issues/745. We could enable it in a later change.

marionleborgne:
When looking at the aggregated data, the heuristic to represent missing data looks good. However, when displaying the raw data, it looks super confusing. See below. 

screen shot 2016-05-05 at 3 52 32 pm

I am investigating why it looks like them. Some of that might not come from your changes but might be related to https://jira.numenta.com/browse/UNI-422

marionleborgne:
@mrcslws actually, there's a problem with the way you show gaps in the raw data. See screenshot

screen shot 2016-05-05 at 4 26 40 pm

marionleborgne:
Here is the file to reproduce what I am showing 
[missing_data.csv.zip](https://github.com/numenta/numenta-apps/files/251356/missing_data.csv.zip)

marionleborgne:
FYI @mrcslws and I discussed the bug where the raw data is missing in the charts and a bug report about that (https://jira.numenta.com/browse/UNI-438) which is blocking the associated JIRA (UNI-424) reporting the bug. Thanks for investigating @mrcslws. Let's not merge this PR until this bug is fixed.

marionleborgne:
Thanks @mrcslws. Nice fix 👍 and great branch name... :trollface:

marionleborgne:
FYI the comment https://github.com/numenta/numenta-apps/pull/761#issuecomment-217308650 has been added in the PR #774 based off this branch. I took the logic to compute the `gapThreshold` out of the `insertIntoGaps` function so that there is a consistent heuristic across aggregated and non-aggregated data (gapThreshold is now the same for aggregated and non-aggregated data)

PR #762, opened 2016-05-06
UNI-438 Bugfix: skipping many raw data points when merging with aggregated

mrcslws:
The issue here was in `_overlayDataSeries`. It's using a `.forEach` and not always doing something with the `item`.

I already reworked the way we build up this data for another change: https://github.com/numenta/numenta-apps/pull/761 . It fixes the bug. We're still discussing that change, so this PR is just that one without the controversial parts.

### DIFF
--- a/unicorn/app/browser/components/Chart.jsx
+++ b/unicorn/app/browser/components/Chart.jsx
@@ -138,7 +138,7 @@ export default class Chart extends React.Component {
     // init, render, and draw chart!
     options.labelsUTC = true;
     options.dateWindow = this._chartRange;  // update viewport of range selector
-    options.axes.y.valueRange = [metaData.min, metaData.max];  // lock y-axis
+    options.axes.y.valueRange = [metaData.min, metaData.max];
     this._previousDataSize = data.length;
     this._dygraph = new Dygraph(element, data, options);
 
@@ -149,7 +149,7 @@ export default class Chart extends React.Component {
   }
 
   /**
-   * DyGrpahs Chart Update Logic and Re-Render
+   * DyGraphs Chart Update Logic and Re-Render
    */
   _chartUpdate() {
     let {data, metaData, options} = this.props;
@@ -180,6 +180,7 @@ export default class Chart extends React.Component {
     }
 
     // update chart
+    options.axes.y.valueRange = [metaData.min, metaData.max];
     options.dateWindow = this._chartRange;
     options.file = data;  // new data
     this._previousDataSize = data.length;
--- a/unicorn/app/browser/components/ModelData.jsx
+++ b/unicorn/app/browser/components/ModelData.jsx
@@ -77,32 +77,8 @@ function sortedMerge(a, b, compareFunction) {
  * @returns {Array} - data with gap values inserted
  */
 function insertIntoGaps(data, vals) {
-  let deltas = [];
-  for (let i = 1; i < data.length; i++) {
-    deltas.push(data[i][DATA_INDEX_TIME].getTime() -
-                data[i-1][DATA_INDEX_TIME].getTime());
-  }
-  deltas.sort();
-
-  let percentile = 0.75;
-  let gapThreshold = 1.5 * deltas[Math.floor(deltas.length*percentile)];
-
-  let newData = [];
-  data.forEach((item, rowid) => {
-    newData.push(item);
-
-    if (rowid + 1 < data.length) {
-      let curr = item[DATA_INDEX_TIME].getTime();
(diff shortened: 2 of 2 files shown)


unicorn/app/browser/components/Chart.jsx
```diff
@@ -138,7 +138,7 @@ export default class Chart extends React.Component {
     // init, render, and draw chart!
     options.labelsUTC = true;
     options.dateWindow = this._chartRange;  // update viewport of range selector
-    options.axes.y.valueRange = [metaData.min, metaData.max];  // lock y-axis
+    options.axes.y.valueRange = [metaData.min, metaData.max];
```
marionleborgne: why did you remove this comment?

unicorn/app/browser/components/ModelData.jsx
```diff
@@ -38,6 +38,132 @@ const {
...
+  }
+
+  for (; j < b.length; j++) {
+    ret.push(b[j]);
+  }
+
+  return ret;
+}
+
+/**
```
marionleborgne: why do you keep the heuristic here if it does nothing? is this to prepare for your other related PR https://github.com/numenta/numenta-apps/pull/761?

unicorn/app/browser/components/ModelData.jsx
```diff
@@ -38,6 +38,132 @@ const {
...
+ * Use a heuristic to detect gaps in timestamps in the data.
+ *
+ * At each gap, insert [midpointOfGap, vals[0], vals[1], ...] as a new datum.
+ *
+ * @param {Array} data - Array of arrays: [[Date, ...], [Date, ...], ...]
+ * @param {Array} vals - Values concatenated to timestamp at every single gap
+ * @returns {Array} - data with gap values inserted
+ */
+function insertIntoGaps(data, vals) {
```
marionleborgne: same question as above for this function

unicorn/app/browser/components/ModelData.jsx
```diff
@@ -38,6 +38,132 @@ const {
...
+ * @param {boolean} aggregated - Whether the model is aggregated.
+ * @param {boolean} rawDataInBackground - Whether the raw data should be drawn
+ *                                        (in addition to the aggregated data)
+ * @returns {Array} - Tuple:
+ *                    Array: Dygraphs multi-dimensional array
+ *                    boolean: minVal
+ *                    boolean: maxVal
+ * @see http://dygraphs.com/tests/independent-series.html
+ */
+function prepareData(
```
marionleborgne: 👍 nice refactoring. it fixes the bug of the missing points.

marionleborgne:
👍 LGTM. If you could just reply to my two questions above, you can merge. It fixes the bug. Tests are passing (minus the one that's still failing on master - still working on that)

unicorn/app/browser/components/ModelData.jsx
mrcslws: Well, this `sortedMerge` function is used independently of the gap detection. It's used for merging the aggregated and raw data.

unicorn/app/browser/components/ModelData.jsx
marionleborgne: No, I am talking about the function `insertIntoGaps`

unicorn/app/browser/components/ModelData.jsx
marionleborgne: No, I am talking about the function `insertIntoGaps`

unicorn/app/browser/components/ModelData.jsx
mrcslws: I don't have a strong opinion here. I could totally remove it. It's just a nice placeholder to keep this code compatible with the gap detection while we figure out the actual heuristic.

unicorn/app/browser/components/Chart.jsx
mrcslws: It's now updated on every chart update. Though yes it's still locked in some sense -- it doesn't shift with the selected timestep range. There's no right place to describe all of this in a single comment, and it just seems like this is all communicated by the code, no comment needed.

unicorn/app/browser/components/ModelData.jsx
mrcslws: Ahh. Same answer.

unicorn/app/browser/components/ModelData.jsx
marionleborgne: ok. we need the heuristic to detect gaps anyway so let's keep it.

marionleborgne:
👍

PR #763, opened 2016-05-06
UNI-423 Use a single y scale for aggregated and raw data

mrcslws:
We don't currently set the valueRange of the "show non-aggregated" data, so its y scale (which doesn't have a visualized axis) quietly changes based on the selected timestep range.

Instead of giving it its own valueRange, just unite with the aggregated data. Share a y scale. Arguably we'd want to draw a second y axis if we want two different y scales.


marionleborgne:
👍 very nice. thanks.

PR #764, opened 2016-05-06
UNI-424 Don't connect dots that are too far apart

mrcslws:
This is a continuation of #761, which was transitively "merged" when #762 was merged because of my cute git branch structure. Oops.

This PR discussion can just focus on the heuristic.


mrcslws:
/cc @marionleborgne 

I think you already know this, but this isn't ready to be merged. We should merge other changes that improve y scaling and show all the aggregated data, then discuss this heuristic.

mrcslws:
The current heuristic is:

Compute all the time-deltas between points. Find the 75th percentile time-delta. Multiply it by 1.5. The result is the gap threshold. Lines will be drawn for every time-delta that is less than the gap threshold.

PR #765, opened 2016-05-06
Fix db export integration test

marionleborgne:
This fixes the integration test and the way we are exporting results to CSV.

The convention [1] is now:
1. To allow parsing of date/time values that have a timezone
2. To not take the timezone into account in the app for vizualization or export. This means that a CSV that has a timezone can be parsed but that the timzone will not be represented in the app (i.e. timezones are not reflected in the charts or in the exported result CSV)

[1] introduced by the PR #751 by @mrcslws and which fixes UNI-334


marionleborgne:
@lscheinkman this should fix the failing integration test. I have also QA-ed the app with the attached dataset and the export format is valid.
[export QA.zip](https://github.com/numenta/numenta-apps/files/251937/export.QA.zip)

Please review. Thanks!

marionleborgne:
OK, I screwed up and ended up in git hell :-/ This started with me accidentally reverting my (good) commit and wanting to fix it. I'm not quite sure how I managed to end up in the current state ... that'll teach me to play with git when it's too late. The correct commit is https://github.com/numenta/numenta-apps/pull/765/commits/2ba782b86427a9714646ce5d15e478e2536c15f4 I think you're better off waiting until this is fixed before reviewing @lscheinkman, I don't think this is in a merge-able state at all now.... sorry.

marionleborgne:
I'm closing this and have a clean PR here: https://github.com/numenta/numenta-apps/pull/766

PR #766, opened 2016-05-06
Fixes integration test and the way we export files.

marionleborgne:
This fixes the integration test and the way we are exporting results to CSV.

The convention [1] is now:
1. To allow parsing of date/time values that have a timezone
2. To not take the timezone into account in the app for vizualization or export. This means that a CSV that has a timezone can be parsed but that the timzone will not be represented in the app (i.e. timezones are not reflected in the charts or in the exported result CSV)

[1] introduced by the PR #751 by @mrcslws and which fixes UNI-334


unicorn/tests/js/integration/services/DatabaseService_test.js
```diff
@@ -94,26 +94,26 @@ const EXPECTED_METRICS = [EXPECTED_TIMESTAMP, EXPECTED_METRIC];
...
   {metric_uid: EXPECTED_METRIC_ID,
-   timestamp: Date.parse('2015-08-26T19:50:31Z'),
-   metric_value: 16},
+    timestamp: Date.parse('2015-08-26T19:50:31'),
+    metric_value: 16},
   {metric_uid: EXPECTED_METRIC_ID,
-   timestamp: Date.parse('2015-08-26T19:51:31Z'),
-   metric_value: 19}
+    timestamp: Date.parse('2015-08-26T19:51:31'),
+    metric_value: 19}
```
mrcslws: Actually these should have the 'Z'. Technically your change works in V8 / Chromium because they default to parsing as UTC. In Firefox this will fail, since it parses as local time.

marionleborgne:
@lscheinkman @mrcslws : I filed a JIRA is capturing the (voluntary) timezone inaccuracy introduced by this PR in the exported CSV results timestamp format. https://jira.numenta.com/browse/UNI-440

The plan is to get to this JIRA after private beta.

I also addressed @mrcslws feedback. Thanks.

unicorn/app/browser/components/Model.jsx
```diff
@@ -359,6 +359,13 @@ export default class Model extends React.Component {
     let {model, file, valueField, timestampField} = this.props;
     let title = model.metric;
 
+    // The convention is to ignore timezones when rendering or exporting time
+    // in the app.
+    let exportedTimestampFormat = timestampField.format;
+    if (exportedTimestampFormat.slice(-1) === 'Z') {
```
lscheinkman: @marionleborgne Can you add a `FIXME` here pointing to `UNI-440`?

unicorn/tests/js/integration/services/DatabaseService_test.js
```diff
@@ -781,7 +781,14 @@ describe('DatabaseService:', () => {
       });
       service.putModelDataBatch(BATCH_MODEL_DATA, (error) => {
         assert.ifError(error);
-        service.exportModelData(EXPECTED_METRIC_ID, EXPORTED_FILENAME, (error, res) => {
+
+        // The convention is to ignore timezones when rendering or exporting time
+        // in the app.
+        let exportedTimestampFormat = EXPECTED_TIMESTAMP.format;
+        if (exportedTimestampFormat.slice(-1) === 'Z') {
```
lscheinkman: @marionleborgne Can you add a FIXME here pointing to UNI-440?

lscheinkman:
👍 LGTM assuming it does not fix UNI-440

marionleborgne:
@lscheinkman correct, it does not fix UNI-440. I added FIXMEs in the code pointing to this JIRA.

PR #768, opened 2016-05-07
UNI-422 Block the dygraphs panning code from messing with the y scale

mrcslws:
I stepped through the Dygraphs interaction code, and this is what I came up with.


marionleborgne:
👍 nice. this fixes the bug. thanks @mrcslws

Issue #503, opened 2015-12-05
HTM Engine should allow users to provide entire model params


rhyolight:
Instead of just min/max for the encoder params, I would like to provide the entire model params myself.

ryanjmccall:
OK @rhyolight  this seems approachable. I'm looking at this from the webapp.py point of view. so call permutations_runner whenever a new model is requested. then it would be easy to add an argument to htmengine.adapters.datasource.custom._CustomDatasourceAdapter#monitorMetric and similar adapters. but maybe the swarming takes so long it should be handled in a separate thread..

rhyolight:
There is no swarming in HTM Engine at all, so you provide the model params, and HTM Engine creates a model.

ryanjmccall:
Got it. I see there are 3 DataSourceAdapters -- Custom, Cloudwatch, and Autostack. Would you prefer this change be implemented uniformly for all 3 or should it only focus on Custom which may be the only place where people would want to experiment with passing in swarm params? 

The model spec is currently a dict with top level keys 'modelParams', 'datasource', and 'metricSpec'. I can add another key 'swarmParams' and check for that inside `monitorMetric()`

rhyolight:
I must defer those questions to those will will ultimately review the PR, likely @oxtopus or @vitaly-krugl

vitaly-krugl:
I would recommend focusing on Custom metrics, which is good for experimentations. The others are mote tightly coupled with the HTM-IT application. In fact, if we were re-implementing HTM-IT, we would probably implement it on top of Custom Metrics, as we had later done in Taurus.

oxtopus:
@rhyolight it may help to limit scope somewhat, since `htmengine` only works with a limited set of parameters.  It's not safe to assume that `htmengine` will always work with every technically valid parameter.

rhyolight:
@oxtopus Do you have an idea of the specific params that might cause problems?

oxtopus:
Classification, non-anomaly, or multistep models come to mind, basically anything but a single-step, single-field scalar temporal anomaly model.  Anything but, and results are "unexpected".

vitaly-krugl:
I concur with @oxtopus

rhyolight:
Hm, this is trickier then. @ryanjmccall What do you think about putting in a validation step before creating the model?

ryanjmccall:
[...]

PR #769, opened 2016-05-09
Complete model params option for CustomDatasourceAdapter.monitorMetric

ryanjmccall:
Fixes https://github.com/numenta/numenta-apps/issues/503


ryanjmccall:
@rhyolight Would you please see if someone can review this? Many thanks

oxtopus:
Thanks @ryanjmccall 

I'll take a look.

rhyolight:
Nice Ryan!

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -281,6 +283,19 @@ def monitorMetric(self, modelSpec):
           "Neither uid nor metric name present in metricSpec; modelSpec=%r"
           % (modelSpec,))
 
+    if "completeModelParams" in modelSpec:
+      # Full model params have been provided by the user
+      swarmParams = dict()
+      swarmParams["modelConfig"] = copy.deepcopy(modelSpec["completeModelParams"])
```
oxtopus: This line, and the ones that follow exceed line length limits, as indicated by running pylint:

```
...
C0301, 289: 0 - Line too long (82/80) (line-too-long)
C0301, 290: 0 - Line too long (116/80) (line-too-long)
C0301, 292: 0 - Line too long (104/80) (line-too-long)
C0301, 293: 0 - Line too long (112/80) (line-too-long)
C0301, 295: 0 - Line too long (101/80) (line-too-long)
C0301, 296: 0 - Line too long (107/80) (line-too-long)
...
```

vitaly-krugl:
@rhyolight and @oxtopus: I would like to make a review pass, too. So, please hold off on merging until you hear from me. Thx.

oxtopus:
Passing test results at https://ci.numenta.com/browse/TAUR-TAUR0-TSTS-1

This could use some linting, but otherwise looks good to me.  My only concern is the possibility of passing incorrect, invalid, or otherwise incompatible model params.  For example, what happens when a user submits a non-anomaly parameter set?

vitaly-krugl:
@oxtopus, were you done with the review?

oxtopus:
Yes, I'm done.  There is some feedback for @ryanjmccall, but the ball's in your court now.

ryanjmccall:
@oxtopus @vitaly-krugl I took the easy route with the json schema validation, partly because I had the impression that the specifics of the swarm structure shouldn't be coupled with htm engine. if you want the full validation of that I can add it. I'll also clean up the lint problem.

oxtopus:
I don't think you need to do full validation, but you do need to assert on some basic assumptions.  For example, that it is, in fact, an anomaly model and not prediction or that it's a single-step and not multi-step.  `htmengine` is built with alot of these sorts of assumptions and the user may get unexpected results if they are mistaken to believe that because they can specify the complete model params, that everything will just magically work with any given technically valid model param set.

vitaly-krugl:
@ryanjmccall, I am starting my code review. I will add a comment when I am done with the round. To make it easier for me, please don't push any changes to the branch until you hear from me that I am done. Thx!

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -281,6 +283,19 @@ def monitorMetric(self, modelSpec):
...
+      swarmParams["inferenceArgs"] = {"predictionSteps": [1], "predictedField": "c1", "inputPredictedField": "auto"}
```
vitaly-krugl: If we're going to accept custom modelConfig from user, then we shouldn't hard-code inferenceArgs; rather, allow the user to specify it. For example, see this schema used by "unicorn": https://github.com/numenta/numenta-apps/blob/master/unicorn/py/unicorn_backend/model_opt_schema.json

htmengine/htmengine/adapters/datasource/custom/__init__.py
vitaly-krugl: I concur with @oxtopus: run pylint using https://github.com/numenta/numenta-apps/blob/master/pylintrc, then make sure that any new errors/warnings/messages introduced by this PR are fixed before we can continue with the next round of code review.

htmengine/htmengine/adapters/datasource/model_spec_schema.json
```diff
@@ -32,6 +32,11 @@
           "type": "number"
         }
       }
+    },
+    "completeModelParams": {
+      "description": "Model creation parameter overrides for entire model.",
```
vitaly-krugl: is `completeModelParams` mutually-exclusive with `modelParams` (I think so), then reflect it in the descriptions of both properties. I think that there might even be a way to use JSON Schema to indicate mutually-exclusive structures. We really want to fail requests that specify properties that are mutually-exclusive.

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -281,6 +283,19 @@ def monitorMetric(self, modelSpec):
           "Neither uid nor metric name present in metricSpec; modelSpec=%r"
           % (modelSpec,))
 
+    if "completeModelParams" in modelSpec:
```
vitaly-krugl: Should fail with `ValueError` exception if both `completeModelParams` and `modelParams` are passed in modelSpec.

htmengine/htmengine/adapters/datasource/model_spec_schema.json
vitaly-krugl: In the description, please reference the data format of this object. If there is no existing JSON Schema for it in nupic (pls check), then refer elsewhere in nupic where a user might more definitively ascertain it (perhaps a core function in nupic that takes the same arg, such as OPF model factory method).

Also, cite the restrictions, such as: data being a scalar (int or float) value,  output anomaly, etc.

htmengine/tests/integration/data/example_model_params.json
```diff
@@ -0,0 +1,87 @@
...
+			"permanenceInc": 0.1,
+			"seed": 1960,
+			"temporalImp": "cpp",
+			"verbosity": 0
+		},
+		"trainSPNetOnlyIfRequested": false
+	},
+	"predictAheadTime": null,
+	"version": 1
+}
```
vitaly-krugl: Missing final newline?

htmengine/tests/integration/data/example_model_params.json
```diff
@@ -0,0 +1,87 @@
+{
```
vitaly-krugl: `example_model_params.json` may not be the best name for it. Perhaps the file's name should reflect the test suite that uses it, to make it easier to correlate the two?

htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py
```diff
@@ -399,6 +400,57 @@ def testMonitorMetricWithMinMax(self):
     self.checkEncoderResolution(metricId, 0, 100)
 
 
+  def _getPathToData(self, filename):
```
vitaly-krugl: Does this need to be an instance method? It has no use for self, so a staticmethod would be more appropriate.

vitaly-krugl: Would be cleaner to turn this method into `_openTestDataFile(filename)`, and then the usage would be simply:

```
with self._openTestDataFile(modelParamsPath) as f:
   completeModelParams = json.load(f)
```

htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py
```diff
@@ -399,6 +400,57 @@ def testMonitorMetricWithMinMax(self):
...
+
+    adapter = datasource_adapter_factory.createCustomDatasourceAdapter()
+
+    g_log.info("Creating htmengine custom metric; name=%s", metricName)
+    metricId = adapter.createMetric(metricName)
+    self.addCleanup(adapter.deleteMetricByName, metricName)
+
+    modelParamsPath = self._getPathToData("example_model_params.json")
+    with open(modelParamsPath, "rb") as f:
+        completeModelParams = json.load(f)
```
vitaly-krugl: Our conventions call for 2-space indents. Pls re-run pylint with pylintrc from numenta-apps, and adjust all such things.

htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py
```diff
@@ -399,6 +400,57 @@ def testMonitorMetricWithMinMax(self):
     self.checkEncoderResolution(metricId, 0, 100)
 
 
+  def _getPathToData(self, filename):
+    """ Returns the absolute path to a file that lives in the relative data/directory """
+    basePath = os.path.split(os.path.abspath(__file__))[0]
+    dataDirPath = os.path.join(basePath, "../../../data")
+    knownDataFilePath = os.path.join(dataDirPath, filename)
+    return knownDataFilePath
```
vitaly-krugl: If returning `knownDataFilePath` immediately after computing it, then the unnecessary local variable convolutes the function. Simply `return os.path.join(dataDirPath, filename)` would be sufficient.

htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py
```diff
@@ -399,6 +400,57 @@ def testMonitorMetricWithMinMax(self):
...
+    self._validateModelSpec(json.loads(metricObj.parameters))
+
+    self.assertIn(metricObj.status, (MetricStatus.CREATE_PENDING,
+                                     MetricStatus.ACTIVE))
+    self.assertEqual(json.loads(metricObj.parameters), modelSpec)
+
+    g_log.info("Waiting for model to become active")
+    self.checkModelIsActive(metricId)
+
+
```
vitaly-krugl: Also need test of export/import with completeModelParams. See `testExportImport` method here as an example.

vitaly-krugl:
@ryanjmccall: thanks for the changes! Please address the cumulative feedback and re-submit for the next round.

htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py
```diff
@@ -399,6 +400,57 @@ def testMonitorMetricWithMinMax(self):
...
+    metricName = "test-" + uuid.uuid1().hex
+
+    adapter = datasource_adapter_factory.createCustomDatasourceAdapter()
+
+    g_log.info("Creating htmengine custom metric; name=%s", metricName)
+    metricId = adapter.createMetric(metricName)
+    self.addCleanup(adapter.deleteMetricByName, metricName)
+
+    modelParamsPath = self._getPathToData("example_model_params.json")
+    with open(modelParamsPath, "rb") as f:
```
vitaly-krugl: Not sure if the single character "f" variable would pass the check with numenta-apps pylinrc.

htmengine/tests/integration/data/example_model_params.json
ryanjmccall: by test suite I take it you mean an instance of `unittest.TestCase`? changing the name helps correlate but what if multiple suites start to reference the file?

htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py
ryanjmccall: addressed

htmengine/htmengine/adapters/datasource/model_spec_schema.json
ryanjmccall: @vitaly-krugl Yes, `completeModelParams` is mutex WRT `modelParams`. I took a look at the JSON schema situation and it's a bit of a rabbit hole. `htmengine` is using 'validictory' which doesn't really support (see https://github.com/jamesturk/validictory/issues/20) "oneOf" and the other combining operations of JSON Schema draft 4 (http://spacetelescope.github.io/understanding-json-schema/reference/combining.html). It is suggested here ( https://github.com/jamesturk/validictory/issues/20#issuecomment-81824458) to use https://github.com/gazpachoking/jsonref to preprocess the refs away. Another viable option may be the `jsonschema` package which states it supports draft 4: https://pypi.python.org/pypi/jsonschema but I don't see it used in Numenta repos... I fear I'm getting far afield of the initial task, would it suffice to make programmatic checks (throwing `ValueError` as you suggested above) in `monitorMetric()` and address the json schema validation in another PR?

vitaly-krugl: Check in code that raises ValueError with a helpful message is fine.

htmengine/tests/integration/data/example_model_params.json
vitaly-krugl: "example*" is clearly not a relevant name here. 

> ...multiple suites start to reference the file?

We'll rename it if/when needed. There is incredibly little development happening in that area, so not likely that anything else will need that file.

htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py
```diff
@@ -665,6 +854,80 @@ def checkExportSpec(exportSpec):
...
+    # Import
+    metricId = adapter.importModel(
+      htmengine.utils.jsonDecode(htmengine.utils.jsonEncode(exportSpec)))
+
+    with self.engine.connect() as conn:
+      metricObj = repository.getMetric(conn,
+                                       metricId,
+                                       fields=[schema.metric.c.parameters,
+                                               schema.metric.c.status])
+    self.assertEqual(metricObj.status, MetricStatus.CREATE_PENDING)
```
ryanjmccall: not 100% certain about the MetricStatus here

ryanjmccall:
@vitaly-krugl @oxtopus thanks for the helpful feedback. I believe I've made the necessary additions and corrections.

htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py
vitaly-krugl: @ryanjmccall: I am glad you asked. This check leads to a race condition between the test app and htmengine. Assuming everything went well (as it should in this test), the metric should be in one of two states at this point: `CREATE_PENDING` or `ACTIVE`. So, it should be checking `self.assertIn(metricObj.status, (MetricStatus.CREATE_PENDING, MetricStatus.ACTIVE))`

ryanjmccall: @vitaly-krugl thanks for the pointer. I've updated the unit test per your suggestion.

vitaly-krugl:
@ryanjmccall: I am reviewing... Please don't push any new changes into this PR until I'm done, at which time I will post another comment. Thx!

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -255,6 +257,33 @@ def monitorMetric(self, modelSpec):
     :raises htmengine.exceptions.MetricAlreadyMonitored: if the metric is
       already being monitored
     """
+    if "completeModelParams" in modelSpec:
+      # User has provided full model params
+
+      # 'completeModelParams' and 'modelParams' are mutex
+      if "modelParams" in modelSpec:
+        raise ValueError("Specify either \"modelParams\" or "
```
vitaly-krugl: The escaped quotes (e.g., `\"modelParams\"`) affect readability. There is no need for them. `"Specify either modelParams or completeModelParams, but not both; modelSpec=%r"` is better.

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -255,6 +257,33 @@ def monitorMetric(self, modelSpec):
...
+                         "\"completeModelParams\" but not both; modelSpec=%r"
+                         % (modelSpec,))
+
+      # 'completeModelParams', 'inferenceArgs', 'timestampFieldName', and
+      # 'valueFieldName' must all be present together
+      if "inferenceArgs" not in modelSpec:
```
vitaly-krugl: I think we would be better off nesting `modelConfig`, `inferenceArgs`, `timestampFieldName`, and `valueFieldName` inside the property named `completeModelParams` instead of polluting the `modelSpec` namespace.

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -281,6 +310,22 @@ def monitorMetric(self, modelSpec):
           "Neither uid nor metric name present in metricSpec; modelSpec=%r"
           % (modelSpec,))
 
+    # Construct singular swarm params structure
+    if "completeModelParams" in modelSpec:
+      swarmParams = dict()
+      swarmParams["modelConfig"] = copy.deepcopy(
+        modelSpec["completeModelParams"])
+      swarmParams["inferenceArgs"] = modelSpec["inferenceArgs"]
```
vitaly-krugl: Why is modelConfig copied (`copy.deepcopy`), but inferenceArgs isn't? This might warrant a code comment, so that the nuance isn't lost on other readers.

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -281,6 +310,22 @@ def monitorMetric(self, modelSpec):
           "Neither uid nor metric name present in metricSpec; modelSpec=%r"
           % (modelSpec,))
 
+    # Construct singular swarm params structure
```
vitaly-krugl: This function's body has become quite long, now over 100 lines. Need to factor out most of the new code into a separate function (e.g., `generateSwarmParamsFromCompleteModelParams`), similar to `scalar_metric_utils.generateSwarmParams` that's used by the legacy code below.

The new function would contain the new property consistency checks (`if "inferenceArgs" not in modelSpec` and subsequent checks) as well as code below that creates the swarmParams from completeModelParams.

htmengine/htmengine/adapters/datasource/model_spec_schema.json
```diff
@@ -32,6 +33,26 @@
           "type": "number"
         }
       }
+    },
+    "completeModelParams": {
+      "description": "Complete model params obtained from calling nupic.swarming.permutations_runner.runWithConfig() passing swarm config specifying timestamp and single scalar volume metric fields as well as 'TemporalAnomaly' inference type'",
```
vitaly-krugl: Since modelConfig may be generated by a number of techniques, not limited to `nupic.swarming.permutations_runner.runWithConfig()` (e.g., `nupic.frameworks.opf.common_models.cluster_params. getScalarMetricWithTimeOfDayAnomalyParams`), prefer to document it by who accepts it:  `nupic.frameworks.opf.modelfactory.ModelFactory.create`

htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py
```diff
@@ -399,6 +397,197 @@ def testMonitorMetricWithMinMax(self):
...
+      "timestampFieldName": "jon",
+      "valueFieldName": "snow",
+      "modelParams": {
+        "min": 0,
+        "max": 100
+      }
+    }
+
+    with self.assertRaises(ValueError):
+      adapter.monitorMetric(modelSpec)
```
vitaly-krugl: `monitorMetric` raises a number of ValueError exceptions with different messages. If you wish to confirm that the correct error was detected (not some other problem in the args or code under test bug), then you would change to `with self.assertRaises(ValueError) as excCtx:` and then verify `excCtx.exception.args[0]` content after control leaves the context manager.

htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py
```diff
@@ -665,6 +854,81 @@ def checkExportSpec(exportSpec):
     checkExportSpec(exportSpec)
 
 
+  def testExportImport_completeModelParams(self):
```
vitaly-krugl: Our convention does not allow underscores in function names, so something like `testExportImportWithCompleteModelParams` would be better.

vitaly-krugl:
@ryanjmccall, thanks for addressing the earlier round of feedback. We're almost there. Please address the new feedback and merge the latest from numenta-apps master into your PR. The latter will enable me to run your PR through our internal CI test pipelines prior to merging into master. As it stands, there is a build bug (at least on linux) that happens when fabric and paramiko are being installed; this bug has already been resolved in numenta-apps master.

Many thanks,
Vitaly

htmengine/htmengine/adapters/datasource/custom/__init__.py
ryanjmccall: I guess I don't see the nuance you're referring to. upon a second look it seems inferenceArgs should be copied as well since it's a dict and we don't want it being modified from elsewhere (however unlikely that may be)

vitaly-krugl: @ryanjmccall, by "nuance" I meant that if there is some special reason that modelConfig needs to be copied, but inferenceArgs doesn't, then it would be nuance that needs to be documented.

vitaly-krugl: On my second thought :), do we really need the deepcopy here at all? As far as I can tell, this API doesn't hold on to the dicts that were passed in (stores the data in databases). The copy is unnecessary, unless this API method modifies the dicts passed in by the user.

ryanjmccall: good point, I'll take it out

htmengine/htmengine/adapters/datasource/custom/__init__.py
ryanjmccall: I'll have it in the next commit

htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py
ryanjmccall: OK I'm making this change but note I'm unable to reasonably compare the entire message because the modelSpec is included (and it probably should be to be informative to the user). To make testing this manageable I've abstracted the key part of the message into a private constant in the `scalar_metric_utils` module and only use this part for the `assertEqual`

ryanjmccall:
@vitaly-krugl Please see above new commits in response to your round 2 feedback. 

I'm wondering if I should add a pylint disable for these kinds of messages `W0212, 492: 6 - Access to a protected member _MUTEX_MODEL_SPEC_MSG of a client class (protected-access)` coming from `htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py` 

Thank you, Ryan

htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py
vitaly-krugl: `assertIn` may be helpful here, too

vitaly-krugl:
In unit tests, where it's common to access private member variables and such, it's useful to disable "Access to a protected member" warnings, such as in https://github.com/numenta/numenta-apps/blob/1ff572a21a5c27fd290822e572ce33f42e1ee19e/htmengine/tests/unit/model_checkpoint_mgr/model_checkpoint_mgr_test.py#L34-L35

ryanjmccall:
OK I've done just that for the unit test file. thx

vitaly-krugl:
@ryanjmccall, have you merged the latest numenta-apps master into your PR? I need that in order to validate your PR against our internal CI pipeline.

ryanjmccall:
@vitaly-krugl I believe I did that yesterday in the 3rd to last commit (https://github.com/numenta/numenta-apps/pull/769/commits/0becc781837aeae298f1706c37389fbf02a3ab49)

vitaly-krugl:
Got it, thanks! I looked at commit history, and somehow missed that entry. I will start the build, then.

Vitaly

vitaly-krugl:
It passed our internal Taurus CI pipeline tests. Now waiting for HTM-IT pipeline build/tests to complete.

vitaly-krugl:
Our internal HTM-IT pipeline passed as well.

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -255,8 +257,11 @@ def monitorMetric(self, modelSpec):
     :raises htmengine.exceptions.MetricAlreadyMonitored: if the metric is
       already being monitored
     """
-    metricSpec = modelSpec["metricSpec"]
+    # Attempt to build swarm params from complete model params if present
+    swarmParams = (
+      scalar_metric_utils.generateSwarmParamsFromCompleteModelParams(modelSpec))
```
vitaly-krugl: @ryanjmccall, this change went a bit too far beyond the recommendation. It was helpful (e.g., for readability) to have the higher-level logic here that determined whether to invoke the logic of  `scalar_metric_utils.generateSwarmParamsFromCompleteModelParams` or `scalar_metric_utils.generateSwarmParams`.

Also, in a language like Python, you would communicate the failure (lack of complete model params) via an exception, not by returning an empty `swarmParams`. Then the code would react to the exception. However, once you restore the higher-level logic that discerns whether to call one or the other, then such an exception wouldn't be needed.

htmengine/htmengine/adapters/datasource/model_spec_schema.json
```diff
@@ -32,6 +32,34 @@
...
         }
       }
+    },
+    "completeModelParams": {
+      "description": "Complete model params to be passed to nupic.frameworks.opf.modelfactory.ModelFactory.create().",
+      "required": false,
+      "type": "object",
+      "additionalProperties":false,
+      "properties": {
+        "modelParams": {
```
vitaly-krugl: Shouldn't this be `modelConfig` instead of modelParams for consistency with usage in the API?

htmengine/htmengine/adapters/datasource/model_spec_schema.json
```diff
@@ -32,6 +32,34 @@
           "type": "number"
         }
       }
+    },
+    "completeModelParams": {
+      "description": "Complete model params to be passed to nupic.frameworks.opf.modelfactory.ModelFactory.create().",
```
vitaly-krugl: This body of this description now belongs with the modelConfig property below. Also need a new description of completeModelParams

htmengine/htmengine/adapters/datasource/model_spec_schema.json
```diff
@@ -32,6 +32,34 @@
...
+      "required": false,
+      "type": "object",
+      "additionalProperties":false,
+      "properties": {
+        "modelParams": {
+          "description": "Model params for a model having a single scalar volume metric and a 'TemporalAnomaly' inference type.",
+          "type": "object",
+          "required": false
```
vitaly-krugl: All these properties inside completeModelParams are required, right?

htmengine/tests/integration/adapters/datasource/custom/custom_datasource_adapter_test.py
```diff
@@ -399,6 +401,239 @@ def testMonitorMetricWithMinMax(self):
...
+        "inferenceArgs": {"predictionSteps": [1], "predictedField": "snow",
+                          "inputPredictedField": "auto"},
+        "valueFieldName": "snow"
+      }
+    }
+
+    with self.assertRaises(ValueError) as excCtx:
+      adapter.monitorMetric(modelSpec)
+
+    excArgZero = excCtx.exception.args[0]
```
vitaly-krugl: nit (no need to fix, just an FYI); the following would have been acceptable, too, instead of this code block:

```
self.assertIn(scalar_metric_utils._NO_TIMESTAMP_FIELD_NAME_MSG,
              excCtx.exception.args[0])
```

htmengine/htmengine/runtime/scalar_metric_utils.py
```diff
@@ -93,6 +110,58 @@ def generateSwarmParams(stats):
   return swarmParams
 
 
+def generateSwarmParamsFromCompleteModelParams(modelSpec):
+  """ Generates a "swarm" parameter structure for model creation based on a
+  complete set of user-specified model parameters.
+
+  :param modelSpec: Model specification structure as defined by
+  'htmengine/adapters/datasource/model_spec_schema.json'
```
vitaly-krugl: thanks for the good docstring content. Per de facto convention here, the wrapped param and returns fields should be indented.

```
:param modelSpec: Model specification structure as defined by
'htmengine/adapters/datasource/model_spec_schema.json'
```

should become:

```
:param modelSpec: Model specification structure as defined by
  'htmengine/adapters/datasource/model_spec_schema.json'
```

etc.

vitaly-krugl:
@ryanjmccall, I found a few more issues in the new commits.

ryanjmccall:
@vitaly-krugl please see new commits addressing your latest feedback. thank you

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -255,8 +257,14 @@ def monitorMetric(self, modelSpec):
     :raises htmengine.exceptions.MetricAlreadyMonitored: if the metric is
       already being monitored
     """
-    metricSpec = modelSpec["metricSpec"]
+    # Attempt to build swarm params from complete model params if present
+    swarmParams = None
+    if "completeModelParams" in modelSpec:
```
vitaly-krugl: For readability and simplicity, this block should be below just before the current `if not swarmParams` block and there is no need for `swarmParams = None` as follows:

```
if "completeModelParams" in modelSpec:
  swarmParams = (
    scalar_metric_utils.generateSwarmParamsFromCompleteModelParams(
      modelSpec))
else:
  # Generate swarm params from specified metric min and max or estimate them
  modelParams = modelSpec.get("modelParams", dict())
  minVal = modelParams.get("min")
  maxVal = modelParams.get("max")
  . . .
```

vitaly-krugl:
@ryanjmccall, thank you for addressing the prior feedback. We're down to one more item.

Best,
Vitaly

htmengine/htmengine/adapters/datasource/custom/__init__.py
ryanjmccall: @vitaly-krugl I agree that approach is simpler, however, I was concerned with this line (https://github.com/ryanjmccall/numenta-apps/blob/f9be9ccb705203922a00b923d68852b84746c3bc/htmengine/htmengine/adapters/datasource/custom/__init__.py#L282) in the block of code that is in charge of getting metric id. So assuming the refactor you propose, then we'd have put the metricId logic first (b/c the else clause and after the else clause use metricId). Then there could be a case where we call `_createMetric`, and then `scalar_metric_utils.generateSwarmParamsFromCompleteModelParams` raises a `ValueError`. I was concerned that this would be an inconsistent state where a metricId is "tracked" by the engine, but doesn't actually reference an existing model. Please advise.

vitaly-krugl: I see what you mean. However, the simplification would not introduce a new problem: see https://github.com/ryanjmccall/numenta-apps/blob/f9be9ccb705203922a00b923d68852b84746c3bc/htmengine/htmengine/adapters/datasource/custom/__init__.py#L299-L301

My recommendation is to simplify and also file an issue for deleting the metric on error if the metric was created here.

ryanjmccall:
@vitaly-krugl ready again

vitaly-krugl:
LGTM, @ryanjmccall. Thank you. I am waiting for an internal HTM-IT CI pipeline to pass before merging.

vitaly-krugl:
@ryanjmccall: HTM-IT CI pipeline passed, merging . . .

rhyolight:
👏

ryanjmccall:
@vitaly-krugl many thanks for your help with this

vitaly-krugl:
@ryanjmccall, thanks for sticking with it to completion.

PR #770, opened 2016-05-09
UNI-404 Control our own y scale

mrcslws:
Now every time the chart draws, it checks back to see what y minimum and maximum it should use. I essentially subclassed Dygraph and added this feature.

I chose a policy of: use a fixed scale with the actual maximum and minimum, but if the data isn't occupying at least 20% of the vertical space, adjust the scale so that we're using 20% of the vertical space. Also, make room for green anomaly bars. The data will no longer crash into the green anomaly bars.

It will be easy to change any of this policy now that it's all hooked up.

This involved changing ModelData.jsx so that more information is available to the `yScaleCalculate` callback.

I made the plugin code a little clearer -- we just import it, we don't need to get the Dygraph function from it.


mrcslws:
cc @marionleborgne

marionleborgne:
@mrcslws here's my feedback:
* data does not crash into anomaly bars: 👍 
* Y-axis scaling: I think there is a problem with the way you scale the data. See attached screenshots and test dataset. 

The data is mostly flat except that there is a spike in the middle. Before the spike, you scale correctly. During the spike too. But after the spike you keep the Y axis range of when the spike was there, even if the data in the current window has a much smaller y-axis range.

Before the spike 👍 
screen shot 2016-05-10 at 3 33 11 pm
screen shot 2016-05-10 at 3 33 12 pm

after the spike 👎 
screen shot 2016-05-10 at 3 33 17 pm

marionleborgne:
FYI: Don't merge this PR for now. This required further discussion about the behavior of the Y-axis. @mrcslws, @lscheinkman and I will discuss tomorrow.

mrcslws:
This is ready for merge, since I pulled out the "use at least 20% vertical space" policy and just use the absolute max/min. It still makes space for the anomaly bars. We're still discussing the right y-scale approach.

marionleborgne:
LGTM 👍 Thanks @mrcslws 

We'll talk about re-introducing the "use at least 20% vertical space" policy later.

PR #771, opened 2016-05-09
UNI-419: Add zoom to charts

lscheinkman:
@marionleborgne Please review.
It includes both zoom handles and shortcuts.


mrcslws:
I'm playing with this now. It's nice!

One part that feels weird: using "point-and-click" on the time navigation bar causes the zoom to reset. It would be better to preserve the zoom. Also, this first click after zoom gets positioned incorrectly, but this will be fixed if we just preserve the zoom.

unicorn/app/browser/components/ModelData.jsx
```diff
@@ -348,13 +380,68 @@ export default class ModelData extends React.Component {
...
+  }
+
+  /**
+   * Describe time zoom level rounding time rounded to the closest time
+   * description (minute,  hour, day, week, month, ...) based on metric data.
+   * @param  {number} zoomLevel Percentage of data to display. [0..1]
+   * @return {string}           Human readable time period description.
+   *                            For example:
+   *                              All
+   *                            	15 minutes
```
mrcslws: Note that there are some \t tabs here. So it's misaligned in some editors (e.g. github)

marionleborgne:
@lscheinkman i have 2 comments:
1. Same comment as @mrcslws: we should preserve the zoom when point-and-cliking
2. if I click on "All" or any other zoom level that's not the max zoom level, I get something like that:
screen shot 2016-05-09 at 11 21 52 am
The green anomaly bars are incorrect, since we don't have anomaly results yet for points that have not been processed yet.

There are 2 options to solve this problem:
1. Disable other zoom options while the model is still outputing results to 
2. Fix how anomaly bars are being displayed (don't display bars if we don't have model results yet)

mrcslws:
Here's an issue: toggling and untoggling "Show non-aggregated data" causes flickering.

unicorn/app/browser/components/ModelData.jsx
lscheinkman: Interesting. It passed `eslint` with both `"indent": [2, 2]` and `no-mixed-spaces-and-tabs` rules.
I guess I cannot trust `eslint` to catch tab issues.

marionleborgne:
@mrcslws can you be more specific about the last issue you reported? I can't reproduce this problem, it works fine for me (NYC taxi data)

marionleborgne:
After discussion we decided to hide the zoom shortcuts until the model finishes

marionleborgne:
@lscheinkman To reproduce the flickering mentioned in https://github.com/numenta/numenta-apps/pull/771#issuecomment-217948760: 
* Select the nyc_taxi.csv data
* Create a model
* Select 'Show non-aggregated data'
* Flickering will happen if:
** You move zoom window with point and click
** You click on the zooming shortcuts

mrcslws:
FWIW, Step 4

> If 
>
> - You move zoom window with point and click
> - You click on the zooming shortcuts

isn't necessary on my computer. It might be a timing issue, maybe my CPU is slower / busier. If I launch HTM studio, show the NYC taxi data, click the checkbox, it happens.

Maybe these are all the same issue.

marionleborgne:
2 more comments:
1. The zoom shortcut text changes depending on wether or not the aggregated data checkbox is selected 
screen shot 2016-05-11 at 3 07 34 pm
screen shot 2016-05-11 at 3 07 40 pm

2. The zoom shortcut text should say `1 day` instead of `a day` to be consistent since we use numbers everywhere else

marionleborgne:
LGTM 👍  once conflicts with master are resolved, tag me and I'll test one last time.

With this PR, we now have:
* Custom zooming
* Zooming shortcuts

In master we already have:
* Point and click while the model is running

To be able to close the UNI-419 JIRA, we are still missing:
* Drag left/right by grabbing the time navigation window at the bottom of the charts
* Drag left/right by grabbing the chart main window

marionleborgne:
@lscheinkman My bad, my 2 last points are part of UNI-432, so disregard this part of my comment

> * Drag left/right by grabbing the time navigation window at the bottom of the charts
> * Drag left/right by grabbing the chart main window

PR #772, opened 2016-05-09
UNI-432: Unlock scroll while the model is active

lscheinkman:
@marionleborgne Please review


marionleborgne:
The point and click will reposition where the time navigation viewer is. That's good.

However, dragging either the navigation rectangle or the chart to the left / right does not work. The chart tries to go back to where it was originally and this causes flickering.

PR #773, opened 2016-05-10
UNI-439 Render anomaly bars grey during probation period, show "N/A" …

oxtopus:
…for anomaly value in app, and in export during probation period


marionleborgne:
Thanks @oxtopus it's great like that 👍

PR #774, opened 2016-05-11
UNI-424: updated heuristic to avoid connecting dots across big gaps of data

marionleborgne:
Based off @mrcslws PR #761 

1. Updated gap threshold heuristic: only the big gaps are taken into account. disregard the small gaps (a gap of max 2 anomaly bars is allowed). 

2. make sure to have the same gap threshold across timeseries. I took this logic out of the insertIntoGaps() function and compute the gap threshold only one time, on the raw data.


unicorn/app/browser/components/ModelData.jsx
```diff
@@ -223,6 +275,13 @@ export default class ModelData extends React.Component {
       options: {
         axisLineColor: muiTheme.rawTheme.palette.accent4Color,
         connectSeparatedPoints: true,  // required for raw+agg overlay
+
```
marionleborgne: @mrcslws can we remove this?

mrcslws: Sure, we'll see if it ever annoys us not having the edge points. I have no idea when they'll release the next dygraphs version, so I like the idea of just giving up on `drawGapEdgePoints` for now.

mrcslws:
After switching from minimum to median (or something similar), you have my 👍

unicorn/app/browser/components/ModelData.jsx
marionleborgne: agree. removed it.

marionleborgne:
Addressed your last comment @mrcslws. Thanks. Merging now.

PR #780, opened 2016-05-12
Gap detection: handle case where aggregation window is wide

mrcslws:
Replaces #776 

/cc @marionleborgne


marionleborgne:
LGTM 👍

PR #781, opened 2016-05-12
ENG-105 Fix DATETIME and TIMESTAMP mysql column defaults to work with MySQL 5.7.8+

### LINKED ISSUE #779: Taurus and htm.it don't work with MySQL 5.7.8+
Taurus and htm.it don't work with MySQL 5.7. For example after installing mysql 5.7.10 on Mac OS X, {{taurus-create-db}} results in the following exception:

```
sqlalchemy.exc.OperationalError: (OperationalError) (1067, "Invalid default value for 'timestamp'") "\nCREATE TABLE instance_status_history (\n\tserver VARCHAR(100) DEFAULT '' NOT NULL, \n\ttimestamp TIMESTAMP DEFAULT '0000-00-00 00:00:00', \n\tstatus VARCHAR(32) DEFAULT '' NOT NULL, \n\tPRIMARY KEY (server, timestamp)\n)\n\n" ()
```

This is the first one that migration encounters. There are actually several DATETIME and TIMESTAMP columns in the migration version scripts in Taurus.Engine and HTM.IT with either `0000-00-00 00:00:00` or `0` DEFAULT values.

This is related to mysql’s wishy-washy position on the NO_ZERO_DATE flag that has gone and come again starting with 5.7.4. See "SQL Mode Changes in MySQL 5.7" http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-changes. Apparently, as of mysql 5.7.8, NO_ZERO_DATE is back in the default mode set. Apparently, you can get rid of the mentioned error by removing NO_ZERO_DATE from mysql’s config file (my.cnf/my.ini?). 

According to the above-mentioned mysql note, "The long term plan is still to have the three affected modes be included in strict SQL mode and to remove them as explicit modes in a future MySQL release. ", but they have been going back and forth to give affected apps additional time to modify their schema.

I think that we need to modify the affected schema to use datetime column attributes that are compatible with mysql 5.6 and 5.7.8+ and don’t break the application logic (update affected logic, if needed), in case something in the logic relies on the original default values.

We need to take into consideration the following scenarios that our users may experience with default mysql installation options (e.g., default options as installed by homebrew); remember that the user's choice of mysql version is not driven solely by numenta software, and considerations such as mysql security fixes and compatibility and features needed by their other apps are critical.
1. Installing Taurus.Engine (or HTM.IT) against MySQL 5.6 (e.g., 5.6.23) for the first time
2. Installing Taurus.Engine (or HTM.IT) against MySQL 5.7.8+ (e.g., 5.7.10) for the first time
3. Migrating (e.g., graceful-update-production-taurus-servers) Taurus.Engine (or HTM.IT) to the new version of migration scripts with fixed TIMESTAMP and DATETIME columns against MySQL 5.6 (e.g., 5.6.23)
4. Migrating (e.g., graceful-update-production-taurus-servers) Taurus.Engine (or HTM.IT) to the new version of migration scripts with fixed TIMESTAMP and DATETIME columns against MySQL 5.7.8+ (e.g., 5.7.10)
5. Original deployment against MySQL 5.6 (e.g., 5.6.23) without the new TIMESTAMP/DATETIME column definitions, then upgrade to mysql 5.7.8+ (e.g., 5.7.10), then update Taurus.Engine (same for HTM.IT) to software with new fixed TIMESTAMP/DATETIME column definitions, and perform database migration

NOTE: you may encounter what appears to be a bug http://bugs.mysql.com/bug.php?id=80163 in MySQL 5.7 (5.7.10 definitely) that affects the "annotation" table htm.it/htm/it/app/repository/migrations/versions/000_3a7e06671df4_base_to_1.6.py, where there are multiple TIMESTAMP columns without defaults. You may need to find suitable defaults to work around that issue, being careful not to break application logic that might rely on the builtin default values.

vitaly-krugl:
Fixes #779 

Fix server defaults for DATETIME and TIMESTAMP columns in taurus.engine, because
0 ("0000-00-00 00:00:00") is deprecated as default for those colum types
as of mysql 5.7.8, and will fail with mysql installed with default config

Fix server defaults for DATETIME and TIMESTAMP columns in htm.it, because
0 ("0000-00-00 00:00:00") is deprecated as default for those colum types
as of mysql 5.7.8, and will fail with mysql installed with default config.

Point tests in app/runtime/aggregator_metric_collection_test.py that require running instances to
the us-west-2 region, since we have no more running test EC2 instances in us-east-1

Fix default in instance_status_history.timestamp column of the htmengine's sleleton-app database.

### DIFF
--- a/htm-it-mobile/pipeline/requirements.txt
+++ b/htm-it-mobile/pipeline/requirements.txt
@@ -1,4 +1,4 @@
 boto==2.27.0
-fabric==1.9.0
+fabric==1.11.1
 jenkinsapi==0.2.25
 PyYAML==3.10
--- a/htm.it/htm/it/app/repository/migrations/versions/000_3a7e06671df4_base_to_1.6.py
+++ b/htm.it/htm/it/app/repository/migrations/versions/000_3a7e06671df4_base_to_1.6.py
@@ -1,6 +1,6 @@
 # ----------------------------------------------------------------------
 # Numenta Platform for Intelligent Computing (NuPIC)
-# Copyright (C) 2015, Numenta, Inc.  Unless you have purchased from
+# Copyright (C) 2015-2016, Numenta, Inc.  Unless you have purchased from
 # Numenta, Inc. a separate commercial license for this software code, the
 # following terms and conditions apply:
 #
@@ -68,7 +68,7 @@ def upgrade():
       sa.Column("server", sa.VARCHAR(length=100), server_default="",
                 nullable=False),
       sa.Column("timestamp", sa.TIMESTAMP(),
-                server_default="0000-00-00 00:00:00", nullable=False),
+                server_default=None, nullable=False),
       sa.Column("status", sa.VARCHAR(length=32), server_default="",
                 nullable=False),
       sa.PrimaryKeyConstraint("server", "timestamp")
@@ -86,7 +86,9 @@ def upgrade():
       "annotation",
       sa.Column("uid", sa.VARCHAR(length=40), nullable=False),
       sa.Column("timestamp", sa.TIMESTAMP(), nullable=False),
-      sa.Column("created", sa.TIMESTAMP(), nullable=False),
+      sa.Column("created", sa.TIMESTAMP(),
+                server_default=sa.text("CURRENT_TIMESTAMP"),
+                nullable=False),
       sa.Column("device", sa.VARCHAR(length=40), nullable=True),
       sa.Column("user", sa.VARCHAR(length=100), nullable=True),
       sa.Column("server", sa.VARCHAR(length=100), nullable=True),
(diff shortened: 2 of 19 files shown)


vitaly-krugl:
https://ci.numenta.com/browse/TAUR-TAUR2-1 passed

vitaly-krugl:
CC @scottpurdy Please review

scottpurdy:
I can't speak for whether this covers all the relevant code but it looks good 👍 

Are there any tests for the migrations though? How do we validate that the changes and additional migrations work?

vitaly-krugl:
@scottpurdy There are no specific tests for migration from one version to the next in our CI test suite. I tested migration from 001 to 002 versions manually on my dev laptop.

In CI, all versions (000 - 002) are executed as part of the database initialization.

scottpurdy:
👍

PR #782, opened 2016-05-12
Fix package names

scottpurdy:
@oxtopus @vitaly-krugl 

Suggestions for testing are welcome. I was able to run tests and haven't found any issues but I'd be surprised if there aren't still some lingering.


oxtopus:
I set up a plan branch for this at https://ci.numenta.com/browse/TAUR-TAUR3-1

It's tracking my fork, but you may want to configure it to track yours to iterate on changes.

scottpurdy:
@oxtopus thanks

oxtopus:
The bamboo build tasks and deployment plans will need to be fixed to use the new paths.

oxtopus:
LGTM, but need to have a plan for the pipelines before I'm comfortable merging since they will certainly break as they are.

scottpurdy:
@oxtopus I got everything working in a plan branch. What else should I be looking into? The deployment setup in Bamboo definitely needs a look. Do you think we should work that out in staging now?

scottpurdy:
https://ci.numenta.com/browse/TAUR-TAUR3

oxtopus:
It's your call.  We can prepare and configure a new environment relatively easily, but really, we can work out whatever we need in staging.  That's what it's there for.

vitaly-krugl:
💯

PR #783, opened 2016-05-13
UNI-445 "About HTM Studio" and "Provide Feedback"

mrcslws:
cc @marionleborgne


marionleborgne:
LGTM 👍 thanks @mrcslws

marionleborgne:
@lscheinkman this is what the about menu looks like:
screen shot 2016-05-12 at 11 50 27 pm

I think it looks fine. it's not a material-ui menu, so i thought i would just run it by you.

When we have all the right links (EULA and Feedback form) live on numenta.com i'll add/update the links accordingly.

marionleborgne:
When we have all the right links (EULA and Feedback form) live on numenta.com i'll add/update the links accordingly. But right now we can merge, since the links won't be up until next monday.

brev:
@marionleborgne numenta.com URLs 
https://jira.numenta.com/browse/WEB-76?focusedCommentId=52108&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-52108

marionleborgne:
Thanks @brev. I'll update the links in a separate PR

PR #784, opened 2016-05-13
UNI-404 Zoom the y scale when the data is scrunched

mrcslws:
When a dataset has a huge outlier, it can mess up the experience of viewing the rest of the chart. This change improves that experience, though it's still not perfect. I'm not sure perfect is possible.

/cc @marionleborgne


marionleborgne:
The plan is to hold the discussion about this PR until #803 and then #797 are merged

marionleborgne:
👍 thanks @mrcslws !!

PR #785, opened 2016-05-13
Fix Advanced Settings Bugs

flysonic10:
- Fix bug that causes model to run without aggInfo
- Fix bug where the CreateModelDialog state is not reset when re-opening the dialog


marionleborgne:
👍 🙏 great, it fixes the 2 bugs

cc @mrcslws

PR #788, opened 2016-05-14
Zooming: don't assume equally spaced data

mrcslws:
Calculating in terms of "pointsPerBar" only works in some cases. It doesn't always work, e.g if there are gaps in the time series.

The new way of thinking: Split the timeline into equal-width buckets, based on the zoom level, and draw 0 or 1 anomaly bars per bucket.

This solves the general problem of collapsing anomaly bars into each other. For example, previously with non-aggregated data the anomaly bars would sometimes collide. Now they'll collapse into each other rather than collide.

@marionleborgne @lscheinkman 

This doesn't solve:

- Dygraphs doesn't let us set a "zoom limit", so on small data sets, you can zoom in further than we'd like users to zoom in.
  - We might be able to solve this by adding another feature to Dygraphs via our new CustomDygraphs class.
  - Or we could have our AnomalyBarChartUnderlay change the zoom and force a redraw when it notices a small zoom. But this will make the experience pretty janky. A draw shouldn't queue a redraw.
- Dygraphs sets its own "zoom limit" just based on the range selector size, so on large data sets you can't zoom in far enough to see one anomaly bar per point. In general with HTM Studio you'll see more points than anomaly bars.
  - You can see this in the Dygraphs demos: http://dygraphs.com/tests/range-selector.html  The zoom limit is set by the size of the range finder.
  - We could probably force a smaller zoom programmatically, and we could make this one of the blue links at the top.

### DIFF
--- a/unicorn/app/browser/components/Chart.jsx
+++ b/unicorn/app/browser/components/Chart.jsx
@@ -157,7 +157,8 @@ export default class Chart extends React.Component {
     options.dateWindow = this._chartRange;  // update viewport of range selector
     this._previousDataSize = data.length;
     this._dygraph = new CustomDygraph(element, data, options,
-      this.props.yScaleCalculate);
+                                      this.props.xScaleCalculate,
+                                      this.props.yScaleCalculate);
 
     // after: track chart viewport position changes
     rangeEl = element.getElementsByClassName(RANGE_SELECTOR_CLASS)[0];
--- a/unicorn/app/browser/components/Chart.jsx
+++ b/unicorn/app/browser/components/Chart.jsx
@@ -24,11 +24,21 @@ import ChartUpdateViewpoint from '../actions/ChartUpdateViewpoint';
 import Dygraph from 'dygraphs';
 import '../lib/Dygraphs/Plugins';
 import CustomDygraph from '../lib/Dygraphs/CustomDygraph';
-import {DATA_FIELD_INDEX} from '../lib/Constants';
+import {ANOMALY_BAR_WIDTH, DATA_FIELD_INDEX} from '../lib/Constants';
 
 const {DATA_INDEX_TIME} = DATA_FIELD_INDEX;
 const RANGE_SELECTOR_CLASS = 'dygraph-rangesel-fgcanvas';
 
+function getDateWindowWidth(resolution, chartElement) {
+  switch (resolution.per) {
+  case 'anomaly bar':
+    return resolution.timespan * (chartElement.offsetWidth / ANOMALY_BAR_WIDTH);
+  case 'chart width':
+    return resolution.timespan;
+  default:
+    throw new Error(`Unrecognized resolution 'per': ${resolution.per}`);
+  }
+}
 
 /**
  * Chart Widget. Wraps as a React Component.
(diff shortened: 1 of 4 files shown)


marionleborgne:
Do you mean that now we don't have a max zoom limit ? I. e. that the a anomaly bars can be very spread out? If so, this is a problem that will prevent this PR form being merged as this is a requirement for the app.

lscheinkman:
Not setting the zoom limit may leave chart with holes. See below:
screen shot 2016-05-14 at 4 36 53 pm

screen shot 2016-05-14 at 4 37 14 pm

mrcslws:
Ok, I added max-zoom. Here's the commit message:

> **Add a mediocre max-zoom feature to Dygraphs**
> 
> This is similar to the experience we had before, except:
> 
> - It no longer flickers the chart.
> - It works when the data isn't uniformly distributed.
> 
> But it's still not good:
> 
> - It flickers the time navigation bar.
> - If you drag the gripper from the left and hit the max-zoom, your
>   zoom causes it to pan to the right.

It will be hard to solve these issues.

These issues exist in master, so I think this PR now has no drawbacks, though you'll probably want to QA it.

marionleborgne:
Let's not merge this branch until we really validate that this is not breaking anything. Currently the charts are broken when the model is running.

screen shot 2016-05-14 at 7 50 36 pm

This is with the NYC taxi data.

marionleborgne:
@mrcslws I confirmed that with your fix https://github.com/numenta/numenta-apps/pull/787 the problem described in https://github.com/numenta/numenta-apps/pull/788#issuecomment-219263296 does not happen 👍

marionleborgne:
Little recap about where this is at: there's a lot f progress. The dataset "missing_values.csv" is now nicely rendered - this is a first with custom zooming 👍 

Here is a screenshot:
mising_data


And here is the CSV if you want to test:
[missing_data.csv.zip](https://github.com/numenta/numenta-apps/files/264852/missing_data.csv.zip)

marionleborgne:
FYI, @lscheinkman: @mrcslws and I synced. He's going to continue working on this branch to stabilize master while I work on a branch called `htm-studio-private-beta` which is basically master prior to custom zooming but with @mrcslws point-and-click fix (https://github.com/numenta/numenta-apps/pull/787), so that we have a fallback in case it's too hard / long to stabilize master with the custom zooming in this PR before monday.

marionleborgne:
FYI, https://github.com/numenta/numenta-apps/pull/788/commits/b900f572b8ef5841ea9b1aca9beae0895914d4ab fixes:
screen shot 2016-05-14 at 10 49 38 pm

If you want to try to reproduce the bug to make sure it is gone:
* Use [network_stats.csv.zip](https://github.com/numenta/numenta-apps/files/264873/network_stats.csv.zip)
* Create a model and let it finish
* Move the navigation window all the way to the left
* Drag the handle all the way to right

mrcslws:
FYI, here's some by-design behavior that might seem like a bug when you see it.

When I say "by-design", I'm referring to my design. :)

image

This is a non-aggregated model. When you see this, it means somewhere in the timeline you'll find something like this:

image

The two data points at 3:00am and 3:01am are causing this. The max-zoom level is set using the model result deltas. You can zoom until the two closest model results have their own anomaly bar.

Also, this has nothing to do with daylight savings. It just happens to be that that's where this CSV has funky data.

PR #789, opened 2016-05-15
Compute probationary width from timestamps, not indices

mrcslws:
This adds support for irregularly-spaced data.

cc @marionleborgne


marionleborgne:
It looks like it fixes the bug 👍 

* Before:
screen shot 2016-05-15 at 12 00 45 am

* After:
screen shot 2016-05-15 at 12 35 12 am

PR #793, opened 2016-05-15
Fix model results summary

marionleborgne:
Problem 
* The HTM Settings at the bottom of the model summary dialog were incorrect. Except after you refresh the app.

Solution:
* This is happened because the DB was updated but not the metricStore - when the model is started.
* The fix consisted in dispatching an action to update the metricStore with param finder results for the specific model/metric that is started.

cc @mrcslws @lscheinkman


mrcslws:
I'm playing with it now. Looks good, it fixes the bug. 👍 

I don't know much about the db / store yet, but the code seems right.

PR #795, opened 2016-05-16
Second stable release for private beta (with custom zoom)

marionleborgne:
Branches merged in master (at https://github.com/numenta/numenta-apps/commit/b6d63d19d3af7648329ecdc22a3aecf5b2ca85b9)
* https://github.com/numenta/numenta-apps/pull/793
* https://github.com/numenta/numenta-apps/pull/792
* https://github.com/numenta/numenta-apps/pull/789
* https://github.com/numenta/numenta-apps/pull/788

We need two more fixes to make this branch stable for the beta:
1. @mrcslws: Fix zoom reset when "show non-aggregated data" checkbox is selected / unselected
2. @marionleborgne: Disable point-and-click when model is running [1]

[1] since there is currently no fix for the bug: "left/right panning is broken when the model is running"

### DIFF
--- a/unicorn/app/browser/actions/StartModel.js
+++ b/unicorn/app/browser/actions/StartModel.js
@@ -34,6 +34,9 @@ export default function (actionContext, payload) {
     let {aggOpts, inputOpts, metricId, modelOpts} = payload;
     let aggregated = (Object.keys(aggOpts).length >= 1);
 
+    // Update metricStore
+    actionContext.dispatch(ACTIONS.UPDATE_METRIC, payload);
+
     // Save to database
     let db = actionContext.getDatabaseClient();
     return promisify(::db.updateMetric, metricId, {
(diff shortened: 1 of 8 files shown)


mrcslws:
The "temporarily disable point-and-click when model is running" commit looks good. I verified it works. 👍

marionleborgne:
This branch is now stable for a private beta release. Thanks @mrcslws for helping stabilize.

@lscheinkman : FYI, the only feature I had to cut to get to this stable state is "allow point-and-click when the model is running". We can re-introduce this once we are able to fix the right/left panning when the model is running. Until then, I'll say let's leave the point-and-click out  (it still works after the model completes)

Issues that remain (but non-blockers for private beta)
* Some sample files are missing in the left menu the first time the app is run (because the fileStore does not get updated correctly)
* The sorting of the left and right menu is random
* Daylight savings (UNI-440)

PR #801, opened 2016-05-18
UNI-440 On export, use original time zone

mrcslws:
This change adds more fields to the database, but it does not grow the store. These ISO timestamps live in the database, and they're only used during export. The MetricData timestamp is never used, but this keeps things consistent, and we have the option of using it later.

For running models, this change moves the timestamp parsing upstream. Before sending a timestamp to the client, node will compute the "naive time". So we're no longer sending a timestamp string through IPC.

@marionleborgne @lscheinkman


unicorn/app/database/schema/MetricData.json
```diff
@@ -9,7 +9,13 @@
       "minLength": 1,
       "required": true
     },
-    "timestamp": {
```
lscheinkman: @mrcslws To avoid problems with current **private beta users** due to the database schema changes, you should use a different database name.

mrcslws:
Alright, this is ready. I changed the database name. @marionleborgne @lscheinkman

marionleborgne:
👍 it does keep the TZ info when the data is being exported.

I still get the gap in network_stats.csv 
[network_stats.csv.zip](https://github.com/numenta/numenta-apps/files/283165/network_stats.csv.zip)
screen shot 2016-05-25 at 4 24 32 pm

mrcslws:
@marionleborgne it's just displaying what's in the the CSV.

mrcslws:
Here's an excerpt from that CSV:

~~~
2014-03-09 01:51:00,45.916,45.916
2014-03-09 01:56:00,44.038,44.038
2014-03-09 03:00:00,44.612,44.612
2014-03-09 03:00:00,43.578,43.578
2014-03-09 03:00:00,47.018,47.018
2014-03-09 03:00:00,46.456,46.456
2014-03-09 03:00:00,44.368,44.368
2014-03-09 03:00:00,43.544,43.544
2014-03-09 03:00:00,44.938,44.938
2014-03-09 03:00:00,43.834,43.834
2014-03-09 03:00:00,47.026,47.026
2014-03-09 03:00:00,42.368,42.368
2014-03-09 03:00:00,44.468,44.468
2014-03-09 03:00:00,47.09,47.09
2014-03-09 03:01:00,45.962,45.962
2014-03-09 03:06:00,44.656,44.656
~~~

marionleborgne:
Thanks for the clarification @mrcslws. LGTM 👍

PR #802, opened 2016-05-20
UNI-416: Automatic Updates

lscheinkman:
@marionleborgne Please review the text in [default.json](https://github.com/numenta/numenta-apps/compare/master...lscheinkman:UNI-416?expand=1#diff-07271de6594d3bf41b9d9977ee726296R5).
Keep in mind it is not easy to test this PR, as it only works with the signed apps.
I will demo it on Friday and I can show what it looks like from the user perspective once you're back.


lscheinkman:
CC: @vitaly-krugl

unicorn/README.md
```diff
@@ -481,3 +481,26 @@ does [not yet work with Electron](https://github.com/atom/electron/issues/915).
 * Update `app/package.json` with new version number
```
vitaly-krugl: Unrelated to this PR, but noticed that the `ECMAScript 7` URL in `Also a few ECMAScript 7 features` yields "404 Page not found"

unicorn/README.md
```diff
@@ -481,3 +481,26 @@ does [not yet work with Electron](https://github.com/atom/electron/issues/915).
...
+
+  Sample manifest file:
+  * *update.0.0.2.json*
+  ```javascript
+  {
+    url: "http://public.numenta.com/updates/htmstudio/darwin/HTM%20Studio-0.0.3-mac.zip",
+    name: "HTM Studio 0.0.3",
+    notes: "* Automatic Updates \n* Improved Charts \n* Fix timezone when exporting results",
+    pub_date: "2016-02-02T21:51:58Z"
+  }
```
vitaly-krugl: Nice explanation

unicorn/app/config/default.json
```diff
@@ -2,6 +2,13 @@
   "title": "HTM Studio",
   "desc": "HTM Studio Desktop Application",
   "company": "Numenta",
+  "update": {
+    "readyTitle": "Software Updated ready",
```
vitaly-krugl: Updated ==> Update ?

unicorn/app/main/AutoUpdate.js
```diff
@@ -0,0 +1,133 @@
...
+    autoUpdater.checkForUpdates();
+  }
+
+  /**
+   * Handle squirrel windows start up process
+   * @return {boolean} Whether or not an update was applied at start up
+   */
+  _handleSquirrelWindowsEvents() {
+    if (process.platform === 'win32') {
+      if (process.argv.length === 1) {
```
vitaly-krugl: Would be nice to have a comment here explaining this special case.

unicorn/app/main/AutoUpdate.js
```diff
@@ -0,0 +1,133 @@
...
+   */
+  checkForUpdates() {
+    if (this._handleSquirrelWindowsEvents()) {
+      return;
+    }
+    autoUpdater.checkForUpdates();
+  }
+
+  /**
+   * Handle squirrel windows start up process
```
vitaly-krugl: Would be helpful to explain why Windows is special-cased versus `autoUpdater.checkForUpdates`

unicorn/app/main/AutoUpdate.js
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Adds support for automatic Updates using Electron's AutoUpdater
+ */
+export default class AppUpdater {
+
+  constructor(browserWindow) {
+    this._window = browserWindow;
+    let updateUrl = config.get('update:url');
+    let feedUrl = `${updateUrl}/${os.platform()}/update.${app.getVersion()}.json`;
```
vitaly-krugl: From this, it looks like if the current app is v 0.0.1 and the most recent update is 0.0.3, then the app will undergo two updates: 0.0.1 ==> 0.0.2, and then 0.0.2 ==> 0.0.3 to become current. Will there be two separate user interactions, one for each update cycle, in this case, or will auto-update perform both update cycles with a single user interaction?

unicorn/app/config/default.json
```diff
@@ -2,6 +2,13 @@
...
+    "errorTitle": "Update Error",
+    "message": "A new update is ready to install.\nWould you like to install it now?",
+    "detail": "What's new in %releaseName:\n%releaseNotes",
+    "url": "http://public.numenta.com/updates/htmstudio"
```
vitaly-krugl: Since the app may be forked by users to create their own apps, would it perhaps be better to add the numenta-specific update URL as part of the build rather than storing it in the repo?

unicorn/app/main/AutoUpdate.js
lscheinkman: It depends on the `update.json` file. The auto update downloads the package referenced by the URL and installs it. We can setup the `update.json` to always point to the `latest` version so it would perform only the transition to the latest version

unicorn/README.md
lscheinkman: This file was initially created before ES7 and only updated for the new additions. Feel free to update the missing references. Thanks

unicorn/app/main/AutoUpdate.js
```diff
@@ -0,0 +1,133 @@
...
+
+    autoUpdater.addListener('update-downloaded', (event,
+      releaseNotes, releaseName, releaseDate, updateURL) => {
+      // Format release notes if given. May not be avaialbe on all platforms
+      let detail = null;
+      if (releaseNotes) {
+        detail = config.get('update:detail')
+                       .replace('%releaseName', releaseName)
+                       .replace('%releaseNotes', releaseNotes)
+                       .replace('%releaseDate', releaseDate);
```
vitaly-krugl: It seems like this should fail, because `update:detail` doesn't have a token for releaseDate: `"detail": "What's new in %releaseName:\n%releaseNotes"`.

This scenario needs to be tested somehow; a mock unit test should do.

unicorn/app/main/AutoUpdate.js
```diff
@@ -0,0 +1,133 @@
...
+   */
+  _handleSquirrelWindowsEvents() {
+    if (process.platform === 'win32') {
+      if (process.argv.length === 1) {
+        return false;
+      }
+
+      const squirrelEvent = process.argv[1];
+      switch (squirrelEvent) {
+      case '--squirrel-install': // no-break
```
vitaly-krugl: Case statements are not indented under switch. Is it not the convention in JS?

vitaly-krugl: nit regarding "no-break" comment. I think this is commonly referred to as "fallthrough", at least in the C language.

vitaly-krugl:
Done with the code review; there is a bug in the update details composition; also other feedback.

unicorn/app/main/AutoUpdate.js
lscheinkman: This is will not fail, see [String.replace](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace). This is working as intended. 

`update:detail` message is free to use any of the passed parameters in its message, in this instance of `default.json` the detail message does not require the `releaseDate`.

unicorn/app/main/AutoUpdate.js
lscheinkman: The `switch/case` indentation is following our current JavaScript style guide. See https://github.com/numenta/numenta-apps/wiki/Javascript-Coding-Standards
The `switch/case` rule is mentioned [here](http://javascript.crockford.com/code.html)

unicorn/app/config/default.json
lscheinkman: This file is numenta specific by definition, if someone would like to fork the app and create their own update site all they have to do is follow the [README](https://github.com/lscheinkman/numenta-apps/blob/a11a1e9b3947ac8a23496e977b520afe212c31b3/unicorn/README.md#automatic-updates) instructions

unicorn/app/main/AutoUpdate.js
lscheinkman: @vitaly-krugl Added link to original README from the installer to the documentation. Let me know if it is clear.

marionleborgne:
I reviewed the flow in the demo video.

I think we should make it more clear what happens when you click on "yes" and "no". It's actually unclear to me in the video, maybe you can clarify? here is what I think happens:
* Clicking on "yes" will (1) quit the app, (2) install the update, (3) the user needs to start the app again manually.
* Clicking on "no" will (1) not do anything, the user can continue using the app. (2) when the user quits the app, the update will be installed and (3) the user needs to start the app again manually to see the new changes.

I think we should use the text space in the window to explain what will happen to the user and we can reference the changelog in a blog post / release page on github / separate window. See how Sublime does it for example:
screen shot 2016-05-25 at 9 22 53 am
screen shot 2016-05-25 at 9 23 11 am
screen shot 2016-05-25 at 9 23 18 am
screen shot 2016-05-25 at 9 23 33 am
screen shot 2016-05-25 at 9 24 10 am

My point is that the effect of the "yes" and "no" buttons & wether or not you should restart the app should be made more explicit to the user.

marionleborgne:
One more question: will the popup keep showing up if the user does not install the update?

marionleborgne:
Another question: what happens if the user starts the app before the install of the update completed?

lscheinkman:
@marionleborgne `electron`'s automatic update is based on [Squirrel.Mac](https://github.com/Squirrel/Squirrel.Mac), you can check the application update workflow [here](https://github.com/Squirrel/Squirrel.Mac#installing-updates). 
That should explain most of your workflow questions (https://github.com/numenta/numenta-apps/pull/802#issuecomment-221632210 and https://github.com/numenta/numenta-apps/pull/802#issuecomment-221632594) .

As for the notification dialog, please add a detailed design to the JIRA, I am not sure I fully understand your comment (https://github.com/numenta/numenta-apps/pull/802#issuecomment-221632173)

unicorn/app/main/AutoUpdate.js
vitaly-krugl: @lscheinkman, the special case of one argument (`process.argv.length === 1`) versus some other number of arguments is not at all clear to me. Could you please explain? As a reviewer or maintainer of the code, I would appreciate a comment in the code that helps me make some sense out of this by providing some guidance and clarification, in addition to providing a general link to the tool in the README.

marionleborgne:
Text for the dialog:

A new version of HTM Studio (<version>) is available. To install the update, the application needs to be closed. Do you want to close the application now or later?

Buttons:  `Wait until I close the app` (white raised button) | `Close and update now` (blue raised button)

screen shot 2016-05-25 at 3 23 06 pm

marionleborgne:
Done with the review 👍

lscheinkman:
@marionleborgne untitled

unicorn/package.json
```diff
@@ -33,8 +33,8 @@
     "clean:npm": "rimraf app/node_modules/ && rimraf node_modules/ && npm cache clean",
     "clean:python": "rimraf py/build && rimraf py/dist && rimraf py/*.egg-info",
     "clean:webpack": "rimraf app/browser/assets/bundle",
-    "electron": "electron app",
-    "electron:debug": "electron --debug=5858 app",
+    "electron": "NODE_ENV=development electron app",
```
oxtopus: This doesn't work on windows:

screen shot 2016-05-27 at 9 58 45 am

oxtopus: When I remove the env var, the app crashes:

screen shot 2016-05-27 at 10 01 54 am

PR #797, opened 2016-05-16
UNI-404 Use big round numbers on the y axis

mrcslws:
Note that it adds a new module. You can use `npm install d3-scale`.

@marionleborgne

PR #803, opened 2016-05-22
Use d3 for chart interaction

mrcslws:
Before this change, Dygraphs encapsulated a lot of the app's user interaction logic, so we weren't able to make the overall experience cohesive. This change takes control of the user interaction logic, and it treats Dygraphs as a simple "rendering layer".

Now that we have more control, here's what has changed:

- Consistent UI, regardless of whether the model is running.
  - You can always navigate.
  - You can always zoom.
  - You can show/unshow non-aggregated data.
- A polished "range selector" experience.
  - You can resize the selection as small as you want. It's no longer
    bound by the Dygraphs grippers colliding with each other.
  - We now have first-class max-zoom support, so the gripper doesn't
    flicker when you go below the max zoom.
  - In addition to dragging the selection, resizing the selection, and
    point-and-clicking the background, you can now "brush" the
    background to choose a range.
- The zoom shortcuts and chart are now cohesive / complementary.
  - As you change the selection, the zoom shortcuts become selected
    when they describe the current zoom level.
  - You can reliably zoom to precisely the maximum zoom level. You
    don't have to rely on the first zoom shortcut.
- Other bugfixes:
  - Hover text doesn't flicker as new model results arrive. It stays
    up-to-date with your hover position.
  - When you hover over a gap, it doesn't show a meaningless entry for
    the NaN gap marker. It now shows the nearest actual point.

Other improvements:

- We now show big round timestamps on the x axis. 🎉 
- When you hide and show a chart, you'll see the same time window as
  before. Previously it would lose the zoom level.
- When you hover over a nonaggregated point, it shows the nearest
  anomaly score. Recently this broke -- nonaggregated points were
  being treated as if they were in the probationary period because
  they didn't exactly match a model result.
- While a model is running, you can reenable "jump to new results" by
  moving the selection window to the end.

Noteworthy implementation details:

- I stopped storing javascript Dates in the store. Now we just work
  with javascript times (i.e. milliseconds since 1970). So datetime
  math \*just works\*.
- I removed the hacks that I've previously used to set the x
  scale, the y scale, to preserve the zoom level, and to dodge
  bugs in Dygraphs interaction code. We now do all of this the
  right way, without hacks.


mrcslws:
Notes:

- The y scaling on this isn't quite as good as before. After we merge https://github.com/numenta/numenta-apps/pull/797 I'll merge it into this change, then everything will be pretty.
- Our UI doesn't handle small browser window widths very well. This is true independent of this change, but it's worse with this change. Basically, the containers above the chart need to be better at enclosing their contents. I think this change is out of scope for this PR, but it deserves a fix.

@marionleborgne

marionleborgne:
It's really great. It fixes a lot of issues. Great job @mrcslws 👍 🎉 

The feedback that I have so far:
1. The middle zooming shortcut is incremented when the model is running. To reproduce: start a model with the NYC data and look at the middle shortcut.
screen shot 2016-05-24 at 10 25 23 am
screen shot 2016-05-24 at 10 25 27 am
2. I find the behavior of the navigation brush is a bit inconsistent with the navigation shortcuts: in most cases, when you resize with the shortcuts, it stays centered. But sometimes is resizes on the right or the left


Awesome:
* When you change the navigation window size, the zooming shortcuts are unselected
* You can see the non-aggregated data while the model is running
* There is no more flickering
* You can stop the model while it's running and drag left/right

marionleborgne:
After discussion @mrcslws  and I decided:
* Hide zooming shortcuts when model is running
* @mrcslws fixed exception
* Add padding at the top of the chart
* Always have the zoom resize in the middle

mrcslws:
Ok, I think this is good, @marionleborgne . Like we discussed, let's wait and add the y padding in https://github.com/numenta/numenta-apps/pull/797

marionleborgne:
CONGRATULATIONS AND A BIG THANK YOU 😄 👍 💯

PR #807, opened 2016-05-24
UNI-362: Add support for Google Analytics

lscheinkman:
Use "Google Analytics" to collect usage data, currently tracking most 'React' actions.

- Based on Google Analytics [Measurement Protocol]( https://developers.google.com/analytics/devguides/collection/protocol/v1/). 
- Works offline by caching events in the browser's [indexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) store.
- In development mode the events are not stored but can be viewed in the debugger
- In production mode a real Google Analytics tracking ID must be provided, see `README`.

@marionleborgne Please Review

CC: @vitaly-krugl

### DIFF
--- a/unicorn/app/browser/actions/DeleteModel.js
+++ b/unicorn/app/browser/actions/DeleteModel.js
@@ -28,11 +28,14 @@ import {ACTIONS} from '../lib/Constants';
  * @return {Promise}
  */
 export default function (actionContext, modelId) {
+  actionContext.getGATracker().event('ACTION', ACTIONS.DELETE_MODEL);
+
   return new Promise((resolve, reject) => {
     let database = actionContext.getDatabaseClient();
     // Delete model
     database.deleteModel(modelId, (error) => {
       if (error) {
+        actionContext.getGATracker().exception(ACTIONS.DELETE_MODEL_FAILED);
         actionContext.dispatch(ACTIONS.DELETE_MODEL_FAILED, {modelId, error});
         reject(error);
       } else {
--- a/unicorn/app/browser/actions/HideFileDetails.js
+++ b/unicorn/app/browser/actions/HideFileDetails.js
@@ -25,5 +25,7 @@ import {ACTIONS} from '../lib/Constants';
  * @emits {HIDE_FILE_DETAILS}
  */
 export default function (actionContext) {
+  actionContext.getGATracker().event('ACTION', ACTIONS.HIDE_FILE_DETAILS);
+
   actionContext.dispatch(ACTIONS.HIDE_FILE_DETAILS);
 }
--- a/unicorn/app/browser/entry.js
+++ b/unicorn/app/browser/entry.js
@@ -58,9 +58,11 @@ const {app} = require('electron').remote;
 
 const modelClient = new ModelClient();
 const paramFinderClient = new ParamFinderClient();
-const gaTracker = new GoogleAnalytics(config.get('GA_TRACKING_ID'),
+/* eslint-disable no-process-env */
+const gaTracker = new GoogleAnalytics(process.env.GA_TRACKING_ID,
                                       app.getName(), app.getVersion(),
-                                      config.get('NODE_ENV') === 'development');
+                                      process.env.NODE_ENV === 'development');
+/* eslint-enable no-process-env */
 
 /**
  * HTM Studio: Cross-platform Desktop Application to showcase basic HTM features
--- a/unicorn/app/browser/entry.js
+++ b/unicorn/app/browser/entry.js
@@ -39,6 +39,7 @@ import databaseClient from './lib/HTMStudio/DatabaseClient';
 import fileClient from './lib/HTMStudio/FileClient';
 import FileDetailsStore from './stores/FileDetailsStore';
 import FileStore from './stores/FileStore';
+import GoogleAnalytics from './lib/analytics/GoogleAnalytics';
 import HTMStudioPlugin from './lib/Fluxible/HTMStudioPlugin';
 import loggerConfig from '../config/logger';
 import MainComponent from './components/Main';
@@ -53,10 +54,13 @@ import {trims} from '../common/common-utils';
 
 const dialog = remote.require('dialog');
 const logger = bunyan.createLogger(loggerConfig);
+const {app} = require('electron').remote;
 
-let modelClient = new ModelClient();
-let paramFinderClient = new ParamFinderClient();
-
+const modelClient = new ModelClient();
+const paramFinderClient = new ParamFinderClient();
+const gaTracker = new GoogleAnalytics(config.get('GA_TRACKING_ID'),
+                                      app.getName(), app.getVersion(),
+                                      config.get('NODE_ENV') === 'development');
 
 /**
  * HTM Studio: Cross-platform Desktop Application to showcase basic HTM features
@@ -68,6 +72,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
   // global uncaught exception handler
   window.onerror = (message, file, line, col, error) => {
+    gaTracker.exception('Unknown Error', `Unknown Error: ${message}`);
     dialog.showErrorBox('Unknown Error', `Unknown Error: ${message}`);
   };
 
@@ -100,7 +105,8 @@ document.addEventListener('DOMContentLoaded', () => {
     databaseClient,
     fileClient,
     modelClient,
-    paramFinderClient
+    paramFinderClient,
+    gaTracker
   });
 
   // Start listening for model events
@@ -111,6 +117,10 @@ document.addEventListener('DOMContentLoaded', () => {
 
   // app exit handler
   window.onbeforeunload = (event) => {
+    // Synchronize google analytics
+    gaTracker.synchronize();
+
+    // Check for running model
     let models = context.getStore(ModelStore).getModels();
     let active = models.filter((model) => model.active === true) || [];
     let modelCount = active.length || 0;
@@ -144,6 +154,7 @@ document.addEventListener('DOMContentLoaded', () => {
       ReactDOM.render(contextEl, container);
     })
     .catch((error) => {
+      gaTracker.exception(error);
       console.log(error); // eslint-disable-line
       dialog.showErrorBox('Startup Error', `Startup Error: ${error}`);
     });
--- a/unicorn/app/browser/entry.js
+++ b/unicorn/app/browser/entry.js
@@ -72,7 +72,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
   // global uncaught exception handler
   window.onerror = (message, file, line, col, error) => {
-    gaTracker.exception('Unknown Error', `Unknown Error: ${message}`);
+    gaTracker.exception('Unknown Error');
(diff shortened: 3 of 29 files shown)


unicorn/app/browser/actions/DeleteModel.js
```diff
@@ -28,11 +28,14 @@ import {ACTIONS} from '../lib/Constants';
...
  */
 export default function (actionContext, modelId) {
+  actionContext.getGATracker().event('ACTION', ACTIONS.DELETE_MODEL);
+
   return new Promise((resolve, reject) => {
     let database = actionContext.getDatabaseClient();
     // Delete model
     database.deleteModel(modelId, (error) => {
       if (error) {
+        actionContext.getGATracker().exception(ACTIONS.DELETE_MODEL_FAILED);
```
vitaly-krugl: Too badf we cannot log the error info via GA, which could be helpful for identifying specific issues.

unicorn/app/browser/actions/HideFileDetails.js
```diff
@@ -25,5 +25,7 @@ import {ACTIONS} from '../lib/Constants';
  * @emits {HIDE_FILE_DETAILS}
  */
 export default function (actionContext) {
+  actionContext.getGATracker().event('ACTION', ACTIONS.HIDE_FILE_DETAILS);
```
vitaly-krugl: Too bad we couldn't just fold all these calls into `actionContext.dispatch`, instead of having to make them explicitly. It seems like the pattern is always the same.

unicorn/app/browser/entry.js
```diff
@@ -68,6 +72,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
   // global uncaught exception handler
   window.onerror = (message, file, line, col, error) => {
+    gaTracker.exception('Unknown Error', `Unknown Error: ${message}`);
```
vitaly-krugl: Might `${message}` contain unobfuscated information that the app should not be collecting (e.g., file or directory path that contains user name or entity/company name)?

unicorn/webpack.config.babel.js
```diff
@@ -15,14 +15,23 @@
...
+const DEVELOPMENT = process.env.NODE_ENV === 'development';
 
 /**
  * WebPack ES6 Config File
  */
 export default {
   bail: true,
+  plugins: [
+    new webpack.EnvironmentPlugin([
+      'NODE_ENV', 'GA_TRACKING_ID'
```
vitaly-krugl: Where is the code that sets `GA_TRACKING_ID` to a specific value?

unicorn/app/browser/lib/Analytics/GoogleAnalytics.js
```diff
@@ -0,0 +1,295 @@
...
+
+  /**
+   * Get all pending hits stored in the database
+   * @return {Promise} Promise wrapping database transaction.
+   *                           `resolve` : With all pending hits on success
+   *                           `reject`: transaction failed
+   */
+  _getAllPening() {
+    return new Promise((resolve, reject) => {
+      if (this._db) {
```
vitaly-krugl: Since there are not tests in this PR, how do we know that any of this will work reliably?

unicorn/app/browser/lib/Analytics/GoogleAnalytics.js
```diff
@@ -0,0 +1,295 @@
...
+// ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// FOR A PARTICULAR PURPOSE. See the GNU Affero Public License for more details.
+//
+// You should have received a copy of the GNU Affero Public License along with
+// this program.  If not, see http://www.gnu.org/licenses.
+//
+// http://numenta.org/licenses/
+
+let uuid = require('node-uuid');
+
```
vitaly-krugl: nicely commented

unicorn/app/browser/lib/Analytics/GoogleAnalytics.js
```diff
@@ -0,0 +1,295 @@
...
+      this.synchronize();
+    });
+  }
+
+  /**
+   * Track page view
+   * @param {string} name    Page Name
+   * @param {string} [title] Page title
+   */
+  pageView(name, title) {
```
vitaly-krugl: `pageView` appears to not be used. Is it used implicitly?

unicorn/app/browser/lib/Analytics/GoogleAnalytics.js
```diff
@@ -0,0 +1,295 @@
...
+      .then((records) => {
+        while (records.length > 0) {
+          // Batch at most GA_BATCH_SIZE hits per request
+          let batch = records.splice(0, GA_BATCH_SIZE);
+          let values = batch.map((item) => item.value);
+          let keys = batch.map((item) => item.key);
+          // Post cached hits and delete successfuly uploaded hits from database
+          this._post(values)
+            .then(() => this._deletePending(keys))
+            .catch((error) => this.exception(error));
```
vitaly-krugl: If the user is not connected to the Internet for a very long time, will we then end up with a ton of "network connection error" exceptions in GA when the user subsequently connects to the network?

vitaly-krugl:
@lscheinkman, I'm done with my review. Didn't find concrete bugs, just some concerns ...

unicorn/app/browser/actions/DeleteModel.js
lscheinkman: Unfortunately google analytics is not the right tools for this, however electron supports a [crash-reporter](http://electron.atom.io/docs/api/crash-reporter/) that could be used for this purpose

unicorn/app/browser/entry.js
lscheinkman: Good point, I will remove that.

unicorn/webpack.config.babel.js
lscheinkman: We use [webpack](http://webpack.github.io/docs/what-is-webpack.html) to preprocess and bundle our javascript code into a single `bundle.js` file.

[webpack.EnvironmentPlugin](https://github.com/webpack/docs/wiki/list-of-plugins#environmentplugin) is a builtin [webpack plugin](http://webpack.github.io/docs/list-of-plugins.html) that replaces statements like `process.env.VARIABLE` with the actual value of the environment variable during the build phase.

Currently we use the following environment variables:
- `NODE_ENV` environment variable used by `nodejs` to differentiate between 'production' and 'development' environments.
- `GA_TRACKING_ID` is described in the [README](https://github.com/numenta/numenta-apps/pull/807/files#diff-6b103ee4e67229eca28ef14369cde2ffR433)

unicorn/app/browser/lib/Analytics/GoogleAnalytics.js
lscheinkman: You can't. the only way to know if it really works is be checking Google Analytics Reports

unicorn/app/browser/lib/Analytics/GoogleAnalytics.js
lscheinkman: Good point, I will remove that

unicorn/app/browser/lib/Analytics/GoogleAnalytics.js
lscheinkman: You are right. This API is currently not being used. I used once to track dialog views but decided to track the actions instead. 
I'd like to least it here for API completion even if it is not being used at the moment.

unicorn/app/browser/actions/DeleteModel.js
vitaly-krugl: However, some other places pass the error to the `exception` method: e.g., `this.exception(error)`, `gaTracker.exception('Unknown Error', `Unknown Error: ${message}`)`

lscheinkman: That was wrong. The `exception` tracking event is used to count exceptions. Adding log info to the exception tracking event would break that count, as you pointed out in a different comment. 
I've fixed that.

unicorn/app/browser/actions/HideFileDetails.js
lscheinkman: We are not tracking all actions blindly, each of those actions were intentionally selected, for example, `ACTIONS.RECEIVE_MODEL_DATA` should not be tracked as it is fired on every single model data results and would not add any value to the usage analytics.

unicorn/app/browser/lib/Analytics/GoogleAnalytics.js
```diff
@@ -0,0 +1,293 @@
...
+//
+// http://numenta.org/licenses/
+
+let uuid = require('node-uuid');
+
+// Googe analytics API version
+const API_VERSION = 1;
+
+// Googe analytics endpoint
+const API_ENDPOINT = 'https://www.google-analytics.com';
```
marionleborgne: If you want to send the payload data via a secure connection (over SSL), then you need to send the data to: 
https://ssl.google-analytics.com

lscheinkman: The endpoint documented is https://www.google-analytics.com

See https://developers.google.com/analytics/devguides/collection/protocol/v1/reference#endpoint

unicorn/app/browser/lib/Analytics/GoogleAnalytics.js
```diff
@@ -0,0 +1,295 @@
...
+   * Synchronize local data with server.
+   * Keep pending hits in local database until they are successfuly uploaded
+   */
+  synchronize() {
+    // Empty memory queue
+    let hits = this._queue.splice(0);
+    // Persist hits before uploading to server in case of network error
+    this._savePending(hits)
+        .then(() => this._upload())
+        .catch((error) => this.exception(error));
```
vitaly-krugl: @lscheinkman, there are several calls to `this.exception(error)` in this file.

lscheinkman: this is an old version, you should check the lastest version
 see https://github.com/lscheinkman/numenta-apps/blob/5ab8f01b357f90038e66ed3cc7b189497d20cf23/unicorn/app/browser/lib/Analytics/GoogleAnalytics.js

unicorn/app/browser/lib/Analytics/GoogleAnalytics.js
marionleborgne: I finished testing this. It works after some settings tweak in the Google Analytics dashboard. For posterity note that we have to create a "Application" property, not a "Website" one. When we do this, the usage tracking data can be sent successfully and we also see it in the real-time dashboard 😄 
screen shot 2016-05-26 at 2 26 13 pm

marionleborgne:
@lscheinkman 👍 This is ready for merge once the issue with the `NODE_ENV=production` export issue is fixed. I am only able to get the GA tracking working if I explicitly export this variable and then run `npm run dev`. Building the app does not set this variable as we would expect.

lscheinkman:
@marionleborgne the export issue should be fixed with 29d1fe6
Let me know if it works for you

marionleborgne:
It works 👍 🎉

PR #808, opened 2016-05-25
UNI-481 Freeze nupic at 0.5.0 via unicorn/py/requirements.txt and automate nupic.core commitish

vitaly-krugl:
UNI-481 Freeze nupic at 0.5.0 via unicorn/py/requirements.txt
UNI-481 Modify unicorn/scripts/OSX/build-python.sh to pick up nupic version from unicorn/py/requirements.txt, then get matching nupic.core commitish from the nupic sources of the required version.

### DIFF
--- a/unicorn/scripts/OSX/build-python.sh
+++ b/unicorn/scripts/OSX/build-python.sh
@@ -26,6 +26,7 @@
 # nupic and nupic.bindings installed.
 
 set -o errexit
+set -o pipefail
 set -o xtrace
 
 OSX_VERSION=$(sw_vers -productVersion | awk -F '.' '{print $1 "." $2}')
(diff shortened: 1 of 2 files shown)


vitaly-krugl:
CC @marionleborgne

oxtopus:
LGTM

unicorn/scripts/OSX/build-python.sh
```diff
@@ -81,8 +84,22 @@ pushd $CAPNP
 make -j6 check && make install
 popd
 
-echo "==> Cloning nupic.core ..."
+echo "==> Getting nupic version from ${UNICORN_BACKEND_REQUIREMENTS_TXT}..."
+NUPIC_VER=$(grep "^nupic==" "${UNICORN_BACKEND_REQUIREMENTS_TXT}" | cut -d "=" -f 3)
```
lscheinkman: We should do the same on Windows instead of passing external parameters to be consistent. 
See https://github.com/numenta/numenta-apps/blob/master/unicorn/scripts/Windows64/simple_build_portable_python_with_nupic.ps1#L4

vitaly-krugl:
Thanks for reviewing, @oxtopus

unicorn/scripts/OSX/build-python.sh
vitaly-krugl: It makes sense, @lscheinkman. UNI-481 was specific to OS X, though ("Nupic version not frozen in OSX build script").

Need to make sure that there is a separate issue filed for Windows. Also, I am guessing that we will want to do it as part of the Windows32 build, which will run on both Win32 and Win64 platforms, and dump the Windows64 build altogether.

PR #811, opened 2016-05-26
UNI-494 throws error on files with less than 2 rows and no valid data row

andrewmalta13:
Fails loudly rather than quietly when a user uploads a file with less than 2 rows an neither of them
are valid data rows. Updated tests to reflect this change.

Please Review:
@marionleborgne


unicorn/app/main/FileService.js
```diff
@@ -287,6 +287,15 @@ export class FileService {
         headers = values;
         offset++;
       });
+
+    stream.on('end', () => {
+      // We reached the end of the csv and we read either 0 or 1 rows (each)
+      // of which were not valid data rows.
+      if (offset < 2) {
+        callback('Csv has no valid data rows to read');
```
marionleborgne: Replace message by  `The CSV file does not have any valid data (see supported formats)`  with `supported formats` a hyperlink to `http://numenta.com/htm-studio#get-started`

marionleborgne:
👍 LGTM after you address the comment

andrewmalta13:
updated @marionleborgne

marionleborgne:
👍

PR #812, opened 2016-05-26
UNI-500 Re-rendering: handle the "no model results" case

mrcslws:
Previously, the shouldComponentUpdate method handled charts that had models, but it would always re-render the chart when it didn't have any model results.

Now chart removal is pretty fast, but not instant, because re-rendering the entire file list still takes some time.

Here are some quick numbers. There's a margin of error, but it shows the order of magnitude improvement.

**Before this change**

Add first chart: ~360ms
Remove first chart: ~200ms
Add 5th chart: ~1450ms
Remove 5th chart: ~1300ms

**After this change**

Add first chart: ~360ms
Remove first chart: ~115ms
Add 5th chart: ~400ms
Remove 5th chart: ~200ms

### DIFF
--- a/unicorn/app/browser/components/ModelData.jsx
+++ b/unicorn/app/browser/components/ModelData.jsx
@@ -138,24 +138,25 @@ export default class ModelData extends React.Component {
   } // constructor
 
   shouldComponentUpdate(nextProps, nextState) {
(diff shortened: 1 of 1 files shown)


marionleborgne:
👍 LGTM

PR #813, opened 2016-05-26
Show metric name in hover text, not just 'Value:'

mrcslws:
This change also fixes the following bug:

1. Unhide a model that has finished running
2. Click "show non-aggregated data"

RESULT:
The selected dateWindow jumps to the end

@marionleborgne Note the label behavior. Some example text is:

image

image

image

image

Let me know what you think.


marionleborgne:
👍 LGTM

PR #814, opened 2016-05-26
UNI-496/UNI-503/UNI-511/UNI-523 400 row minimum, NA filtering

andrewmalta13:
-implements 400 row minimum
-updates tests to comply

-the change to the tests may have revealed a bug, see comment here https://jira.numenta.com/browse/UNI-485

Please Review:
@marionleborgne


marionleborgne:
the message is `File does not have at least 400 rows`

But we should say: `The CSV file needs to have at least 400 rows with valid values (see supported formats)`  with `supported formats` a hyperlink to `http://numenta.com/htm-studio#get-started`

marionleborgne:
@andrewmalta13 and i discussed. We agreed on the fact that this PR is also related to UNI-496.

We needs 2 checks: 
1. A check to make sure the row does not have ANY missing values at _file validation_ time, to determine if the format is correct [1] and that we are using the right headers. 
2. A check _for each column_ to determine if a value needs to be skipped [2] so that we only feed valid data to the model

We should store all the data to the DB.

[1] correct format: one and only timestamp, at least 1 numeric value, skip all other columns that don't comply with that.
[2] a value will be skipped if is equal to "NaN", "None", "null", "N/A", "NA" (all of them NOT case sensitive)

andrewmalta13:
going to combine issue UNI-496 and UNI-503 int his PR

andrewmalta13:
still need to implement tests

andrewmalta13:
Don't merge, there is a bug currently where nupic can not generate a model if there were missing metric values in a column.

andrewmalta13:
Fixed bugs and you can now create models successfully from files who have missing values. 
Just tests to implement now. @marionleborgne

unicorn/app/config/na.js
```diff
@@ -0,0 +1,22 @@
...
+// You should have received a copy of the GNU Affero Public License along with
+// this program.  If not, see http://www.gnu.org/licenses.
+//
+// http://numenta.org/licenses/
+
+/**
+ * Setings for NA's in files
+ */
+
+export const NA_STRINGS = ['nan','none', 'null', 'n/a', 'na'];
```
marionleborgne: add new line at EOF

marionleborgne: Can you add `''` (empty string) to this list?

andrewmalta13:
Ready for review, also fixed UNI-511 here

unicorn/app/config/na.js
```diff
@@ -0,0 +1,22 @@
...
+// You should have received a copy of the GNU Affero Public License along with
+// this program.  If not, see http://www.gnu.org/licenses.
+//
+// http://numenta.org/licenses/
+
+/**
+ * Setings for NA's in files
+ */
+
+export const NA_STRINGS = ['nan','none', 'null', 'n/a', 'na', ''];
```
marionleborgne: 👍

unicorn/app/browser/components/FileDetails.jsx
```diff
@@ -259,7 +259,7 @@ export default class FileDetails extends React.Component {
             ref="fileSize"
             value={`${fileSize} KB`}/>
           <TextField
-            floatingLabelText="Number of rows"
+            floatingLabelText="Number of valid rows"
```
marionleborgne: You can keep the original text if you change the logic to count every timestamp

marionleborgne:
@andrewmalta13 and I discussed the following update to the code: 
* we count the number of non "missing values" timestamps
* before checking if data in a column is valid, first check that it is not a missing value

unicorn/py/unicorn_backend/utils/na.py
```diff
@@ -0,0 +1,30 @@
...
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+NA_STRINGS = ['nan','none', 'null', 'n/a', 'na', '']
+
+'''
+  determine if a value is NA.
+  field: string to test 
+'''
+def isNA(field):
```
andrewmalta13: add newline

unicorn/app/main/DatabaseService.js
```diff
@@ -690,15 +700,19 @@ export class DatabaseService {
...
     .on('error', callback)
     .on('end', () => {
+      let metricsDeleted = 0;
       for (let i = 0; i < metrics.length; i++) {
         this.deleteMetric(metrics[i], (error) => {
           if (error) {
             callback(error);
             return;
           }
+          if (++metricsDeleted === metrics.length) {
```
andrewmalta13: This is the fix for UNI-511

marionleborgne:
@andrewmalta13 there is still an issue. See `empty.csv`: [empty.csv.zip](https://github.com/numenta/numenta-apps/files/296695/empty.csv.zip)

screen shot 2016-06-02 at 4 47 01 pm
screen shot 2016-06-02 at 4 48 47 pm

marionleborgne:
Found another "issue" with [just-headers.csv.zip](https://github.com/numenta/numenta-apps/files/296738/just-headers.csv.zip)

You get this:
screen shot 2016-06-02 at 5 32 05 pm

I found this error message a bit confusing. I think we should have the error "The file should have at least 400 rows" before the "The file should have at least one row with no missing values."

What do you think?

marionleborgne:
@andrewmalta13: if the validRows counter is over 400 we throw the error "no rows with missing values". otherwise we throw the not enough rows one instead

marionleborgne:
Will fix UNI-523 as well.

marionleborgne:
Let's have a look at UNI-526 too. It might be an easy fix with this PR

andrewmalta13:
@ marion I agree that 523 should be added since it is already implemented; however I am hesitant to add 526 as I don't this branch to get way too large. Could I perhaps do that separately?

marionleborgne:
@andrewmalta13 absolutely. I meant to say that both of these JIRAs might be automatically fixed once we merge your PR. But seems like it is not the case for UNI-526

marionleborgne:
👍 🎉 🎈 nice!

unicorn/py/unicorn_backend/utils/na.py
```diff
@@ -0,0 +1,30 @@
...
+  return NA_STRINGS.__contains__(field.lower())
```
vitaly-krugl: @marionleborgne @andrewmalta13 

this looks a bit like javascript. The following is more pythonic:

```
return field.lower() in NA_STRINGS
```

PR #815, opened 2016-05-26
UNI-433 Make charts stop disappearing

mrcslws:
@marionleborgne

### DIFF
--- a/unicorn/app/browser/components/ModelList.jsx
+++ b/unicorn/app/browser/components/ModelList.jsx
@@ -19,6 +19,7 @@ import connectToStores from 'fluxible-addons-react/connectToStores';
 import IconCheckbox from 'material-ui/lib/svg-icons/toggle/check-box';
 import Paper from 'material-ui/lib/paper';
 import React from 'react';
+import ReactDOM from 'react-dom';
 
 import Model from './Model';
 import ModelStore from '../stores/ModelStore';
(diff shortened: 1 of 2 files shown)


unicorn/app/browser/components/ModelList.jsx
```diff
@@ -108,9 +109,50 @@ export default class ModelList extends React.Component {
...
+        // display handling assumes the transform won't change.
+        let scale = canvas.height / canvas.offsetHeight;
+        ctx.transform(scale, 0, 0, scale, 0, 0);
+      }
+    }
+  }
+
+  componentDidMount() {
+    this._onVisibilityChangeWrapper = () => this._onVisibilityChange();
+    document.addEventListener('visibilitychange', this._onVisibilityChangeWrapper);
```
marionleborgne: Why do you need to wrap `this._onVisibilityChange()`?

unicorn/app/browser/components/ModelList.jsx
```diff
@@ -108,9 +109,50 @@ export default class ModelList extends React.Component {
...
+  /**
+   * Work around a Chromium bug.
+   *
+   * Chromium "hibernates" each canvas when the page is in the
+   * background. On Mac OS X with a "retina" display, it often fails
+   * to unhibernate, leaving a blank canvas.
+   *
+   * @see https://bugs.chromium.org/p/chromium/issues/detail?id=588434
+   */
+  _onVisibilityChange() {
```
marionleborgne: can you add a comment here `FIXME: https://jira.numenta.com/browse/UNI-484`? thanks

marionleborgne:
👍 LGTM after addressing the 2 small comments. Thanks @mrcslws !

unicorn/app/browser/components/ModelList.jsx
mrcslws: It's serving two purposes:

- It allows us to remove it on unmount. `removeEventListener` takes in a function as an argument, so we need to send in the same function object.
- It's an alternative to using `function.bind`. It makes it so that when the method says `this`, the right thing happens.

I use this same strategy in chart.jsx https://github.com/numenta/numenta-apps/blob/ba09b0b9b329a4a10da756a43811c41170600350/unicorn/app/browser/components/Chart.jsx#L722

marionleborgne: `_onVisibilityChange` is already a function. Can't you just have:

```
 document.addEventListener('visibilitychange', this._onVisibilityChange);
```

and then 
```
document.removeEventListener(this._onVisibilityChange);
```

marionleborgne: To wrap up my question (don't change your code it's fine). Is it the same things as:

````
this._onVisibilityChangeWrapper = this._onVisibilityChange.bind(this);
 document.addEventListener('visibilitychange', );
```

and then 
```
document.removeEventListener(this._onVisibilityChangeWrapper);
```

mrcslws: Correct.

marionleborgne: Thanks. I learned something today :metal:

PR #816, opened 2016-05-26
UNI-502: Use local 'portable_python' if available downloading only when necessary

lscheinkman:
@oxtopus Please review


oxtopus:
:+1:

PR #817, opened 2016-05-26
UNI-461 Display the proper expand/collapse icon upon reload.

vitaly-krugl:
UNI-461 Display the proper expand/collapse icon upon reload.
UNI-461 Make uploaded and sample files expanded upon (re)load of the page (per Marion's request)


vitaly-krugl:
CC @marionleborgne

unicorn/app/browser/components/FileList.jsx
```diff
@@ -27,7 +27,7 @@ import IconButton from 'material-ui/lib/icon-button';
 import IconClose from 'material-ui/lib/svg-icons/hardware/keyboard-arrow-down';
 import IconMenu from 'material-ui/lib/menus/icon-menu';
 import IconMore from 'material-ui/lib/svg-icons/navigation/more-vert';
-import IconOpen from 'material-ui/lib/svg-icons/hardware/keyboard-arrow-up';
+import IconOpen from 'material-ui/lib/svg-icons/hardware/keyboard-arrow-right';
```
marionleborgne: Nice 😄

marionleborgne:
Thanks @vitaly-krugl. I reviewed the changes. 

I think we should keep the behavior where when the use adds a file, it is expanded by default. Even if there is no reload. After your changes, when you add a file, it's collapsed.

vitaly-krugl:
@marionleborgne -- regarding
> After your changes, when you add a file, it's collapsed

Please try again. I made that fix (using rebase) after opening the PR; you may have picked up an earlier version of the changes. On my dev laptop, when I add a new file, it's expanded automatically, and all metric entries are visible.

marionleborgne:
I tried again, and the file is definitely collapsed. Do you run `npm run clean:db:osx` before `npm run dev`? I do that before starting the app and uploading the first file.

marionleborgne:
Also, let me be more explicit about the expected behavior:
* On startup: 
** If there no file of type `uploaded` (i.e. only sample files) then collapse all the sample files except the first one in the list (index 0 = first file in the menu)
** If there is at least one file of type `uploaded` then collapse everything
* When a new file is added:
** Set the state of this file to `expanded`
** The state of the rest of the files stays the same (unless, of course, the user expands/collapses some files)

vitaly-krugl:
@marionleborgne, after `npm run clean:db:osx`, I am getting the behavior that you describe - the newly-open files are collapsed. Prior to `npm run clean:db:osx`, they were getting opened as expanded. I wonder if deleting files via the UI leaves something behind in the database that affects subsequent behavior?

vitaly-krugl:
Regarding {quote} Also, let me be more explicit about the expected behavior{quote}, that is outside the scope of UNI-461. It's like trying to hit a moving target. Please file a separate JIRA for the newly-described behavior. Thx.

marionleborgne:
Ok @vitaly-krugl . I created a sub-task, since this is linked to the parent task. You can find it here: https://jira.numenta.com/browse/UNI-507

unicorn/app/browser/components/FileList.jsx
```diff
@@ -354,11 +352,17 @@ export default class FileList extends React.Component {
         );
         let leftIconButton, toggleIcon;
 
+        // Fix up showNested for newly-added file
+        // TODO it would be nice to clean up showNested for deleted files, too
+        if (typeof this.state.showNested[fileId] === 'undefined') {
```
vitaly-krugl: I would prefer to handle `DELETE_FILE` and `UPLOADED_FILE` dispatches like `FileStore` does, instead of fixing it up in an arbitrary piece of code here. However, I don't know how to do that. Any suggestions?

CC @lscheinkman @mrcslws

vitaly-krugl:
@marionleborgne, this fixes expanded file after a file is uploaded. However, I am not perfectly happy with the solution. See my comment "I would prefer to handle DELETE_FILE and UPLOADED_FILE  ..."

unicorn/app/browser/components/FileList.jsx
marionleborgne: This is a little bit of refactoring, but you could remove `showNested` from the `FileList.jsx` state (i.e. remove `  let showNested = {};`), make `showNested` a property a file in `FileStore` and add this as a property of the file. So a file would still have properties like `file.uid` or `file.name` (etc.) but also a `file.showNested` boolean property. This way you can fire an action UPDATE FILE with the new file state (so in your case, `file.showNested` is now `true` if it was previously `undefined`)

```
import FileUpdateAction from '../actions/FileUpdate';

...

if (typeof this.state.showNested[fileId] === 'undefined') {
   file.showNested = true;
   this.context.executeAction(FileUpdateAction, file);
}
```
See http://fluxible.io/api/actions.html

There are a couple of other spots in this file that you would needs to update so that they use `file.showNested` and not the `this.state.showNested`

vitaly-krugl:
Added commit b49faa20d2a087293d18a77846dd113687f646ca that moves management of expanded/collapsed file view state to FileStore from FileList component.

CC @marionleborgne @lscheinkman

vitaly-krugl:
Ready for review again

unicorn/app/browser/components/FileList.jsx
```diff
@@ -355,20 +340,26 @@ export default class FileList extends React.Component {
...
         }
         leftIconButton = (
           <IconButton style={this._styles.fileToggle}>
               {toggleIcon}
           </IconButton>
         );
 
+        // piggyback on default MaterialUI nested show/hide
+        let ref = this.refs[`file-toggle-${fileId}`];
+        if (typeof ref !== 'undefined') {
```
vitaly-krugl: @brev, IIRC, you showed some JS code in one of your presentations that would elegantly skip execution of code if the variable was undefined (instead of the explicit `if (typeof ref !== 'undefined')` check). How would you do that here? Thx.

unicorn/app/config/default.json
```diff
@@ -116,4 +116,4 @@
   "samples": {
     "path": "../samples"
   }
-}
+}
```
vitaly-krugl: I noticed tat when I run unicorn (`npm run dev`), the file unicorn/app/config/default.json gets overwritten, causing it to lose its trailing newline. It doesn’t seem right that the application should be overwriting a file  that is under source control.

unicorn/app/browser/components/FileList.jsx
marionleborgne: @vitaly-krugl: `if (ref) {...}` should work here, because you know `ref` was declared (L355). See: http://stackoverflow.com/a/5515349

unicorn/app/browser/stores/FileStore.js
```diff
@@ -43,13 +43,15 @@ export default class FileStore extends BaseStore {
...
    */
   static get handlers() {
     return {
       DELETE_FILE: '_handleDeleteFile',
       UPDATE_FILE: '_handleSetFile',
-      UPLOADED_FILE: '_handleSetFile',
-      LIST_FILES: '_handleListFiles'
+      UPLOADED_FILE: '_handleUploadedFile',
+      LIST_FILES: '_handleListFiles',
+      SET_FILE_EXPANDED_STATE: '_handleSetFileExpanedState'
```
marionleborgne: typo in `_handleSetFileExpanedState ` (should be `_handleSetFileExpandedState `)

unicorn/app/browser/stores/FileStore.js
```diff
@@ -91,4 +93,24 @@ export default class FileStore extends BaseStore {
...
+
+  /**
+   * Set expanded/collapsed view state of file.
+   *
+   * @param {object} payload - Action payload
+   * @param {String} payload.filename - File Name
+   * @param {Boolean} payload.expanded - true for expanded file view; false for
+   *   collapsed
+   */
+  _handleSetFileExpanedState(payload) {
```
marionleborgne: same comment as above about the typo

unicorn/app/browser/components/FileList.jsx
vitaly-krugl: Thanks @marionleborgne, I am going to move this back to `_handleFileToggle` (where I won't need to check for undefined) due to 

> Warning: setState(...): Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.

vitaly-krugl:
@marionleborgne: I resolved your feedback via the latest commit into this PR.

marionleborgne:
Thanks @vitaly-krugl. The last feedback is that there is a failing test due to the additional attribute added to  `file` in the `FileStore`:

```
  1) FileStore #_handleListFiles:

      AssertionError: [ { expanded: true,
    name: 'file1.csv',
    filename: 'fixtures/file1.csv',
    type: 'sample' },
  { expanded: true,
    nam deepEqual [ { name: 'file1.csv',
    filename: 'fixtures/file1.csv',
    type: 'sample' },
  { name: 'file2.csv',
    filename: 'fixtures/
      + expected - actual

       [
         {
      -    "expanded": true
           "filename": "fixtures/file1.csv"
           "name": "file1.csv"
           "type": "sample"
         }
         {
      -    "expanded": true
           "filename": "fixtures/file2.csv"
           "name": "file2.csv"
           "type": "sample"
         }
```

it's just about setting the default value `expanded: true` in the expected values. So in `FileStore_test.js`

```
const EXPECTED_SINGLE_FILE = [{
  name: 'file.csv',
  filename: '../fixtures/file.csv',
  type: 'uploaded',
  expanded: true

}];

const EXPECTED_MULTIPLE_FILES = [{
  name: 'file1.csv',
  filename: 'fixtures/file1.csv',
  type: 'sample',
  expanded: true
},{
  name: 'file2.csv',
  filename: 'fixtures/file2.csv',
  type: 'sample',
  expanded: true
}];
```


You can test it by running `npm run test` . There is still one failing test in `ParamFinderService2` but you can ignore it, it's a known issue that we'll fix.

marionleborgne:
👍 LGTM after addressing the comment about the test

vitaly-krugl:
@marionleborgne: sorry about the test; fixed and squashed into single commit.

marionleborgne:
Thanks @vitaly-krugl 👍

PR #818, opened 2016-05-27
Windows ia32 support

oxtopus:
This adds support for windows.  It's still partially 64-bit due to lack of 32-bit nupic dependencies at 0.5.0, but fully 32-bit envs can be supported in the future with updates to user newer nupic version.


vitaly-krugl:
@oxtopus , @lscheinkman: does this mean that we absolutely cannot have a single Win32-only build that runs universally on Win32 and Win64 versions of Windows? I was hoping that a single Win32 build would suffice for all windows types targeted by Unicorn.

oxtopus:
This is a single 32-bit build that should work universally in 32-bit and 64-bit windows environments.

vitaly-krugl:
Thanks for the clarification, @oxtopus! I was looking at PR #816 earlier, and 32 + 64-bit entries (along with `// FIXME: Create ia32 build scripts`) lead me to believe that we're pursuing two builds types for Windows.

lscheinkman:
@oxtopus 
When we upgrade nodejs version in **HTM Studio** do we also need a new locked down version of **leveldown**?
Is there a way to use the current nodejs version here?
https://github.com/numenta/leveldown/blob/bf6dc51b812ad4668dc1bd1df8fd3958d0373866/package.json#L52-L55

lscheinkman:
The current nodejs version is available from `$npm_config_node_version` variable.

oxtopus:
Yes, I believe so.

PR #819, opened 2016-05-27
UNI-449 Don't show scrambled or duplicated model results

mrcslws:
Before this change, model results were sent to two places: the client and the database. Both of them simultaneously built up their own copy of the model results.

Now the model results are sent to the database, and content-free change notifications are sent to the client. The client chooses when to pull in the latest results. Now there's one source of truth.

Before this change, when you hid a model in the UI, we continued to accrue model results for it. After you unhid the model, the database was queried, and the accrued results were appended. No reconciliation happened, so it caused duplication of displayed data, and it caused the chart's line to strike through the data at the append point.

This commit makes a breaking change to the database schema. We no longer use timestamps as keys. We use row numbers. This allows the client to specify row ranges without needing to know the timestamp
string, and it allows the database to contain multiple rows with the same timestamp, which can happen in input data.

The integration test changes are mostly due to changed DatabaseService method signatures.


marionleborgne:
@mrcslws This is introducing a bug:
1. select a metric (for example `machine_data.csv`) and click on analyze
2. Let it run for a couple of seconds and hit comand+Q to quit.
3. You get an exception:
screen shot 2016-05-31 at 10 58 46 am

mrcslws:
@marionleborgne thanks, nice catch. I pushed a fix.

marionleborgne:
👍

PR #821, opened 2016-05-31
UNI-485: Make sure application is initialized before starting

lscheinkman:
@marionleborgne I could not reproduce but added a check just in case there is a race condition.
Also removed auto updater dialog messages such as "Update check failed: (null)" that may occur on network or other errors.

### DIFF
--- a/unicorn/app/main/AutoUpdate.js
+++ b/unicorn/app/main/AutoUpdate.js
@@ -53,8 +53,9 @@ export default class AppUpdater {
     autoUpdater.setFeedURL(feedUrl);
 
     autoUpdater.addListener('error', (event, error) => {
-      let title = config.get('update:errorTitle');
-      dialog.showErrorBox(title, error);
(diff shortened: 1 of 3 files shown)


unicorn/app/main/AutoUpdate.js
```diff
@@ -53,8 +53,9 @@ export default class AppUpdater {
     autoUpdater.setFeedURL(feedUrl);
 
     autoUpdater.addListener('error', (event, error) => {
-      let title = config.get('update:errorTitle');
-      dialog.showErrorBox(title, error);
+      // Ingore auto update errors
+      // See https://github.com/electron/electron/issues/4699
+      console.warn('AppUpdater:error', error);
```
vitaly-krugl: Whenever I use console.info or console.log, ESLint complains: "unexpected console statement"

marionleborgne:
👍

PR #822, opened 2016-05-31
UNI-507 Implement expanded/collapsed file view state rules at startup.

vitaly-krugl:


marionleborgne:
👍 thanks @vitaly-krugl

PR #823, opened 2016-06-01
UNI-502: Add script to remove cached `portable_python`

lscheinkman:
@marionleborgne per your request.
To remove your cached version of `portable_python` run the following command:
```shel
npm run clean:portable_python:osx
```
This will force the new version of python to be downloaded from the latest successful build like before.


marionleborgne:
👍

PR #824, opened 2016-06-01
UNI-518: Fix error messages when creating models

lscheinkman:
@marionleborgne This PR fixes messages being logged as  'UNKNOWN_MODEL_ERROR' and 'UNKOWN_PARAM_FINDER_ERROR' when creating the models. Now it logs the real error name.
We should fix the real error on a separate PR.


marionleborgne:
@lscheinkman 
1. is there a JIRA to track the fact that we need to fix the real error? 
2. does the code need a FIXME comment with this JIRA ID?

lscheinkman:
There is no JIRA or FIXME because because now we will be able to see the actual error as we use the application. Until now all I saw was unknown errors.

PR #825, opened 2016-06-01
UNI-501 Identify first csv row with numbers and no date as header row.

vitaly-krugl:


vitaly-krugl:
@marionleborgne: Ready for review

unicorn/tests/js/fixtures/numeric-only-header.csv
```diff
@@ -0,0 +1,401 @@
+0,1,2,3,4
```
marionleborgne: what happens if `0` is a unix timestamp? there is now way to disambiguate a data row from a header row.

vitaly-krugl: @marionleborgne, this PR doesn't support unix timestamps. Unix timestamps will be supported in a later PR that will address UNI-435.

vitaly-krugl: @marionleborgne, also in "there is now way to disambiguate a data row from a header row", did you mean to say "now" or "no"?

marionleborgne: 👍

PR #827, opened 2016-06-01
UNI-495 Enable YYYY-MM-DD and YYYY/MM/DD date formats.

vitaly-krugl:
This was already being tested in unicorn_backed, but none of it is tested on the JS side.

Marion recommended to defer JS-side tests, since this is so well tested in unicorn_backend, in favor of higher-priority issues.


vitaly-krugl:
@marionleborgne: ready for review

marionleborgne:
👍 thanks

PR #828, opened 2016-06-02
Bugfix: We're cancelling model data queries too often

mrcslws:
The stop model action gets called whenever a model ends, so it's the wrong place to cancel any pending queries.

With super-small datasets, we were seeing 0 results appear.

Thanks @marionleborgne for catching this.


marionleborgne:
👍 thanks @mrcslws

PR #829, opened 2016-06-02
UNI-490: Restore development tools in development mode

lscheinkman:
Restore JS Development tools menu option when launching the app in development mode, either via `npm run dev` or when the `NODE_ENV` variable is set to 'development'

CC: @marionleborgne, @andrewmalta13


andrewmalta13:
LGTM

PR #830, opened 2016-06-02
UNI-387 charts are ordered correctly

andrewmalta13:
Implements desired ordering of  charts.

The the behaviors stemmed from the use of Map() objects whose iterables always iterate in the order
that the objects were added. (Therefore hiding and showing did not work as intended with the charts)

Models were a quick fix just using a naive stack

Please review:
@lscheinkman


andrewmalta13:
@lscheinkman and I have determined that we have no method as of now (without changing the schema of the database) to have the files sorted by time of adding. I am removing my attempt at this from this PR and this only addresses the chart issue of the JIRA

unicorn/app/browser/stores/FileStore.js
```diff
@@ -61,7 +61,7 @@ export default class FileStore extends BaseStore {
   }
 
   getFiles() {
-    return Array.from(this._files.values());
+    return Array.from(this._files.values()).reverse();
```
lscheinkman: The database stores files by `uid. When loading the files from the database this list will be sorted by `uid` descending

lscheinkman:
@marionleborgne It is easier to break this JIRA into 2 different PRs:
- [x] **Charts**: Most recently selected chart at the top (👍)
- [ ] **Files**: most recently added file at the top.  Requires database schema changes, separate PR.

Let me know if you agree or would like to break the JIRA into 2 separate sub-tasks

marionleborgne:
@lscheinkman  @andrewmalta13 @cbaranski I broke the part of this JIRA related to the files into a different JIRA  I took the left menu sorting out of this JIRA as we discussed and created https://jira.numenta.com/browse/UNI-537

marionleborgne:
👍 thanks @andrewmalta13

PR #835, opened 2016-06-03
UNI-529 adds excel default date/time formats

andrewmalta13:
adds excel's default date/time formats

@cbaranski @marionleborgne


unicorn/app/config/momentjs_to_datetime_strptime.json
```diff
@@ -107,5 +107,13 @@
...
       "MM/DD/YY": "%m/%d/%y"
     }
+  },
+  {
+    "cateogry": "US Date, time (excel default)",
+    "mappings": {
+      "MM/DD/YY H:mm": "%m/%d/%Y %H:%M",
+      "MM/D/YY H:mm": "%m/%d/%Y %H:%M",
+      "M/DD/YY H:mm": "%m/%d/%Y %H:%M"
+    }
```
vitaly-krugl: Does the test https://github.com/numenta/numenta-apps/blob/master/unicorn/tests/py/unit/unicorn_backend/date_time_utils_test.py cover all three cases on the left?

unicorn/app/config/momentjs_to_datetime_strptime.json
```diff
@@ -107,5 +107,13 @@
       "MM/DD/YYYY": "%m/%d/%Y",
       "MM/DD/YY": "%m/%d/%y"
     }
+  },
+  {
+    "cateogry": "US Date, time (excel default)",
+    "mappings": {
+      "MM/DD/YY H:mm": "%m/%d/%Y %H:%M",
+      "MM/D/YY H:mm": "%m/%d/%Y %H:%M",
+      "M/DD/YY H:mm": "%m/%d/%Y %H:%M"
```
vitaly-krugl: What's the difference between "MM" and "M"; also between "DD" and "D"? Do we really need all those permutations?

andrewmalta13: They are necessary because excel can export a date like:
1/2/12 2:01 and also 10/2/12 2:01
and similarly
1/2/12 2:01 and also 1/22/12 2:01

andrewmalta13:
@vitaly-krugl and I found that moment.utc(timestamp, format, true).isValid() is true if and only if the timestamp matches exactly with the format. For example,

moment.utc("12/2/13 21:15", "MM/DD/YY H:mm", true).isValid()   => false
moment.utc("12/2/13 21:15", "MM/D/YY H:mm", true).isValid()     => true

this is the code we are using on the javascript side to check the format of the timestamps, so this is good
to understand completely.

unicorn/app/config/momentjs_to_datetime_strptime.json
```diff
@@ -107,5 +107,18 @@
...
+    "category": "US Date, time (excel default)",
+    "mappings": {
+      "MM/DD/YY H:mm": "%m/%d/%y %H:%M",
+      "MM/DD/YY HH:mm": "%m/%d/%y %H:%M",
+      "MM/D/YY H:mm": "%m/%d/%y %H:%M",
+      "MM/D/YY HH:mm": "%m/%d/%y %H:%M",
+      "M/DD/YY H:mm": "%m/%d/%y %H:%M",
+      "M/DD/YY HH:mm": "%m/%d/%y %H:%M",
+      "M/D/YY  H:mm": "%m/%d/%y %H:%M",
+      "M/D/YY  HH:mm": "%m/%d/%y %H:%M"
```
marionleborgne: What if you have seconds (HH:MM:SS or H:MM:SS) or just the hour?(H or HH) ? with the the same permutations for month, day, year.

marionleborgne:
@andrewmalta13 👍 LGTM. I just have one question about additional formats (not sure if they are default excel formats though)

marionleborgne:
👍

PR #836, opened 2016-06-06
UNI-527: Copy file to application cache

lscheinkman:
@marionleborgne  Run `npm run clean:db:osx` before testing. This PR requires a fresh database because it changes all the File/Metric IDs.


unicorn/tests/js/integration/services/DatabaseService_test.js
```diff
@@ -283,7 +284,7 @@ describe('DatabaseService:', () => {
   let service;
 
   before(() => {
-    service = new DatabaseService(TEMP_DIR);
+    service = new DatabaseService(TEMP_DB_DIR, TEMP_CACHE_DIR);
```
marionleborgne: Do you need to delete these at some point? in `after()` or `afterEach()`

marionleborgne:
@lscheinkman 👍 LGTM after resolving the merge conflicts (and I also have 1 question). Here is a gist with my resolution: https://gist.github.com/marionleborgne/820b7302b87f2e2a8f84f85bddc46500

PR #837, opened 2016-06-07
UNI-528: Don't call stop action after process closes. Use close action instead

lscheinkman:
@marionleborgne @andrewmalta13 Please review


andrewmalta13:
LGTM, works as intended now. 👍

PR #838, opened 2016-06-07
UNI-456 fixes default export model name

andrewmalta13:
implements 

"The spec says that the export reultss filename should be pre-populated with: htm_results_<fileName>_<metricName>.csv"

Though I made a guess that we didn't want a random .csv in the filename so I stripped that.

Please Review:
@lscheinkman @marionleborgne


lscheinkman:
👍 LGTM

marionleborgne:
👍

PR #840, opened 2016-06-08
UNI-538 better warning message upon quitting the app when a model is running

marionleborgne:


unicorn/app/browser/entry.js
```diff
@@ -153,13 +153,20 @@ document.addEventListener('DOMContentLoaded', () => {
...
-                  running. All models will be interrupted upon quitting, and
-                  it won’t be possible to restart these models. All results
-                  obtained so far will be persisted. Are you sure you want to
-                  quit the app and stop all running models?`,
+        message: trims`Are you sure you want to quit the app and stop all
+                  running models?`,
+        detail: trims`There ${isOrAre} still ${modelCount} active
+                  model${trailingS} running. All models will be interrupted upon
+                  quitting, and it won’t be possible to restart these models.
+                  All results obtained so far will be persisted.`,
```
lscheinkman: @marionleborgne Ideally, for future localization purposes, you would have two different strings defined in [unicorn/app/config/default.json](https://github.com/numenta/numenta-apps/blob/master/unicorn/app/config/default.json) instead of formatting plurals in place.
Something like this:

```
let detail;
if (modelCount === 1) {
   detail = config.get('dialog:close_application:singular'),
} else {
   detail = config.get('dialog:close_application:plural'),
}
```

And `unicorn/app/config/default.json` would have the strings:
```
...
"dialog": {
   "close_application": {
        "singular": "There is still one active model running. The active model will be interrupted upon quitting, and it won’t be possible to restart the model. All results obtained so far will be persisted",
        "plural": "There are still ${modelCount} active models running. All models will be interrupted upon quitting, and it won’t be possible to restart these models. All results obtained so far will be persisted"
  }
}
...
```

marionleborgne:
@lscheinkman addressed your comment and made the update to use the translation file.

PR #842, opened 2016-06-08
UNI-540 Better log info and checks

marionleborgne:
* suppress misleading error message that happens even if uninstall was sucessfull
* make sure that unicorn backend was successfully uninstalled & installed

### DIFF
--- a/unicorn/app/package.json
+++ b/unicorn/app/package.json
@@ -43,8 +43,8 @@
   ],
   "scripts": {
     "postinstall": "npm run install:backend",
-    "install:backend": "npm run clean:backend && python ../py/setup.py install",
-    "clean:backend": "python -m pip uninstall unicorn_backend -y || true"
+    "install:backend": "npm run clean:backend && python ../py/setup.py install && python -c 'import unicorn_backend'",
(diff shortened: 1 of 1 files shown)


unicorn/app/package.json
```diff
@@ -43,8 +43,8 @@
   ],
   "scripts": {
     "postinstall": "npm run install:backend",
-    "install:backend": "npm run clean:backend && python ../py/setup.py install",
-    "clean:backend": "python -m pip uninstall unicorn_backend -y || true"
+    "install:backend": "npm run clean:backend && python ../py/setup.py install && python -c 'import unicorn_backend'",
+    "clean:backend": "python -m pip uninstall unicorn_backend -y 2> /dev/null || true && if python -c 'import unicorn_backend'; then exit 1; else exit 0; fi"
```
lscheinkman: @marionleborgne `/dev/null` will not work on windows

oxtopus: How about:

```
python -m pip uninstall unicorn_backend -yqqq
```

Without the stream redirection.

oxtopus:
@marionleborgne can you check in on this PR?  It's old, and no longer in a mergeable state.

marionleborgne:
@oxtopus I would close this PR. The `clean:backend` works and is cross platform https://github.com/numenta/numenta-apps/blob/master/unicorn/package.json#L27

PR #843, opened 2016-06-08
TAUR-1932 Remove TWC stock symbol from taurus configuration (due to acquisition by Charter Communications).

vitaly-krugl:


vitaly-krugl:
@oxtopus: Please review

PR #847, opened 2016-06-09
Fix Auto Update instructions in README

lscheinkman:


lscheinkman:
@marionleborgne 
👍  All files working as design 👏

PR #849, opened 2016-06-10
UNI-545 fixes exception and date_time_utils.py bug

andrewmalta13:
please review @marionleborgne 

Note the file in the UNI-545 JIRA will still fail param finder and get UNI-528 like symptoms however, the error message it now gets is in line with @lscheinkman  sees on the other files that exhibit the UNI-528 bug which is:

{"errorText": "negative dimensions are not allowed", "diagnosticInfo": "Traceback (most recent call last):\n  File \"build/bdist.macosx-10.5-x86_64/egg/unicorn_backend/param_finder_runner.py\", line 201, in main\n  File \"build/bdist.macosx-10.5-x86_64/egg/unicorn_backend/param_finder.py\", line 241, in findParameters\n    medianSamplingInterval)\n  File \"build/bdist.macosx-10.5-x86_64/egg/unicorn_backend/param_finder.py\", line 363, in _resampleData\n    newTimeStamps = numpy.empty(nSampleNew, dtype=\"datetime64[s]\")\nValueError: negative dimensions are not allowed\n"}


unicorn/app/main/FileService.js
```diff
@@ -69,6 +69,9 @@ const TIMESTAMP_HEADER_SUBSTRINGS = ['time', 'date'];
  *                    , and false otherwise.
  */
 function isNA(entry) {
+  if (typeof entry === 'undefined') {
```
marionleborgne: @andrewmalta13 why do you need this check?

andrewmalta13: I didn't think I had to. Celeste encountered this: https://jira.numenta.com/secure/attachment/18288/Uncaught%20exception.png

unicorn/app/main/FileService.js
```diff
@@ -69,6 +70,10 @@ const TIMESTAMP_HEADER_SUBSTRINGS = ['time', 'date'];
...
+    log.error(new Error('Param "entry" of FileService.isNA() is not defined.'));
+    return false;  // somewhat arbitrary decision to return false in this case
```
vitaly-krugl: If we think that something is wrong with the value of `entry`, shouldn't it raise an exception, instead?

unicorn/py/unicorn_backend/utils/date_time_utils.py
```diff
@@ -84,6 +87,11 @@ def parseDatetime(dateString, dateFormat):
       raise ValueError(
         "Expected non-negative Unix Timestamp, but got {}".format(dateString))
 
+    # check to see if the timestamp float is beyond the range of UNIX SECONDS
+    # if it is interpret is as a ms timestamp
+    if timestampFloat >= _MAX_UNIX_SECONDS:
+      timestampFloat /= 1000
```
vitaly-krugl: So, if the previous value was `< _MAX_UNIX_SECONDS`, it would have been interpreted as seconds, but the next value would be interpreted as milliseconds?

unicorn/py/unicorn_backend/utils/date_time_utils.py
```diff
@@ -84,6 +87,11 @@ def parseDatetime(dateString, dateFormat):
       raise ValueError(
         "Expected non-negative Unix Timestamp, but got {}".format(dateString))
 
+    # check to see if the timestamp float is beyond the range of UNIX SECONDS
+    # if it is interpret is as a ms timestamp
+    if timestampFloat >= _MAX_UNIX_SECONDS:
```
vitaly-krugl: If `_MAX_UNIX_SECONDS` is "the maximum value datetime parses", shouldn't this be `timestampFloat > _MAX_UNIX_SECONDS`?

andrewmalta13: Oops I should have made it more clear the constant is the largest value
that can be parsed by the function.

vitaly-krugl: Well, if it's the largest seconds value that may be parsed (i.e., parsing it as seconds would be successful), then the comparison above doesn't make sense, and should be `timestampFloat > _MAX_UNIX_SECONDS`, instead.

andrewmalta13: Ok I will look into it tommorow

andrewmalta13: Ok I had forgot:

datetime.utcfromtimestamp(253402300800.0) => ValueError: year is out of range
datetime.utcfromtimestamp(253402300799.0) =>  datetime.datetime(9999, 12, 31, 23, 59, 59)

I could update the name of the constant if you think that would make it more clear, but the constant really is the UNIX timestamp of the year 10000 in seconds which can not be parsed by datetime.

unicorn/app/main/FileService.js
andrewmalta13: @marionleborgne told me to do this, but I tend to agree (or consider undefined a NA value).

unicorn/py/unicorn_backend/utils/date_time_utils.py
vitaly-krugl: @andrewmalta13, when you use the word "MAX" in this constant's name, it makes it sound like that is the max legal value (another example might be MAX_INT). So, you can keep the name, but change the value to `253402300799.0` and the logic to `timestampFloat > _MAX_UNIX_SECONDS`

unicorn/py/unicorn_backend/utils/date_time_utils.py
vitaly-krugl: @andrewmalta13, when you use the word "MAX" in this constant's name, it makes it sound like that is the max legal value (another example might be MAX_INT). So, you can keep the name, but change the value to `253402300799.0` and the logic to `timestampFloat > _MAX_UNIX_SECONDS`

unicorn/py/unicorn_backend/utils/date_time_utils.py
```diff
@@ -84,6 +87,11 @@ def parseDatetime(dateString, dateFormat):
       raise ValueError(
         "Expected non-negative Unix Timestamp, but got {}".format(dateString))
 
+    # check to see if the timestamp float is beyond the range of UNIX SECONDS
+    # if it is interpret is as a ms timestamp
+    if timestampFloat > _MAX_UNIX_SECONDS:
+      timestampFloat /= 1000
```
vitaly-krugl: If I have rows with the following timestamps:

253402300799.0 - 600
253402300799.0 - 300
253402300799.0
253402300799.0 + 300
253402300799.0 + 600

then the first three will be interpreted as seconds, and the last two will be interpreted as milliseconds, which is not what the user intended. 

CC @andrewmalta13 @marionleborgne

andrewmalta13: Keep this updated with what is the desired behavior. (what we wish to support)
I'll address it when I have the required details.

vitaly-krugl: Please work it out with @marionleborgne

marionleborgne: @andrewmalta13 @vitaly-krugl I commented on the JIRA.

The convention that we are taking is that we only support unix timestamps in s  that are above `253402300.800` ( `1978-01-11T21:31:40+00:00`) and below `253402300800`  (`10000-01-01T00:00:00+00:00`) 

This should be reflected in the code and in the comments so that other people taking over understand that it's a convention that we are taking.  


Also @andrewmalta13 use 253402300800.0 instead of 253402300799.0 for _MAX_UNIX_SECONDS

unicorn/app/main/FileService.js
marionleborgne: @andrewmalta13 @vitaly-krugl if isNA() is passed an undefined value, we should return False (because it might actually not be a missing value) and we should log the error so that we can further investigate. This bug happened one time to @cbaranski but we haven't been able to see it again. So at this point let's just capture the log of the error.

marionleborgne:
@andrewmalta13 done with review. let me know when you're done addressing the comments. I'll merge this (it's going to be useful for #853 )

unicorn/app/main/FileService.js
vitaly-krugl: @marionleborgne @andrewmalta13: in this case, I advocate adding a code comment that justifies this choice, since it's a non-standard pattern for python. Otherwise, the next maintainer/contributor/user will be scratching her head.

unicorn/py/unicorn_backend/utils/date_time_utils.py
vitaly-krugl: @marionleborgne , regarding {quote} Also @andrewmalta13 use 253402300800.0 instead of 253402300799.0 for _MAX_UNIX_SECONDS{quote}

This is incorrect advice, because the name `_MAX_UNIX_SECONDS` implies that this would be a maximum valid value to most readers. However, 253402300800.0 is not valid, and that forces the code to use a strange comparison (from readability point of view) `timestampFloat >= _MAX_UNIX_SECONDS` instead of the more meaningful `timestampFloat > _MAX_UNIX_SECONDS`. @andrewmalta13 and I already had this discussion, and he changed it to 253402300799.0 based on that reasoning. @marionleborgne, please discuss with me if you disagree. Thx.

CC @andrewmalta13

vitaly-krugl:
@marionleborgne , regarding {quote} Also @andrewmalta13 use 253402300800.0 instead of 253402300799.0 for _MAX_UNIX_SECONDS{quote}

This is incorrect advice, because the name `_MAX_UNIX_SECONDS` implies that this would be a maximum valid value to most readers. However, 253402300800.0 is not valid, and that forces the code to use a strange comparison (from readability point of view) `timestampFloat >= _MAX_UNIX_SECONDS` instead of the more meaningful `timestampFloat > _MAX_UNIX_SECONDS`. @andrewmalta13 and I already had this discussion, and he changed it to 253402300799.0 based on that reasoning. @marionleborgne, please discuss with me if you disagree. Thx.

CC @andrewmalta13

marionleborgne:
Nice catch in https://github.com/numenta/numenta-apps/pull/849#issuecomment-226330217 @vitaly-krugl. Yes I agree with that. cc @andrewmalta13

unicorn/py/unicorn_backend/utils/date_time_utils.py
```diff
@@ -84,6 +87,14 @@ def parseDatetime(dateString, dateFormat):
       raise ValueError(
         "Expected non-negative Unix Timestamp, but got {}".format(dateString))
 
+    # check to see if the timestamp float is beyond the range of UNIX SECONDS
+    # if it is, interpret is as a ms timestamp. To be clear we are taking the 
+    # convention that we only support unix timestamps in seconds that are above
```
marionleborgne: @andrewmalta13 The code is correct but i need to update the comment:

Check to see if the timestamp float is beyond the range of _MAX_UNIX_SECONDS. If it is, interpret is as a ms timestamp. To be clear we are taking the convention that we only support unix timestamps in seconds that are between 0 and strictly below 253402300800.0 (10000-01-01T00:00:00+00:00). And we support unix timestamps in milliseconds that are at or above 253402300800 (unix timestamp in ms) which is 1978-01-11T21:31:40+00:00. 

@vitaly-krugl proof 😄

marionleborgne:
Aha 💡 I was able to reproduce the case leading to an undefined entry. This seems to happen when there is an empty row. So actually making isNA() return `true` is the right way to go. I made this change in #856

marionleborgne:
👍 thanks @andrewmalta13 and @vitaly-krugl

andrewmalta13:
👍 thanks @vitaly-krugl for the help

PR #856, opened 2016-06-15
Combine UNI-528 and UNI-545

marionleborgne:
@andrewmalta13 FYI This two bugs were related so I am merging them here. I had to follow up with some additional fixups for model_runner and param_finder on the python side (prevent them to crash with empty rows)


marionleborgne:
@lscheinkman This is the PR to test. Here are all the files to make it simpler:

[UNI-528-UNI-545.zip](https://github.com/numenta/numenta-apps/files/317382/UNI-528-UNI-545.zip)

lscheinkman:
👍 Works as expected 👏

PR #858, opened 2016-06-16
UNI-496/503/511 changes a line to be more pythonic

andrewmalta13:
addresses a comment from @vitaly-krugl.

Same logic, just more pythonic.


vitaly-krugl:
LGTM

CC @marionleborgne

marionleborgne:
👍

PR #860, opened 2016-06-17
UNI-545 Unicorn - fix interface and unexpected unix timestamp parsing results

vitaly-krugl:
Also appears to fix a rendering bug when the dataset contains unix millisecond timestamps (see @vitaly-krugl's subsequent comment with the screenshot).

Unicorn - follow-up to fix interface and unexpected unix timestamp parsing results introduced by PR #849 (UNI-545), by determining the format in FileService.js as we do for all other formats, and then passing the corresponding date format to date_time_utils.parseDatetime without second-guessing. The new millisecond backend format is #t (versus #T for seconds).

Unicorn Backend's function `date_time_utils.parseDatetime` must not second-guess the format passed in the dateFormat arg. Besides violating the interface, the second-guessing was causing unexpected results when the prior timestamp was <= 253402300799 and the following timestamp in the sequence was > 253402300799.


vitaly-krugl:
CC @marionleborgne @andrewmalta13

vitaly-krugl:
Also appears to fix this issue when processing a file with unix timestamps expressed in milliseconds:


screen shot 2016-06-17 at 11 03 14 am

The data file corresponding to this mangled view was https://github.com/vitaly-krugl/numenta-apps/blob/9aba4bba7f4f9b9abfef605cf91dab8e57193f4a/unicorn/tests/js/fixtures/unix-millisecond-timestamp.csv

marionleborgne:
@vitaly-krugl and I reviewed this PR and it does fix the scrambled data issue. We have the issue in the current master, but his changes fix the bug. Nice catch! 👍 

LGTM after removing the unused .csv file in `fixtures`.

Thanks @vitaly-krugl

marionleborgne:
👍

PR #864, opened 2016-06-19
UNI-542: Fix call to 'Update.exe' with quoted parameters 

lscheinkman:
@marionleborgne, @cbaranski 
There should be a desktop and start menu shortcuts once the app is done installing.

### DIFF
--- a/unicorn/app/main/AutoUpdate.js
+++ b/unicorn/app/main/AutoUpdate.js
@@ -21,7 +21,7 @@ import {
 import config from './ConfigService';
 import os from 'os';
 import path from 'path';
-import spawn from 'child_process';
+import childProcess from 'child_process';
 import log from './Logger'
 
 // Squirrel Windows update process name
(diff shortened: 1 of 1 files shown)


marionleborgne:
👍 nice

unicorn/app/main/AutoUpdate.js
```diff
@@ -34,7 +34,7 @@ const UPDATE_EXE =
  */
 function _updateWin32(args) {
   try {
-    spawn(UPDATE_EXE, args, {detached: true})
+    childProcess.spawn(UPDATE_EXE, args, {detached: true})
```
marionleborgne: @lscheinkman why this modification?

marionleborgne:
@lscheinkman It works for me on Windows 8

unicorn/app/main/AutoUpdate.js
lscheinkman: because the `import` at the begging of the file was importing the `child_process` module not the `spawn` function see https://nodejs.org/api/child_process.html

marionleborgne:
👍

PR #866, opened 2016-06-20
UNI-558: Make app single instance

lscheinkman:
@marionleborgne this should make the app single instance on Windows and Mac preventing database lock errors when trying to relaunch the app.

CC: @cbaranski

### DIFF
--- a/unicorn/app/main/index.js
+++ b/unicorn/app/main/index.js
@@ -136,6 +136,20 @@ app.on('window-all-closed', () => {
   app.quit();
 });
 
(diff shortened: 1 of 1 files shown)


marionleborgne:
👍 it works in dev mode

marionleborgne:
@lscheinkman When I build the app though, I get the following error if I try to run 2 HTM Studio at the same time:

screen shot 2016-06-20 at 1 32 07 pm

marionleborgne:
👍 works after packaging the app as well. The app cannot be started 2 times 💯

PR #868, opened 2016-06-22
UNI-575 ignore env variables like PYTHONPATH

marionleborgne:
Thanks @mrcslws and @lscheinkman


lscheinkman:
👍  LGTM. Reproduced the original behavior and validated the fix works
Thank you @mrcslws

PR #869, opened 2016-06-22
UNI-562 updated delete file confirmation message

marionleborgne:
New text updated to:

screen shot 2016-06-22 at 3 12 06 pm

cc @lscheinkman @cbaranski

### DIFF
--- a/unicorn/app/config/default.json
+++ b/unicorn/app/config/default.json
@@ -48,8 +48,8 @@
         "title": "Add new data file"
       },
       "delete": {
-        "title": "Delete File",
-        "message": "Removing this file from HTM Studio will delete the associated models. Are you sure you want to delete this file?"
+        "title": "Remove File",
+        "message": "Removing this file from HTM Studio will delete the associated HTM models. Are you sure you want to remove the file? Your original file will not be touched."
       }
     },
     "model": {
@@ -110,7 +110,7 @@
   },
   "menu": {
     "file": {
-      "delete": "Delete File",
+      "delete": "Remove File",
       "details": "File Details"
     }
   },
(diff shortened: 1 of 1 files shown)


lscheinkman:
The dialog question still says: '**Are you sure you want to delete this file?**' and button says '**Delete**'
screen shot 2016-06-22 at 3 48 32 pm

The menu still says '**Delete File**'
menu

marionleborgne:
@lscheinkman this wording comes from @subutai in UNI-562 and I took it as is. Should I change the body of the text to "Deleting this file from HTM Studio ..." ?

@subutai feel free to chime in. Ideally, I'd like to merge this today.

marionleborgne:
Also @lscheinkman I think the feedback was only about the text in the body of the waning message. Not about the buttons titles. Or am I wrong @subutai ?

subutai:
@marionleborgne I didn't suggest any specific text but I would avoid the words "delete file" everywhere, since the file is not actually being deleted.  For example it still says "delete file" at the top and "Are you sure you want to delete this file?" Also we might want to be explicit that the original file is not being harmed.

It would be clearer to say:

"Remove file.  Removing this file from HTM Studio will delete the associated models. Are you sure you want to remove the file? (Your original data file will not be touched.)"

For the menu item, instead of "delete file" it could simply say "Remove".

marionleborgne:
Thanks for the feedback @subutai (and for pointing this out @lscheinkman )

I updated the text and buttons. Let me know if that's ok:

screen shot 2016-06-23 at 7 33 12 am
screen shot 2016-06-23 at 7 32 59 am

subutai:
:+1:

PR #873, opened 2016-06-23
UNI-578 Handle rapid checkbox clicking

mrcslws:
This change fixes two similar bugs. When you rapidly click the show/hide checkbox, events can happen in an unexpected order. Now we detect this scenario in ModelDataStore and ModelStore.

@marionleborgne @lscheinkman

### DIFF
--- a/unicorn/app/browser/stores/ModelDataStore.js
+++ b/unicorn/app/browser/stores/ModelDataStore.js
@@ -66,22 +66,33 @@ export default class ModelDataStore extends BaseStore {
     let {modelId, data} = payload;
 
     let pendingModel = this._pendingModels.get(modelId);
-    if (!pendingModel) {
-      throw new Error('Listen for new results before querying.', modelId);
-    }
-    this._pendingModels.delete(modelId);
+    if (pendingModel) {
+      this._pendingModels.delete(modelId);
+      let model = {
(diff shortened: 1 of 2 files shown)


marionleborgne:
@mrcslws I got it to happen again when clicking fast on the checkbox of an already created model:

screen shot 2016-06-23 at 4 03 23 pm

lscheinkman:
@mrcslws I was able to reproduce on the running model as well.
screen shot 2016-06-23 at 4 10 23 pm

mrcslws:
Thanks for catching that. It was another variant of the issue. I pushed another commit.

marionleborgne:
@mrcslws the error dialog doesn't show up anymore 👍

marionleborgne:
Please not: this is making the show-stopper behavior go away (and that's why I'll merge the PR) but it's not fixing the underlying bug. We will leave the JIRA (UNI-578) open and will prioritize with Celeste accordingly. Thanks @mrcslws that's allowing us to have another release candidate at least. Whether or not we decide to prioritize fully fixing this bug (VS leaving it hidden like it is now in this PR) is a decision we'll make later after we have 0.0.10 release candidate. @cbaranski needs to test a new version tomorrow so that we can stay on track for public release.

PR #878, opened 2016-07-12
UNI-581: handle case where the sampling rate is slower than 4 weeks

marionleborgne:
This is fixing Len's bug. @ywcui1990 please have a look. The sample file is attached in UNI-581

cc @lscheinkman


marionleborgne:
@ywcui1990 when you have a chance, could you look at the changes? Thanks 😄

ywcui1990:
@marionleborgne LGTM :thumbsup:

PR #879, opened 2016-07-12
UNI-583 When a point has gaps on both sides, show a dot

mrcslws:
This used to work. It broke when I made the big d3 change, commit 48799da40e296b4bec4507c9127e622b2fa8f0d0

Dygraphs uses the "highlightCircleSize" parameter to choose the size of these dots. We no longer use Dygraphs highlight dots, so I disabled them. This had unintended consequences.

The fix: give the circles some size, and disable them via a callback.

@marionleborgne

### DIFF
--- a/unicorn/app/browser/components/Chart.jsx
+++ b/unicorn/app/browser/components/Chart.jsx
@@ -263,7 +263,7 @@ export default class Chart extends React.Component {
   }
 
   _yScaleUpdate() {
-    let paddingPx = 2;
+    let paddingPx = 3;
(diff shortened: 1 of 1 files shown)


marionleborgne:
👍 LGTM except that some dots at the top are cropped. See screenshot:
screen shot 2016-07-15 at 2 10 33 pm

mrcslws:
@marionleborgne good call. I added an extra pixel to the top and bottom.

marionleborgne:
👍 thanks. LGTM

PR #880, opened 2016-07-13
Fix relative path when script is called by "source" or "." operator

lscheinkman:
@marionleborgne Please review


marionleborgne:
👍 LGTM

Issue #504, opened 2015-12-07
Predictions are available as well as anomaly results


rhyolight:
@ryanjmccall said:

> My motivation is that on the project I'm working on we found that the anomaly likelihood was unsuitable for certain metrics. Instead we have had some success with using the predicted values in combination with some post processing to produce alerts.

Depends on #503 because model params with the right inference type must be provided by the user.

ryanjmccall:
@rhyolight thx, I'll take a look at this. I'm guessing it might be enough to add a 'prediction' row to the 'metric_data' schema and figure out where that should be populated

rhyolight:
I don't think it's that simple. You'll probably need to update some SQL tables. @oxtopus knows more.

ryanjmccall:
I am going to sketch out all the pieces that will have to change to expose the predicted value in addition to the anomaly scores. I am imagining there will be a new column added to the `metric_data` table for the current predicted value. (alternatively, it could be the previous predicted value, but, in that case, I don't know where the current predicted value would get stored). I have identified this method where the htmengine meets the OPF:

`htmengine.model_swapper.model_runner.ModelRunner#_processInputRow`

In this method there is a line: `r = self._model.run(inputRecord)`

Currently, the r.inferences dict only contains anomaly info like this: `{'anomalyLabel': '[]', 'anomalyScore': 0.0}`. I believe this is b/c the 'inferenceType' is 'TemporalAnomaly'. So one thing I have to figure out is how to coerce the OPF to return the predicted values but still comply with the restriction of HTM engine for TemporalAnomaly models. Do we relax this constraint or is there a way to ask for the prediction even if it's a TemporalAnomaly model? (@vitaly-krugl  @oxtopus)

Changing the metric_data table requires updating a lot of areas. So far I've identified, at the very least, the following changes:
- `htmengine.repository.schema`  (also consider whether to add an index for predicted value column)
- `model_inference_results_msg_schema.json` will require the predict value in the results section
- AnomalyService's `_composeModelInferenceResultsMessage` will have to support the predicted value
- `AnomalyService#_scrubInferenceResultsAndInitMetricData` employs a `MutableMetricDataRow` that would change
- `ModelInferenceResult` in the model swapper interface would have to be extended as well

Well that's a start for today. I'll keep studying the code and update this comment.

ryanjmccall:
[...]

PR #882, opened 2016-07-18
Multistep Best Predictions in HTM Engine

ryanjmccall:
Implements https://github.com/numenta/numenta-apps/issues/504


ryanjmccall:
currently missing unit tests for new functionality

ryanjmccall:
@rhyolight Hey Matt! How it going? This PR ready for a review. =) Thank you.

rhyolight:
@ryanjmccall I'm afraid you'll have to wait until either @vitaly-krugl or @oxtopus are available to review this. They are the experts in this codebase.

ryanjmccall:
@rhyolight understood, I'll wait. I wasn't sure if it's acceptable to ask them directly or I should go through you..

rhyolight:
@ryanjmccall I'm concerned about this getting merged because of what @vitaly-krugl said [on the original issue](https://github.com/numenta/numenta-apps/issues/504#issuecomment-231425151). Numenta wrote HTM Engine as an anomaly engine. Although I didn't see anyone directly object or 👎  the issue, his tone wasn't encouraging. 

@vitaly-krugl and @oxtopus can you guys weigh in here? Is this PR merge-able? Or was #504 doomed from the start?

ryanjmccall:
@rhyolight I agree with @vitaly-krugl that we have to consider the change carefully. However, I'm more optimistic about its feasibility. The core change is simply to add a column to a database table, which may be optionally populated. (The rest of the changes are simple ones to support this optional field). I imagine there may concerns backwards compatibility, so I'll await y'all's feedback. Thanks

htmengine/htmengine/model_swapper/model_swapper_interface.py
```diff
@@ -373,19 +373,28 @@ class ModelInferenceResult(_ModelRequestResultBase):
...
 
   __STATE_SIGNATURE__ = "iR"
 
 
-  def __init__(self, rowID, status, anomalyScore=None, errorMessage=None):
-    """ __init__(rowID, status, anomalyScore|errorMessage)
+  def __init__(self, rowID, status, anomalyScore=None,
+               multiStepBestPredictions=None,
+               errorMessage=None):
+    """ __init__(rowID, status, anomalyScore) or
```
oxtopus: What is the story behind these three lines in the docstring?

The original line they replace should not have passed muster, either.  I think someone was trying to document potential use-cases, but `__init__()` is never called directly, so it doesn't make sense.

Instead, the first line should be descriptive.

ryanjmccall: yeah I interpreted that line as use-cases as well -- `ModelInferenceResult` can be an error or it can be bona fide results. I can add a description instead

oxtopus:
There's something that doesn't quite feel right in this implementation with respect to how this is being feature-flagged.  I'm going to try to include related feedback here, rather than in multiple disparate inline comments.

* Config directive at https://github.com/numenta/numenta-apps/pull/882/files#diff-4433109f2b4b1126996825eb8920bf96R81 should be named something like "includeMultiStepBestPredictions" or "enableMultiStepBestPredictions", provided it is to be used as a global boolean "allow multistep predictions". 
* The flag is being used to toggle the `classifierEnabled` argument to the now modified `generateSwarmParams()`, some consistency in naming would be beneficial.

I would prefer to see a clearer code path for prediction models, distinct from anomaly model code path, but assuming tests pass, I don't see any reason not to move forward with this.  I'm going to see about running it through CI and will report back on results.

Thanks for all the cleanup!

htmengine/tests/support/skeleton-app/conf/application.conf
```diff
@@ -1,3 +1,7 @@
+[global]
```
vitaly-krugl: This may not be a good place for this feature flag. I will have more to say once I read more into the review.

htmengine/tests/support/skeleton-app/conf/application.conf
```diff
@@ -1,3 +1,7 @@
...
+# Flag specifying whether to additionally output multi-step best predictions
```
vitaly-krugl: What constitutes "best predictions"? Is it a certain percentage or something else?

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -25,20 +25,24 @@
...
+from htmengine.adapters.datasource.datasource_adapter_iface import (
+  DatasourceAdapterIface)
+import htmengine.exceptions as app_exceptions
+import htmengine.model_swapper.utils as model_swapper_utils
 from htmengine.repository import schema
 from htmengine.repository.queries import MetricStatus
 from htmengine.runtime import scalar_metric_utils
 import htmengine.utils
-import htmengine.model_swapper.utils as model_swapper_utils
-from nupic.data import fieldmeta
```
vitaly-krugl: Thanks for fixing up pylint and other findings along the way.

vitaly-krugl:
@ryanjmccall, does the change pass all HTM-IT tests?

ryanjmccall:
@vitaly-krugl Good point. I haven't tried the HTM-IT tests

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -25,20 +25,24 @@
...
+
+
+
+config = Config("application.conf", os.environ.get("APPLICATION_CONFIG_PATH"))
```
vitaly-krugl: There is a convenience constant for this already: just import and use APP_CONFIG; see https://github.com/numenta/numenta-apps/blob/6db0d4f6ab3a1d8fb3f339dba483346322f87196/htmengine/htmengine/__init__.py#L49

oxtopus:
See https://ci.numenta.com/browse/TAUR-TAUR6-TSTS-1/log for failing test results.  Most test failures seem to be related to `NoSectionError: No section: 'global'` errors.  Additionally:

* Need to mock out the config directive in the tests
* Consider adding a utility for setting the config that may be used in https://github.com/numenta/numenta-apps/blob/master/taurus_engine/docker/configure.sh

htmengine/htmengine/adapters/datasource/custom/__init__.py
vitaly-krugl: Also, the name `config` doesn't follow our naming conventions for either global vars (g_xxxx) or private constants (_CONFIG). This is just an FYI, since you don't need to define this here per my other comment concerning the existing `APP_CONFIG` constant.

vitaly-krugl:
@ryanjmccall: all all `taurus_engine` tests

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -73,6 +77,8 @@ def __init__(self, connectionFactory):
 
     self.connectionFactory = connectionFactory
 
+    self._includeMultiStepBestPredictions = config.getboolean(
```
vitaly-krugl: Does it need to be cached in the instance, or can we just get it from config where it's needed? It's definitely not on performance critical path. I am not sure what's best yet, but those thoughts crossed my mind.

ryanjmccall:
@vitaly-krugl do you mean: also run all `taurus_engine` tests?

htmengine/htmengine/adapters/datasource/custom/__init__.py
ryanjmccall: The parameter is only used by methods `activateModel` and `monitorMetric`. we could get it from config on the fly. that could be more parsimonious...

htmengine/tests/support/skeleton-app/conf/application.conf
ryanjmccall: looks like it is the most likely (max probability) of all the possibilities that the classifier is considering: https://github.com/numenta/nupic/blob/master/src/nupic/frameworks/opf/clamodel.py#L831-L839

vitaly-krugl:
@ryanjmccall, both taurus_engine and htm.it use htmengine, so might break by changes in htmengine.

htmengine/htmengine/runtime/scalar_metric_utils.py
```diff
@@ -77,7 +77,7 @@
 
 
 
-def generateSwarmParams(stats):
+def generateSwarmParams(stats, classifierEnabled=False):
   """ Generate parameters for creating a model
 
   :param stats: dict with "min", "max" and optional "minResolution"; values must
```
vitaly-krugl: Document the new parameter `classifierEnabled`

htmengine/htmengine/anomaly_likelihood_helper.py
```diff
@@ -114,7 +114,7 @@ def _generateAnomalyParams(self, metricID, statsSampleCache,
     assert len(scores) >= self._statisticsMinSampleSize, (
```
vitaly-krugl: Just to make sure I didn't miss anything, the changes in this module were limited to just aesthetic cleanup, right?

htmengine/htmengine/model_swapper/model_runner.py
```diff
@@ -191,7 +191,7 @@ def run(self):
         lastRequestBatch = None
 
         with self._swapperAPI.consumeRequests(
-            modelID=self._modelID, blocking=False) as consumer:
+          modelID=self._modelID, blocking=False) as consumer:
```
vitaly-krugl: We actually don't want any changes like this one. This is described somewhere in our python coding guidelines. Essentially, we want an extra level of indentation here so that there is a clear delineation between this outer statement and the enclosed body below it.

htmengine/htmengine/model_swapper/model_runner.py
```diff
@@ -509,16 +510,21 @@ def _processInputRow(self, row, currentRunInputSamples):
       return ModelInferenceResult(
         rowID=row.rowID,
         status=0,
-        anomalyScore=r.inferences["anomalyScore"])
+        anomalyScore=r.inferences["anomalyScore"],
+        multiStepBestPredictions=r.inferences.get("multiStepBestPredictions",
+                                                  dict()))
```
vitaly-krugl: Should we be constructing a new instance of (empty) dict alongside each result when best predictions is disabled? All the little things that make performance worse add up to something. Any reason not to let it default to None?

htmengine/htmengine/model_swapper/model_swapper_interface.py
vitaly-krugl: Better to replace `__init__` in the docstring with the class name.

htmengine/htmengine/model_swapper/model_swapper_interface.py
```diff
@@ -373,19 +373,28 @@ class ModelInferenceResult(_ModelRequestResultBase):
...
+        __init__(rowID, status, anomalyScore, multiStepBestPredictions) or
+        __init__(rowID, status, errorMessage)
 
     :param rowID: rowID id of the corresponding input record
     :param status: integer; 0 (zero) means success, otherwise it's an error code
       from htmengine.htmengineerno
     :param anomalyScore: the Anomaly Score floating point value if status is 0
       (zero), omit otherwise
+    :param multiStepBestPredictions: best model predictions according to the
+     associated classifier; key is the number of steps in the future for which
```
vitaly-krugl: Insufficiently indented

vitaly-krugl: Explain when to expect values in multiStepBestPredictions and when not to.

vitaly-krugl: In general, since this PR doesn't facilitate specification of the number of steps for the prediction, what is the only possible key to be expected here?

htmengine/htmengine/model_swapper/model_swapper_interface.py
```diff
@@ -395,6 +404,10 @@ def __init__(self, rowID, status, anomalyScore=None, errorMessage=None):
       assert isinstance(anomalyScore, (int, long, float)), (
         "Expected numeric anomaly score with status=0, but got: " +
         repr(anomalyScore))
+      assert (multiStepBestPredictions is None or
+              isinstance(multiStepBestPredictions, dict)), \
+        ("Expected dict multi-step best predictions with status=0, but got: " +
```
vitaly-krugl: Disconnect: The check is for None or dict, but the text says "Expected dict ..."

htmengine/htmengine/model_swapper/model_swapper_interface.py
```diff
@@ -395,6 +404,10 @@ def __init__(self, rowID, status, anomalyScore=None, errorMessage=None):
       assert isinstance(anomalyScore, (int, long, float)), (
         "Expected numeric anomaly score with status=0, but got: " +
         repr(anomalyScore))
+      assert (multiStepBestPredictions is None or
```
vitaly-krugl: Disconnect: the docstring says that multiStepBestPredictions is dict, but the check here is for None or dict.

htmengine/htmengine/anomaly_likelihood_helper.py
ryanjmccall: yep just nits

htmengine/htmengine/model_swapper/model_runner.py
ryanjmccall: I will look into making it None.

htmengine/htmengine/monitors/rmq_metric_collector_agent.py
```diff
@@ -247,7 +247,7 @@ def _parseArgs():
     rmqPort = int(rmqPort)
   except ValueError:
     msg = ("RabbitMQ Management Interface port must be an integer, but got %r"
-           % (metricDestPort,))
+           % (rmqPort,))
```
vitaly-krugl: Good catch

htmengine/htmengine/repository/queries.py
```diff
@@ -129,14 +129,14 @@ def deleteModel(conn, metricId):
     update = (schema.metric_data.update() # pylint: disable=E1120
               .values(anomaly_score=None,
                       raw_anomaly_score=None,
-                      display_value=None)
+                      display_value=None,
+                      multi_step_best_predictions=None)
```
vitaly-krugl: Make sure this is reflected in a test

htmengine/htmengine/model_swapper/model_swapper_interface.py
```diff
@@ -373,19 +373,28 @@ class ModelInferenceResult(_ModelRequestResultBase):
...
+  __slots__ = ("rowID", "status", "anomalyScore", "multiStepBestPredictions",
+               "errorMessage")
 
   __STATE_SIGNATURE__ = "iR"
 
 
-  def __init__(self, rowID, status, anomalyScore=None, errorMessage=None):
-    """ __init__(rowID, status, anomalyScore|errorMessage)
+  def __init__(self, rowID, status, anomalyScore=None,
+               multiStepBestPredictions=None,
```
vitaly-krugl: Make sure the new parameter `multiStepBestPredictions` is reflected in a test.

htmengine/htmengine/runtime/anomaly_service.py
```diff
@@ -87,14 +84,15 @@ def _getLogger():
 
 
 class MutableMetricDataRow(object):
-  __slots__ = ("anomaly_score", "display_value", "metric_value",
-               "raw_anomaly_score", "rowid", "timestamp", "uid")
+  __slots__ = ("multi_step_best_predictions", "anomaly_score", "display_value",
```
vitaly-krugl: There was a semblance of sorted order in the slots tuple. Please restore it.

htmengine/htmengine/runtime/anomaly_service.py
```diff
@@ -243,7 +241,7 @@ def _processModelCommandResult(self, metricID, result):
     # Create Model
     if result.method == "defineModel":
       self._log.info("Model was created for <%s>" % (
-                      getMetricLogPrefix(metricObj)))
+        getMetricLogPrefix(metricObj)))
```
vitaly-krugl: reexamine indentation

htmengine/htmengine/runtime/anomaly_service.py
```diff
@@ -462,7 +462,8 @@ def _composeModelInferenceResultsMessage(cls, metricRow, dataRows):
           ts=epochFromNaiveUTCDatetime(row.timestamp),
           value=row.metric_value,
           rawAnomaly=row.raw_anomaly_score,
-          anomaly=row.anomaly_score
+          anomaly=row.anomaly_score,
+          multiStepBestPredictions=json.dumps(row.multi_step_best_predictions)
```
vitaly-krugl: Is there a corresponding JSON schema file that needs to be updated?

vitaly-krugl: Make sure the new property `multiStepBestPredictions` is reflected in tests.

htmengine/htmengine/runtime/json_schema/model_inference_results_msg_schema.json
```diff
@@ -90,6 +90,12 @@
             "description": "Anomaly score computed by anomaly-likelihood algorithm",
             "required": true,
             "type": "number"
+          },
+
+          "multiStepBestPredictions": {
+            "description": "Multi-step best predictions from model's associated classifier",
```
vitaly-krugl: Would need a more detailed descriptions to be able to code against this documentation.

htmengine/htmengine/runtime/json_schema/model_inference_results_msg_schema.json
```diff
@@ -90,6 +90,12 @@
...
+            "required": true,
+            "type": "string"
```
vitaly-krugl: This should be an object; and the schema should describe the structure of this object.

htmengine/htmengine/runtime/anomaly_service.py
vitaly-krugl: The dump of row.multi_step_best_predictions is unnecessary, requiring an extra level of decoding by the recipient.  Get rid of `json.dumps` from this line, update the corresponding JSON schema, and update the test and recipient code as needed.

vitaly-krugl:
For future work, I am having mixed feelings about the abundance of aesthetic corrections (indentations and the like). Although I appreciate the +1 effort, the abundance of those unrelated corrections is a big detractor for reviewing the code.

htmengine/tests/integration/model_swapper/model_runner_test.py
```diff
@@ -25,7 +25,6 @@
 # Disable warning: Access to a protected member
```
vitaly-krugl: Shouldn't there be tests that test both cases: when inclusion of best predictions is enabled and when it is disabled?

htmengine/tests/support/skeleton-app/migrations/versions/003_1b8bb975b1d3_adding_multi_step_best_predictions_column.py
```diff
@@ -0,0 +1,48 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""adding multi_step_best_predictions column
+
+Revision ID: 1b8bb975b1d3
+Revises: 1d2eddc43366
```
vitaly-krugl: Based on the disconnect between the value of Revises in this comment "1d2eddc43366" and the value of `down_revision` in the code "872a895b8e8", I am guessing that this file was generated by copying another one and editing the values. We want all the linking to be done by the alembic tool. You want to have a recent version of alembic on your system and let it generate the revision file.

You want to start with code preceding your PR (e.g., `git checkout master`, assuming your master contains fresh numenta-apps master content), and use it to create the database (sans your changes). Then checkout your working branch that contains the relevant changes in `htmengine/htmengine/repository/schema`, and run the following commands. Then rename the generated file to follow our established naming pattern for revision files.

```
cd htmengine/htmengine/repository/migrations/
alembic revision --autogenerate -m “the comment that will show up in the file and whose prefix will be used to name the file"
```

htmengine/tests/support/skeleton-app/conf/application.conf
vitaly-krugl: A follow-up thought on global/multiStepBestPredictions config: "global" is not a good section name for it for reasons cited in prior comment about it; "htmengine" may be butter, but perhaps it doesn't belong in the config at all - shouldn't it be a model parameter instead?

htmengine/htmengine/model_swapper/model_runner.py
vitaly-krugl: I am curious what recent pylint with our latest and greats pylintrc from numenta-apps passed explicitly on command-line thinks about this case before vs. after the change. But regardless, we want an extra level of indentation here.

vitaly-krugl:
@ryanjmccall, thanks for your work on this PR. I am done with the first round of code review. Let me know when you're ready for the next round.

CC @rhyolight @scottpurdy

vitaly-krugl:
An additional workflow to consider that could result in bad trouble with the current state of this PR: let's say htm.it or Taurus (numenta is running an instance of it in production and staging) is stopped for a software upgrade and there happen to be some unprocessed messages in RabbitMQ that were created without the new "multi-step best predictions" property (e.g., model results published by model_runner and model inference results published by anomaly service).  After upgrading the software to include the commits from this PR, anomaly service (in htm.it or Taurus) and dynamodb service (in Taurus) begin consuming messages from RabbitMQ once again.

What happens if the new code is consuming a message generated by the old code without the "multi-step best predictions" property. E.g., will the code path that parses and consumes model results in model swarpper interface and anomaly service choke on this? I think that it might.

htmengine/tests/support/skeleton-app/conf/application.conf
ryanjmccall: @vitaly-krugl thanks for those thoughts. I don't like it in the .conf now either. I will try nestling it inside the `modelParams` of the metric spec passed into `htmengine.adapters.datasource.custom._CustomDatasourceAdapter#monitorMetric`.

vitaly-krugl: @ryanjmccall, I'm eagerly waiting to see what you will come up with 😺

htmengine/htmengine/model_swapper/model_runner.py
ryanjmccall: FYI, the numenta-apps pylintrc gives 'Wrong hanging indentation' and I initially removed the indentation in several places at its behest, later realizing all of these extra indentations with 'with-as' statements were for readability.

htmengine/htmengine/model_swapper/model_swapper_interface.py
ryanjmccall: I don't follow you. Are you suggesting that 1 is the only possible key? Since we allow custom model params now, the steps / keys can be altered by passed-in model params. I just confirmed it for 1 and 2 steps together. Nonetheless, I will work on the comments to your suggestion.

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -319,7 +325,8 @@ def monitorMetric(self, modelSpec):
       stats = {"min": minVal, "max": maxVal, "minResolution": minResolution}
       self._log.debug("monitorMetric: metric=%s, stats=%r", metricId, stats)
 
-      swarmParams = scalar_metric_utils.generateSwarmParams(stats)
+      swarmParams = scalar_metric_utils.generateSwarmParams(
+        stats, self._includeMultiStepBestPredictions)
```
vitaly-krugl: @ryanjmccall, why isn't self._includeMultiStepBestPredictions used in the completeModelParams case above?

htmengine/htmengine/model_swapper/model_swapper_interface.py
vitaly-krugl: @ryanjmccall: referring to https://github.com/ryanjmccall/numenta-apps/blob/d4e291ea8783e7093022ee91c4e39eabd79d546f/htmengine/htmengine/adapters/datasource/custom/__init__.py#L328-L329, it looks like your change is not acting on _includeMultiStepBestPredictions in the completeModelParams case, or did I misunderstand something?

ryanjmccall: @vitaly-krugl yes, your statement is correct about the current state of the PR. moving forward, I'm removing the multiStepBestPredictions 'directive'. in each case of the monitor metric call, whether to include predictions is passed into the model spec under modelParams. I'lld update the `ModelInferenceResult.__init__` docstring and let me know how ya like it =)

htmengine/htmengine/adapters/datasource/custom/__init__.py
ryanjmccall: in the completeModelParams case the modelParams should contain  'clEnable': True key-value pair

ryanjmccall:
@vitaly-krugl let's talk about the upgrade problem. I was able to generate situation like you describe (and seemingly just a the problem a guy in the forums was having...). the problem occurs if the [db-name].mswapper.results queue contains "old" messages lacking the best predictions. then the model swapper interface objects to reconstructing the `ModelInferenceResult` object from the received message raising an AssertionError because it detects that the number of attributes (defined in slots) doesn't match the number of attributes in the item's state. I see two possible solutions:
1) Add conditional logic to the __setstate__ method (https://github.com/numenta/numenta-apps/blob/master/htmengine/htmengine/model_swapper/model_swapper_interface.py#L167-Lundefined). Specifically, if the signature is "iR" and length of state is 5, insert None into the `state` array at position 4. that's pretty gross but I don't see a technical problem with it
2) Make no such code change. Instead, when upgrading, turn off the tap, let HTM engine empty this result queue, assert the queue is empty, then shutdown HTM engine and perform the upgrade. Seems fine assuming the queue empties reasonably quickly.

vitaly-krugl:
Hi @ryanjmccall, I'm sorry I didn't get a chance to follow up recently. I was neck-deep into my manylinux nupic.bindigns work in the past week+. I made a note to follow up this week, so please stay tuned.

By the way, htm.it was removed from numenta-apps, so there is no need for the "htm.it db migration" commit. htm.it is now archived at https://github.com/numenta/numenta-apps-archived-htm.it-grokcli-htmitmobile.

htm.it/htm/it/app/repository/migrations/versions/003_59c9ef218d3f_multi_step_best_predictions.py
```diff
@@ -0,0 +1,47 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have purchased from
```
vitaly-krugl: 2015 ==> 2016

htmengine/tests/support/skeleton-app/migrations/versions/003_315d6ad6c19f_adding_multistep_best_predictions_column.py
```diff
@@ -0,0 +1,47 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have purchased from
```
vitaly-krugl: 2015 ==> 2016

taurus_engine/taurus_engine/repository/migrations/versions/003_2695f59d78bd_multi_step_best_predictions.py
```diff
@@ -0,0 +1,47 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have purchased from
```
vitaly-krugl: 2015 ==> 2016

htm.it/htm/it/app/repository/migrations/versions/003_59c9ef218d3f_multi_step_best_predictions.py
```diff
@@ -0,0 +1,47 @@
+# ----------------------------------------------------------------------
```
vitaly-krugl: htm.it was removed from numenta-apps, so don't need this any longer.

vitaly-krugl:
@ryanjmccall, regarding
> and seemingly just a the problem a guy in the forums was having

Could you please post a link to that discussion on the forum?

vitaly-krugl:
I'll post a comment when I am done with this review round.

ryanjmccall:
Hi @vitaly-krugl thank a lot for checking in. I was also working on another task last week, but today I'm getting back to this PR. The PR is not done, but I did make some changes. I intend to next work on the missing integration/unit tests, which I have not gotten to. (I may have some questions for you about that, which I'll add to this PR.)

I did spend some time about a week ago trying to get the HTM engine dependencies and their test suites running. I got the unit tests to run but I was have troubles getting the integration test working for both taurus_engine and htm.it. I'll focus on adding the new tests first and try that later.

vitaly-krugl:
@ryanjmccall, regarding
> I got the unit tests to run but I was have troubles getting the integration test working for both taurus_engine and ~~htm.it~~. I'll focus on adding the new tests first and try that later.

Just to make sure you didn't miss my previous comment: no need to do anything for htm.it. htm.it got pulled out of numenta-apps and archived in a separate repo https://github.com/numenta/numenta-apps-archived-htm.it-grokcli-htmitmobile

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -411,7 +411,20 @@ def activateModel(self, metricId):
...
 
-    swarmParams = scalar_metric_utils.generateSwarmParams(stats)
+    enableClassifier = False
+    if metricObj.parameters:
+      try:
+        metricParameters = json.loads(metricObj.parameters)
+        if "modelParams" in metricParameters:
+          enableClassifier = metricParameters["modelParams"].get(
+            "enableClassifier", False)
+      except ValueError:
```
ryanjmccall: @vitaly-krugl upon reflection, it seems highly unlikely to me that a ValueError could occur here, given that a metric's parameters are serialized to string by `htmengine.adapters.datasource.custom._CustomDatasourceAdapter#_startMonitoringWithRetries`. If you can think of a way that the parameters could get messed up, I'll leave it in. Otherwise I think it's overkill.

ryanjmccall:
@vitaly-krugl Hi Vitaly. :smiley: I pushed up some more commits centered around testing and am ready for your feedback. From my end there are a few outstanding issues:

1) Upgrade problem (please see my comment above in this PR starting with the sentence "let's talk about the upgrade problem")
2) Question about catching a ValueError (please see my last comment in this PR).
3) Testing taurus_engine. The Bamboo CI is now running the taurus_engine unit and integration tests for the last two commits. Does this mean I can finally take a nap? :sunglasses:

vitaly-krugl:
@ryanjmccall: Regarding the upgrade problem comment https://github.com/numenta/numenta-apps/pull/882#issuecomment-238054824: it might be better to override `__setstate__` in `ModelInferenceResult`, check the state value, fix it up if needed, then pass control to the parent (_ModelRequestResultBase.__setstate`). We really don't want the base class `_ModelRequestResultBase` knowing any specifics of a derived class.

vitaly-krugl:
Regarding
> 3) Testing taurus_engine. The Bamboo CI is now running the taurus_engine unit and integration tests for the last two commits. Does this mean I can take a nap?

@oxtopus, can @ryanjmccall count on Bamboo CI he is referring to ?

htmengine/htmengine/adapters/datasource/custom/__init__.py
vitaly-krugl: @ryanjmccall, this code doesn't look right no matter how you slice it. An parameter format error is an error, and we shouldn't just warn, suppress, and assign some default behavior in case of the error.

htmengine/htmengine/adapters/datasource/model_spec_schema.json
```diff
@@ -30,6 +30,10 @@
         "minResolution": {
           "required": false,
           "type": "number"
+        },
+        "enableClassifier": {
```
vitaly-krugl: @ryanjmccall, a couple of things here:

1. `enableClassifier` seems pretty generic. There could be different classifiers, and one might never guess correctly which one is being referred to. For code readability, need a more specific property name.
2. Please add the description property, explaining what affect `True` will have. I don't think anyone would make the right connection to the persistence of best prediction results in the metric data table, otherwise.

htmengine/htmengine/adapters/datasource/custom/__init__.py
```diff
@@ -411,7 +411,20 @@ def activateModel(self, metricId):
 
     stats = self._getMetricStatistics(metricId)
 
-    swarmParams = scalar_metric_utils.generateSwarmParams(stats)
+    enableClassifier = False
+    if metricObj.parameters:
```
vitaly-krugl: @ryanjmccall, under which circumstances would `if metricObj.parameters` ever evaluate to false? I don't think that there are any. The reason I ask is that unnecessary checks mislead code maintainers into thinking that such a possibility might exist and get to wondering what else may be missing such checks, and start adding them unnecessarily elsewhere, too.

htmengine/htmengine/adapters/datasource/custom/__init__.py
vitaly-krugl: @ryanjmccall, so yes, I agree that you need to get rid of this ValueError check, and definitely not suppress errors. :smiley:

vitaly-krugl:
Regarding
> 2) Question about catching a ValueError (please see my last comment in this PR).

Pls see my comments starting with https://github.com/numenta/numenta-apps/pull/882/files#r76358819

htmengine/htmengine/adapters/datasource/custom/__init__.py
vitaly-krugl: That said, let's limit future fixes to real errors, and avoid aesthetic cleanup. I am really struggling with the amount of extra time the aesthetics demand during code review.

htmengine/htmengine/model_swapper/model_swapper_interface.py
```diff
@@ -373,19 +373,34 @@ class ModelInferenceResult(_ModelRequestResultBase):
...
     :param status: integer; 0 (zero) means success, otherwise it's an error code
       from htmengine.htmengineerno
     :param anomalyScore: the Anomaly Score floating point value if status is 0
       (zero), omit otherwise
+    :param multiStepBestPredictions: The best model predictions according to the
+     model's associated classifier. If the classifier is not enabled
+     the value is None. Otherwise, a dict of predicted value(s) stored by key(s)
+     equal to the number of steps in the future for which the prediction is
+     made. The temporal anomaly model default steps is [1], that is,
+     prediction for 1 step in the future, however, user-specified model params
```
vitaly-krugl: Reference to `completeModelParams`?

htmengine/htmengine/model_swapper/model_swapper_interface.py
```diff
@@ -403,18 +422,26 @@ def __init__(self, rowID, status, anomalyScore=None, errorMessage=None):
...
     self.errorMessage = errorMessage
 
 
   def __repr__(self):
-    return "%s<rowID=%s, status=%s%s>" % (
-      self.__class__.__name__, self.rowID, self.status,
-      (", anomalyScore=%s" % (self.anomalyScore,) if self.status == 0
-       else ", errorMsg=%s" % (self.errorMessage,)))
+    if self.status == 0:
+      reprBody = ", anomalyScore=%s, multiStepBestPredictions=%s" % (
```
vitaly-krugl: @ryanjmccall: do we have a unit test of this repr?

htmengine/htmengine/runtime/anomaly_service.py
```diff
@@ -106,10 +105,10 @@ def __init__(self, anomaly_score, display_value, metric_value,
 
   def __repr__(self):
     return ("%s<uid=%s, rowid=%s, ts=%s, value=%s, raw=%s, anomlik=%s, "
-            "display=%s>") % (
-      self.__class__.__name__, self.uid, self.rowid, self.timestamp,
-      self.metric_value, self.raw_anomaly_score, self.anomaly_score,
-      self.display_value)
+            "display=%s, multi_step_best_predictions=%s>") % (
```
vitaly-krugl: @ryanjmccall, do we have a unit test for this `__repr__`?

htmengine/htmengine/runtime/anomaly_service.py
```diff
@@ -242,8 +241,8 @@ def _processModelCommandResult(self, metricID, result):
 
     # Create Model
     if result.method == "defineModel":
-      self._log.info("Model was created for <%s>" % (
-                      getMetricLogPrefix(metricObj)))
+      self._log.info("Model was created for <%s>" % (getMetricLogPrefix(
+        metricObj)))
```
vitaly-krugl: This doesn't seem like the desired indentation. In fact, since you're changing it here anyway, this really aught to be following the correct python logger usage pattern:
```
self._log.info("Model was created for <%s>",
               getMetricLogPrefix(metricObj))
```
```

File: htmengine/tests/unit/model_swapper/model_swapper_interface_test.py [hunk 36]
```diff
@@ -550,17 +599,17 @@ def testMessageQueueNameOverrideViaConfig(self,
...
-        ((ModelSwapperInterface._CONFIG_SECTION,
-          ModelSwapperInterface._SCHEDULER_NOTIFICATION_Q_OPTION_NAME,
-          notificationQName),
-         (ModelSwapperInterface._CONFIG_SECTION,
-          ModelSwapperInterface._RESULTS_Q_OPTION_NAME,
-          resultsQName),
-         (ModelSwapperInterface._CONFIG_SECTION,
-          ModelSwapperInterface._MODEL_INPUT_Q_PREFIX_OPTION_NAME,
-          modelInputQPrefix))):
+      modelSwapperConfig.CONFIG_NAME,
```
vitaly-krugl: @ryanjmccall, please undo the "dedent". We really want the extra indentation as discussed in earlier comments

htmengine/tests/unit/runtime/anomaly_service_test.py
```diff
@@ -524,9 +523,9 @@ def testComposeModelInferenceResultsMessage(self, *_args):
       dataRows=dataRows)
 
     # Verify message against schema
-    with pkg_resources.resource_stream(
-        "htmengine.runtime.json_schema",
-        "model_inference_results_msg_schema.json") as msgSchemaStream:
+    with pkg_resources.resource_stream("htmengine.runtime.json_schema",
+                                       ("model_inference_results_msg_schema"
+                                        ".json")) as msgSchemaStream:
```
vitaly-krugl: @ryanjmccall, please don't break up a string in this fashion, just to make this style of indentation work out. That's not a good tradeoff. For example, when we `git grep` for this filename, we won't get a hit. In this case, please restore the original indentation, as it was correct per our guidelines.

vitaly-krugl:
@ryanjmccall, I am done with this round of code review. There are still a few issues outstanding, but it looks like we're almost there. Thanks!

htmengine/htmengine/model_swapper/model_swapper_interface.py
ryanjmccall: @vitaly-krugl Yes, in the model_swapper_interface_test I've updated the existing set of tests: https://github.com/numenta/numenta-apps/pull/882/files#diff-1a5ae049dd32f754ffb24c890418aa95R357

htmengine/htmengine/runtime/anomaly_service.py
ryanjmccall: nope. I can add some

htmengine/htmengine/model_swapper/model_swapper_interface.py
```diff
@@ -373,19 +373,34 @@ class ModelInferenceResult(_ModelRequestResultBase):
   """ Model inference result container """
 
   # NOTE: also used by serialization/deserialization in our base class
-  __slots__ = ("rowID", "status", "anomalyScore", "errorMessage")
+  __slots__ = ("rowID", "status", "anomalyScore", "multiStepBestPredictions",
+               "errorMessage")
 
   __STATE_SIGNATURE__ = "iR"
```
vitaly-krugl: @ryanjmccall, here is a brainstorming idea for keeping the logic that fixes up the old response (that someone might still have in the queue during software upgrade) outside the main code path and can be deleted later without impact on `ModelInferenceResult` and its tests . Note that I haven't tried it myself:

1. Modify `__STATE_SIGNATURE__` in `ModelInferenceResult`; e.g., `__STATE_SIGNATURE__ = "iRv2"`
2. Implement a tiny class with the original signature that fixes up parameters in `__setstate__`: ```
@_ModelRequestResultBase.__register__
class ModelInferenceResultLegacyV1(ModelInferenceResult):
  __STATE_SIGNATURE__ = "iR"

  def __setstate__(self, state):
    # Fix up: if status is 0, append `None` for multiStepBestPredictions to `state` and call super's `__setstate__`
    if state[2] == 0:
      state.append(None)
    
    super(ModelInferenceResultLegacyV1, self).__setstate__(state)
```

That's all that should be needed

File: htmengine/htmengine/model_swapper/model_swapper_interface.py [hunk 39]
```diff
@@ -373,19 +373,34 @@ class ModelInferenceResult(_ModelRequestResultBase):
   """ Model inference result container """
 
   # NOTE: also used by serialization/deserialization in our base class
-  __slots__ = ("rowID", "status", "anomalyScore", "errorMessage")
+  __slots__ = ("rowID", "status", "anomalyScore", "multiStepBestPredictions",
```
vitaly-krugl: It would be better to add the new property `multiStepBestPredictions` at the end, after the previously-existing fields, which is fairly common. It also maxes fix-up cleaner.

htmengine/htmengine/model_swapper/model_swapper_interface.py
```diff
@@ -373,19 +373,34 @@ class ModelInferenceResult(_ModelRequestResultBase):
...
+               "errorMessage")
 
   __STATE_SIGNATURE__ = "iR"
 
 
-  def __init__(self, rowID, status, anomalyScore=None, errorMessage=None):
-    """ __init__(rowID, status, anomalyScore|errorMessage)
+  def __init__(self, rowID, status, anomalyScore=None,
+               multiStepBestPredictions=None, errorMessage=None):
```
vitaly-krugl: If we're going to insert new parameters in front of existing ones, then I would prefer to to use this opportunity to re-order the parameter list in a more meaningful way:
```
def __init__(self, rowID, status, errorMessage=None, anomalyScore=None,
             multiStepBestPredictions=None):
```

since status and errorMessage are more closely related and are likely to be stable over time, while new parameters can then be added at the end in the future, which is common practice.

vitaly-krugl:
@ryanjmccall, I added a few more comments over the weekend.

htmengine/htmengine/adapters/datasource/model_spec_schema.json
ryanjmccall: @vitaly-krugl This parameter's name is derived from the NuPIC (>= v0.3.4) OPF parameter `clEnable` (https://github.com/numenta/nupic/blob/master/src/nupic/frameworks/opf/clamodel.py#L148-L149). I thought `clEnable` was quite impoverished myself and took the liberty of spelling out the word. I can add to the description that this parameter corresponds to this CLAModel param.

htmengine/htmengine/model_swapper/model_swapper_interface.py
```diff
@@ -445,6 +447,22 @@ def __repr__(self):
...
 
+@_ModelRequestResultBase.__register__
+class ModelInferenceResultLegacyV1(ModelInferenceResult):
+  """ Legacy model inference result lacking multi-step best predictions. """
+
+  __STATE_SIGNATURE__ = "iR"
+
+  def __setstate__(self, state):
+    # Convert STATE_SIGNATURE
+    self.__STATE_SIGNATURE__ = ModelInferenceResult.__STATE_SIGNATURE__
```
ryanjmccall: @vitaly-krugl note this line where I made a change beyond your initial suggestion. This is a work around to pass this check in the superclass: https://github.com/numenta/numenta-apps/blob/master/htmengine/htmengine/model_swapper/model_swapper_interface.py#L165

htmengine/htmengine/model_swapper/model_swapper_interface.py
ryanjmccall: Many thanks for thinking of this!

htmengine/htmengine/model_swapper/model_swapper_interface.py
vitaly-krugl: @ryanjmccall, thanks - I missed that in my brainstorming idea. Come to think of it, that assert `assert state[0] == self.__STATE_SIGNATURE__, repr(state)` is probably unnecessary, and should be covered through tests instead of a runtime check.

ryanjmccall:
@vitaly-krugl I'm finished with my round of changes.

vitaly-krugl:
Hi @ryanjmccall, I was on vacation. Will have a look this week. Thank you.

Vitaly

rhyolight:
@vitaly-krugl reminder ping!

vitaly-krugl:
Thanks @rhyolight. @ryanjmccall, I've not forgotten. BambooCI-related work took up all of my cycles last week. Things are looking better now.

vitaly-krugl:
Reviewing...

htmengine/tests/unit/model_swapper/model_swapper_interface_test.py
```diff
@@ -382,51 +404,117 @@ def testModelInferenceResultConstructorInvalidAnomalyScore(self):
...
-  def testModelInferenceResultSerializableStateWithAnomalyScore(self):
+  def testModelInferenceResultSerializableStateWithAnomalyScoreWithPreds(self):
+    self._auxTestModelInferenceResultSerializStateAnomScore(predictions={1: 1})
+
+
+  def testModelInferenceResultSerializableStateWithAnomalyScoreNoPreds(self):
+    self._auxTestModelInferenceResultSerializStateAnomScore(predictions=None)
+
+
+  def _auxTestModelInferenceResultSerializStateAnomScore(self, predictions):
```
vitaly-krugl: @ryanjmccall, please fix unusual abbreviations in `_auxTestModelInferenceResultSerializStateAnomScore`. It's okay to have longer function names in unit tests versus non-standard abbreviations.

vitaly-krugl:
Please also resolve the merge conflict, so that the PR test will be able to run.

htmengine/tests/unit/model_swapper/model_swapper_interface_test.py
ryanjmccall: @vitaly-krugl conflicts merged; unit tests renamed

vitaly-krugl:
Thank you @ryanjmccall; https://ci.numenta.com/browse/TAUR-TAUR-287 is happy with f3354851fca93f683b79770d3a4611ec6cb1a22a, so merging ...

ryanjmccall:
Many thanks for your help @vitaly-krugl

vitaly-krugl:
@ryanjmccall, you're welcome!

rhyolight:
Good job you guys! Way to stick it out, Ryan. :)

ryanjmccall:
Thanks @rhyolight and thanks for pushing this along. Enjoy all the multi-step best predictions.

PR #883, opened 2016-07-19
Making the gap detection a little less aggressive and more robust to outliers

marionleborgne:
The gap detection heuristic was too agressive and this was leading to a couple of datasets being represented as a cloud of dots. This PR fixes it: `water_treatment_log.csv` metrics are now represented as continuous series while the `JAO_apple_*.csv` still shows gaps between daily recordings. See screenshot below:

screen shot 2016-07-19 at 1 39 15 pm

@mrcslws please review


mrcslws:
Looks good 👍

PR #884, opened 2016-07-19
y-scaling: handle case where min == max

mrcslws:
The y scale was broken when there was only one point on-screen.

Thanks @marionleborgne for pointing out the bug.


marionleborgne:
👍

PR #894, opened 2016-09-17
logExceptions now takes a logger not a logger-getter function

### LINKED ISSUE #893: Invalid arg passed to logExceptions, causing failure when an exception needs to be logged
A number of uses of the logExceptions decorator in htmengine's model_swapper use the wrong implementation of `logExceptions` decorator from nupic.support.decorators instead of the one from nta.utils.error_handling. And they also pass the wrong arg type to it. The decorator accepts a logger object, but what's passed there is a function that returns the logger object.

https://discourse.numenta.org/t/problem-running-htmengine-skeleton-app/1403/5 reports service failures that are masqueraded by this bug as 
```
logger.exception(
AttributeError: 'function' object has no attribute 'exception'
```

History: this decorator used to accept a function that returns a logger, but then we had a PR that changed it to a logger object (instead of the function), however the implementer of that PR must have missed a few spots.

vitaly-krugl:
Fixes #893

Someone's earlier PR failed to fix them all up, so this should do the trick.

### DIFF
--- a/htmengine/htmengine/model_swapper/model_runner.py
+++ b/htmengine/htmengine/model_swapper/model_runner.py
@@ -39,7 +39,9 @@ import traceback
 from nupic.data.fieldmeta import FieldMetaInfo
 from nupic.data.record_stream import RecordStreamIface
 from nupic.frameworks.opf.modelfactory import ModelFactory
-from nupic.support.decorators import logExceptions
+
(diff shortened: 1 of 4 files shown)


vitaly-krugl:
CC @rhyolight @scottpurdy

PR #902, opened 2017-03-03
Add support for executing taurus monitors in docker containers

oxtopus:


install-taurus-monitoring.sh
```diff
@@ -24,10 +24,12 @@ set -o errexit
 
 function install {
   pushd $1
-    python setup.py develop --prefix=$2
+    python setup.py develop --prefix=$2 --no-deps
```
scottpurdy: This means that we have to manually install each projects dependencies?

What was the problem that led you to add `--no-deps`?

taurus_monitoring/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+    MODELS_MONITOR_TAURUS_MODELS_URL=CHANGEME \
+    MODELS_MONITOR_EMAIL_AWS_REGION=CHANGEME \
+    MODELS_MONITOR_EMAIL_SES_ENDPOINT=CHANGEME \
+    MODELS_MONITOR_EMAIL_SENDER_ADDRESS=CHANGEME \
+    MODELS_MONITOR_EMAIL_RECIPIENTS=CHANGEME \
+    MODELS_MONITOR_TAURUS_DYNAMODB_REGION=CHANGEME \
+    MODELS_MONITOR_TAURUS_DYNAMODB_AWS_ACCESS_KEY_ID=CHANGEME \
+    MODELS_MONITOR_TAURUS_DYNAMODB_AWS_SECRET_ACCESS_KEY=CHANGEME
+
+ENTRYPOINT ["/opt/numenta/taurus_monitoring/docker/entrypoint.sh"]
```
scottpurdy: nit: end file with newline character

taurus_monitoring/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+COPY htmengine /opt/numenta/htmengine
+COPY taurus_engine /opt/numenta/taurus_engine
+COPY taurus_monitoring /opt/numenta/taurus_monitoring
+COPY install-taurus-monitoring.sh /opt/numenta/
+
+# Install taurus monitoring
+WORKDIR /opt/numenta
+RUN ./install-taurus-monitoring.sh
+
+ENV DB_HOST=CHANGEME \
```
scottpurdy: It would be nice to have explicit checks for the environment variables. Perhaps leave these lines out and instead have checks in `entrypoint.sh` that the env variables have been set?

[How to check variable is set in bash](http://stackoverflow.com/a/307735)

taurus_monitoring/Dockerfile
```diff
@@ -0,0 +1,58 @@
+# ----------------------------------------------------------------------
+# Copyright (C) 2017, Numenta, Inc.  Unless you have purchased from
```
scottpurdy: This copyright header isn't correct. Looks like first line is missing

taurus_monitoring/docker/configure.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2016, Numenta, Inc.  Unless you have purchased from
```
scottpurdy: nit: year

install-taurus-monitoring.sh
oxtopus: To subvert `nupic` et al. dependencies that get installed by `htmengine`.  Don't need `nupic` for this, and don't want to have to accommodate.

scottpurdy: Are the `requirements.txt` incorrect or we just happen to be using only parts of the project that don't rely on the nupic dependency? It seems like it would be better to fix dependency graph rather that ignore it.

taurus_monitoring/Dockerfile
oxtopus: This same copyright is used many other places.  Sometimes there is `# Numenta Platform for Intelligent Computing (NuPIC)`, but this isn't part of nupic.  According to https://www.gnu.org/licenses/agpl-3.0.en.html, the first line in their example is `<one line to give the program's name and a brief idea of what it does.>`, so please let me know if you have a suggestion and I'm happy to add something, otherwise I don't think it's omission invalidates the copyright notice.

install-taurus-monitoring.sh
oxtopus: We don't need the full set of dependencies.  We're only using parts of `htmengine` and `nta.utils` and installing the full set of dependencies for those projects adds unnecessary burden.  I've compensated in `taurus_monitoring/requirements.txt`, which even in its current state is aggressive, but can be pared down as necessary.

taurus_monitoring/Dockerfile
oxtopus: `docker/configure.sh`, which specifies `set -o nounset`, uses these env vars to configure the environment including constructing a configuration file.  If any of the vars are unbound, `docker/configure.sh` will fail.

Regardless, the approach I've taken here is consistent with other `Dockerfile` implementations in taurus and elsewhere and is consistent with https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#env

scottpurdy: Oh cool, it looks like `-o nounset will do essentially what I was asking for. But `-o nounset` won't matter if you set the env vars to dummy values here, right?

Can we just list the env vars that are required in a comment here so that `-o nounset` can alert users that don't set them?

taurus_monitoring/Dockerfile
scottpurdy: Ok that seems fine to me.

@rhyolight - not specifically for this PR but can you provide guidance for future on whether we should be putting a first line describing the project or just using this generic header without the line?

install-taurus-monitoring.sh
scottpurdy: I don't want to hold up this PR but want to make sure we address this soon. Can you send me your thoughts on what we would have to do to these projects to not have to hand pick which dependencies we want?

oxtopus: I'll think about it.  Concretely, `htmengine` is included for the sole purpose of re-using what is effectively an enum: https://github.com/numenta/numenta-apps/blob/master/taurus_monitoring/taurus_monitoring/models_monitor/taurus_models_monitor.py#L32

A simple response is to just not import from `htmengine` in that one case and use a hard-coded value instead

PR #908, opened 2017-03-09
TAUR-2059 Break from depending on numenta/nupic:latest docker image

oxtopus:
cc @vitaly-krugl

### DIFF
--- a/install-taurus-metric-collectors.sh
+++ b/install-taurus-metric-collectors.sh
@@ -20,19 +20,7 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
-# Install taurus_metric_collectors and its dependencies
-# ARGS:
-# First position arg: installation prefix;
-#   e.g., Linux: /opt/numenta/anaconda
-#   e.g., Mac OS X: ~/Library/Python/2.7
-
 set -o errexit
 
-function install {
-  pushd $1
-  python setup.py develop --prefix=$2
-  popd
-}
-
-install nta.utils $1
-install taurus_metric_collectors $1
+pip install -e ./nta.utils \
+            -e ./taurus_metric_collectors
--- a/install-taurus-metric-collectors.sh
+++ b/install-taurus-metric-collectors.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 # ----------------------------------------------------------------------
 # Numenta Platform for Intelligent Computing (NuPIC)
-# Copyright (C) 2015, Numenta, Inc.  Unless you have purchased from
+# Copyright (C) 2015-2017, Numenta, Inc.  Unless you have purchased from
 # Numenta, Inc. a separate commercial license for this software code, the
 # following terms and conditions apply:
 #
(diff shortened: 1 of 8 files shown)


install-taurus-metric-collectors.sh
```diff
@@ -20,19 +20,7 @@
 # http://numenta.org/licenses/
 # ----------------------------------------------------------------------
 
-# Install taurus_metric_collectors and its dependencies
```
vitaly-krugl: Also, would be good to add that it needs to be executed from the numenta-apps directory as the CWD (since the implementation relies on it)

install-taurus-monitoring.sh
```diff
@@ -22,15 +22,12 @@
...
-pip install -r taurus_monitoring/requirements.txt
-
-install nta.utils $1
-install htmengine $1
-install taurus_engine $1
-install taurus_monitoring $1
+pip install -e ./nta.utils \
+            -e ./htmegine \
+            -e ./taurus_engine \
+            -e ./taurus_monitoring
```
vitaly-krugl: Missing terminating newline

install-taurus-monitoring.sh
```diff
@@ -22,15 +22,12 @@
...
 
-function install {
-  pushd $1
-    python setup.py develop --prefix=$2 --no-deps
-  popd
-}
+# nupic.bindings 0.2.1 (required by nupic 0.3.4) is not installable w/ pip.
+# Install explicitly w/ easy_install, which will install from .egg (as opposed
+# to .whl)
+easy_install nupic.bindings==0.2.1
```
vitaly-krugl: Pls add the reason that causes us to do this explicitly, as you did in the new comment in install-taurus.sh. It helps maintainers.

vitaly-krugl: Add `pip install numpy==1.9.2` for consistency with install-taurus.sh?

taurus_engine/requirements.txt
```diff
@@ -7,8 +7,6 @@ alembic==0.6.7
 web.py==0.37
 msgpack-python==0.3.0
 validictory==0.9.1
-nta.utils
-htmengine
```
vitaly-krugl: I am going to miss these lines, as they documented the dependency on these packages in the same place as the other dependencies, if nothing else.

install-taurus-monitoring.sh
oxtopus: Will do.  numpy isn't actually needed, however.  This line is really only in there so that subsequent calls to install nupic pass, itself not even really necessary under the circumstances.

taurus_engine/requirements.txt
oxtopus: Nonetheless, we have tests that will fail without them.  Following instructions in the readme will result in correct behavior anyway.

install-taurus-monitoring.sh
oxtopus: https://www.youtube.com/watch?v=kXsIfpnThFA

oxtopus:
@vitaly-krugl I've addressed your feedback.  Is there anything else to consider before merge?

PR #932, opened 2017-08-11
DEVOPS-369: Remove "grok" from code in public folders

lscheinkman:


unixorn:
I can't speak to the android/IOS changes, but the rest of it looks reasonable to me.

Pr0duktiv:
iOS Code looks fine to me syntaxwise

